PR #2, opened 2013-06-19
Support empty ping and pong packets.

fancycode:
According to RFC6455, Section 5.5.2 the "application data" for ping and pong packets is optional, this pull requests allows writing these packets even with empty payload.


warmcat:
Thanks for upur patch, I'm sorry it was overlooked and solved the same way in the meantime.

PR #3, opened 2013-06-19
Declare "sha1_init" statically to avoid naming conflicts

fancycode:
Such conflicts can happen if linked statically with another library providing a method with the same name.


warmcat:
Sorry this was fixed by d8b8542 at almost the same time.

PR #5, opened 2013-06-26
Minor changes.

ntdavidson:
Updated with new merge (thanks). Only patch left is the fix for erroneous byte count in write function.

Would you like me to put together a test function to prove that there is a problem?


warmcat:
Thanks for the patches, the remaining one got fixed in the meantime.

PR #6, opened 2013-06-27
Request "send" not to trigger SIGPIPE on errors.

fancycode:
Added "MSG_NOSIGNAL" to other occurrences of "send(...)".


warmcat:
These got refactored out, but I found one similar and patched it thanks.

PR #33, opened 2013-11-19
Support for HTTP POST. Use lws_body_copy() to retrieve the body from the...

kapejod:
I have addes support for the HTTP POST method and the Content-Length header.
You can retrieve the POSTed data with lws_body_copy().


select-this:
Would it be possible to revisit this? Currently test-server.c will end up getting to LWS_CALLBACK_HTTP_BODY_COMPLETION, but the client won't actually receive any response (unless I terminate the server with ctrl + c), it'll just sit there spinning forever.

PR #43, opened 2014-01-22
fix mingw64 builds

martell:
Fixed Mingw64 builds

_SSIZE_T_DEFINED is in mingw64 headers so It's a good swap :)

Also _TIMEZONE_DEFINED is in the mingw64 bit headers also :)

Please merge


warmcat:
Sorry every part of this is refactored.  If it's still needed and you have time, please reopen with an updatedpatch.

PR #44, opened 2014-01-22
removed needless __MINGW64__ #defines

martell:
Mingw64 also defines __MINGW32__ so a lot of code is redundant.

Also visual studio includes are case sensitive so websock-w32.c just needs to
#include winsock2 in lower case to support all platforms


warmcat:
The only remaining use isin gettimeofday.c, which is probably needed.

PR #141, opened 2014-07-05
Feature/http options method

andrew-canaday:
Add HTTP OPTIONS method, required for a sockjs library I'm writing which uses libwebsockets.


stozk:
Did you have any success with this? I'm trying to figure out how to create a c client to connect to a sockjs server but I don't know where to start. I've looked at the examples but I'm not sure which one would be the right one (I guess the test client one), but now I don't know how to implement the protocol part...

PR #171, opened 2014-09-04
Remove warning

deep125:
Remove warninig on windows and linux. Make x64 version on windows.


warmcat:
Sorry... some of these need different fixes, most have moved around since your patch (which is my fault).  I can see lws blows a lot of warnings on Windows, but there are no warnings with default compiler flags on gcc on Linux.  It means we're basically filling the sources with casts to shut the windows compiler up.

Issue #214, opened 2014-12-10
Feature request: Proxy Protocol


jmillan:
Hi,

I'm just wondering whether the Proxy Protocol[*] implementation is in the roadmap or if there is any interest at all.

It is useful for scenarios where the WebSocket server is behind a proxy, since it provides the IP:port information of the real WebSocket client, which is lost otherwise.

My understanding is that the content of such information, which comes previous to the HTTP request, could be stored in the allocated_headers struct under the token 'proxy ' so the information could be obtained as any other HTTP header is.

[*]: http://www.haproxy.org/download/1.5/doc/proxy-protocol.txt

Ie:

PROXY TCP4 192.168.0.1 192.168.0.11 56324 443\r\n
    GET / HTTP/1.1\r\n
    Host: 192.168.0.11\r\n
    \r\n

socketpair:
this commit is here fdde0f0a03a711bc08c2069ce92a9fee1526d450

@warmcat @jmillan  
does it work for WSS ? I mean plaintext PROXY header before the first byte of TLS HELLO message ?

PR #218, opened 2014-12-13
proxy protocol (version 1) parsing implementation

jmillan:
The proxy line is dumped in the token buffer as any other HTTP header

This PR corresponds to the Feature Request #214


socketpair:
this commit is here fdde0f0a03a711bc08c2069ce92a9fee1526d450

socketpair:
@jmillan @warmcat does it work for WSS ? I mean plaintext PROXY header before the first byte of TLS HELLO message ?

Issue #228, opened 2015-01-12
Support for PUT/PATCH/DELETE verbs


qpfiffer:
I have some patches in a fork that I'd love to see in mainline, but I don't want to just dump it in here with some discussion.

You can see the differences [here.](https://github.com/qpfiffer/libwebsockets/compare/warmcat:master...master) If theres any interest or feed back I'll make a pull request.

JoakimSoderberg:
You don't want a discussion? Or did you mean: "I don't want to just dump it in here **without** some discussion **first**."

The point with pull requests is to show the changes made and to use a discussion thread. It makes it easier, since individual lines and such can be commented. And the pull request gets updated if you push new changes to your branch.

Also, it is important, since if you create a pull request all the tests will run on your changes automatically so it's easy to see if compilation breaks on some platform.

In short: Create a pull request :)

qpfiffer:
Will do.

JoakimSoderberg:
@qpfiffer Hey! Are you going to create that pull request? I didn't mean to scare you away :)

JoakimSoderberg:
@qpfiffer Actually it was possible for me to do it for you in #230, didn't know that... Any further discussion can continue there

PR #230, opened 2015-01-22
Support for PUT/PATCH/DELETE verbs

JoakimSoderberg:
This is changes by @piffer commented on in #228

His comments:
I have some patches in a fork that I'd love to see in mainline, but I don't want to just dump it in here with some discussion.


qpfiffer:
The only thing I don't really see a reason for is the Uri-Args stuff, which I was using in another project in an older fork. The reason I didn't create the pull request was because it seemed kind of weird to include something that was removed.

warmcat:
Thanks... I squashed these into one patch without uri-args, and rewrote the bloating tests to use an array of method token indexes and a loop, since it seems the number of interesting methods is only growing.

qpfiffer:
Thats pretty cool, thanks guys

PR #231, opened 2015-01-24
Fix error handling when calling "REQUIRES PRIVATE KEY" callback.

cmuellner:
A recent patch in ssl.c introduced a callback for setting the
private key of the SSL context. This code contained a bug, which
resulted in lws_context_init_server_ssl() returning always
with a return value of 1, indicating an error.

This patch introduces the missing curly braces to fix the code's
intended behaviour.


JoakimSoderberg:
The failed test is due to missing OpenSSL download, see fix #232

warmcat:
Thanks I'll push it in the next minutes.... good find....

PR #234, opened 2015-01-27
Callback with LWS_CALLBACK_CLOSED_HTTP when HTTP closed before serve

drewnoakes:
If the remote HTTP client closes the connection before serving commences, the only notification client code receives is `LWS_CALLBACK_WSI_DESTROY`. This commit gives `LWS_CALLBACK_CLOSED_HTTP` as well, as would happen if HTTP serving had already commenced.

Without this patch we experienced a hard-to-debug segfault where we weren't aware the other end of the connection was gone and tried to asynchronously write to a stored wsi which has since been destroyed.

### DIFF
--- a/lib/libwebsockets.c
+++ b/lib/libwebsockets.c
@@ -92,6 +92,11 @@ libwebsocket_close_and_free_session(struct libwebsocket_context *context,
 		goto just_kill_connection;
 	}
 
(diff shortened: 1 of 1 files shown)


warmcat:
Thanks, it's pushed.

PR #238, opened 2015-01-30
Fix bug parsing unknown headers

drewnoakes:
Bug was introduced in 49f72aa45.

Discovered this as a user was sending `DNT: 1` (Do Not Track) due to enabling a browser setting. This is actually a handy way to test sending an unknown header from a browser.


warmcat:
Thanks it's pushed.

PR #245, opened 2015-02-28
call lws_plat_inet_ntop instead of inet_ntop

banthonywalker:


warmcat:
Thanks... I modified it to also build the lws_plat version without ipv6 and pushed.

JoakimSoderberg:
Thanks Travis-CI ;)

PR #248, opened 2015-03-04
updated SO version in spec file when building RPM

brobits:
received an error when building RPM, looking for libwebsocekts.so.4.0.0 but libwebsockets.so.5 was actually built.  updated the spec file to reflect the version change


warmcat:
Ah.... thanks, it's pushed

PR #249, opened 2015-03-06
fix win32 context memory leak

banthonywalker:


JoakimSoderberg:
Hello @banthonywalker, Please see the failed builds https://travis-ci.org/warmcat/libwebsockets/jobs/53371532 ... You need to have an `#ifdef _WIN32` for `int i` as well.... What surprises me is that this is windows only? Seems weird...

Btw @warmcat, any progress on the autobahn suite? Would be nice to run that through valgrind/dr.memory to catch stuff like these :)

http://www.drmemory.org/

banthonywalker:
sorry for build warning ...

warmcat:
OK it's pushed thanks.

I'm still anxious to use autobahn.  Maybe this weekend.

PR #251, opened 2015-03-09
Add windows installer file and configure appveyor to create it.

ralight:
This change adds an NSIS script for generating a Windows installer that contains the documentation, executables, dynamic library, linker libraries and include file for libwebsockets. It doesn't include dependencies - like openssl for example. I've tested the installer and it works just fine.

The changes to appveyor.yml should mean that the installer gets built and stored by appveyor. I've not tested this specifically, but have used the same technique in a different project. This means that there would be a automatically compiled version of libwebsockets available, for Windows. The version number in the installer comes from the appveyor build number, which is by default something 1.0.x, with x incrementing per build. I'm not sure what the best solution is here, perhaps creating the installer could be made something that only happens on tags being created.

### DIFF
--- a/lib/context.c
+++ b/lib/context.c
@@ -164,7 +164,11 @@ libwebsocket_create_context(struct lws_context_creation_info *info)
 #endif
 
 	if (lws_plat_init_fd_tables(context)) {
-#ifndef _WIN32
+#ifdef _WIN32
+		for (i = 0; i < FD_HASHTABLE_MODULUS; i++) {
+			lws_free(context->fd_hashtable[i].wsi);
+		}
+#else
 		lws_free(context->lws_lookup);
 #endif
 		lws_free(context->fds);
@@ -339,7 +343,11 @@ libwebsocket_context_destroy(struct libwebsocket_context *context)
 	lws_ssl_context_destroy(context);
 
 	lws_free(context->fds);
-#ifndef _WIN32
+#ifdef _WIN32
+	for (n = 0; n < FD_HASHTABLE_MODULUS; n++) {
+		lws_free(context->fd_hashtable[n].wsi);
+	}
+#else
 	lws_free(context->lws_lookup);
 #endif
 	lws_plat_context_late_destroy(context);


JoakimSoderberg:
Awesome! Actually artifacts from appveyour could be uploaded to github. I think doing that on tagged releases could be a good idea.

That's something @warmcat would need setup for this repo. From another projects of mine:

```yml
artifacts:
  - path: build\bin\Release  # Copies everything in this path

deploy:
  - provider: GitHub
    auth_token:
      secure: <secret thingie that needs to be specific to this repo>
    draft: false
    prerelease: false
    artifact: /.*\.zip/
    on:
      branch: master           # Release from master branch only.
      appveyor_repo_tag: true  # Deploy on tag push only.  <------ Tags only is probably a good idea
```

appveyor.yml
```diff
@@ -16,13 +16,26 @@ environment:
 install:
   - appveyor DownloadFile http://slproweb.com/download/Win32OpenSSL-1_0_1L.exe
```
JoakimSoderberg: https://chocolatey.org/packages/OpenSSL.Light

appveyor.yml
```diff
@@ -16,13 +16,26 @@ environment:
...
   - Win32OpenSSL-1_0_1L.exe /silent /verysilent /sp- /suppressmsgboxes
+  - appveyor DownloadFile "http://sunet.dl.sourceforge.net/project/nsis/NSIS 2/2.46/nsis-2.46-setup.exe" -FileName nsissetup.exe
```
JoakimSoderberg: You could replace this with a chocolatey package I think:
https://chocolatey.org/packages/nsis

`cinst nsis` should work.

appveyor.yml
```diff
@@ -16,13 +16,26 @@ environment:
...
   - cd build
-  - cmake %CMAKE_ARGS% ..
-  - cmake --build .
+  - cmake -DCMAKE_BUILD_TYPE=Release %CMAKE_ARGS% ..
+  - cmake --build . --config Release
+
+after_build:
+  - cd ..
+  - cd installer
+  - '%NSIS_ROOT%\makensis -DVERSION=%APPVEYOR_BUILD_VERSION% libwebsockets.nsi'
```
JoakimSoderberg: CMake actually has a builtin generator for CPack that generates NSIS stuff...

The benefit of using the built-in one is that the `INSTALL` targets in the CMake projects is used.
http://www.cmake.org/Wiki/CMake:CPackPackageGenerators#NSIS

This is not critical. But just so you know, and if you feel like changing it to use that instead.

warmcat:
Thanks a lot guys.  I'll try this out thisevening (~ +10h)

PR #253, opened 2015-03-10
Windows installer related fixes.

ralight:
Remove incorrectly committed cpack changes.
Fix call to makensis.
Update README filenames.

This results in an installer with the version number of the appveyor build. The test server certificates aren't generated, so they don't end up in the installer. Not sure why that is. I do think it would be better to just produce an installer for the tagged releases, but this is at least a start.


warmcat:
I committed it since it's better than what we had, but this installer script does not understand cmake options like "no server".  It tries to package the server binary anyway causing AppVeyor to blow up.

I feel a bit like the "I have no idea what I am doing" meme dog ^^ I get no urge to research windows problems... still any problems are localized to windows while we sort this out, the thing still builds even there, so I don't think it's a crisis.  But if it can grow some kind of preprocessor conditionals that know about cmake defines that'd be good.

JoakimSoderberg:
Yes this is what the CPack NSIS generator does. It only ties to copy things that are `INSTALL`ed by the CMake project: http://www.cmake.org/Wiki/CMake:CPackPackageGenerators#NSIS

appveyor.yml
```diff
@@ -28,7 +29,7 @@ build_script:
 after_build:
   - cd ..
   - cd win32port
-  - 'C:\Program Files\NSIS\makensis.exe -DVERSION=%APPVEYOR_BUILD_VERSION% libwebsockets.nsi'
+  - makensis -DVERSION=%APPVEYOR_BUILD_VERSION% libwebsockets.nsi
```
JoakimSoderberg: Only build this when `LWS_METHOD` is `default` (not sure how to check this in appveyor)

`if LWS_METHOD == default then makensis......`

warmcat:
Makensis seems to be solved thanks

PR #258, opened 2015-03-27
fix #257. Properly set the 'use_ssl' attribute on server wsi

### LINKED ISSUE #257: 'use_ssl' attribute is not set on server connections
Hi,

`use_ssl`attribute is set on client handshake for client connections but it's not set for server connections. This makes the `lws_is_ssl()` function always return 0 in this case.

Let me please provide a simple patch.

jmillan:


warmcat:
Thanks... I pushed it after making it conditional on SSL being enabled in Cmake.

PR #267, opened 2015-04-21
Ensure proper cleanup is performed if last websocket is closed.

### LINKED ISSUE #266: Crash on windows (with patch)
On windows libwebsockets can crash under the following conditions.

* The application only has a single websocket open.
* That websocket is closed and then a new socket is opened in quick succession.

The problem is that the codepath for closing the last websocket that is open skips removing the websocket from the filedescriptor hashtable, so a dead entry is left there. If windows then gives you back the same socket handle when you create the next websocket (which it can), attempts to read or write from the socket will access the dead entry which leads to a crash.

The fix is to do the same thing when closing the last websocket as we do when closing any other websocket. This leads to a few redundant operations, but simplifies the code and also fixes #213

joekilner:
Fixes #266


joekilner:
The build failed but I don't see why - it looks like one of the builds just stalled randomly. I'm going to close and reopen this push request to trigger another build

JoakimSoderberg:
Dead OpenSSL url was the cause for windows build failing

joekilner:
Looks like you've already pulled the commit. Thanks!

warmcat:
Yes... thanks for the patch.

PR #268, opened 2015-04-21
Fix windows build

JoakimSoderberg:
New version of OpenSSL.

This also fixes the failure in #267


JoakimSoderberg:
Hrm, ok so this way of building the nsis (installation program) project will not work properly since it does not care about what is being compiled or not... It tries to add the test-server when server is not being compiled :/

PR #269, opened 2015-04-21
Fix windows build properly

JoakimSoderberg:
Changed the nsis script instead so that missing files are not fatal


JoakimSoderberg:
This solves the problem in #267

warmcat:
Thanks a lot Joakim it's pushed.

PR #270, opened 2015-04-22
Multiple changes in the build process

wonder-mice:
I do realize that you will be against some of the changes, but hopefully we could have a small discussion, so I would better understand where the library is going from your point of view.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,9 @@
 cmake_minimum_required(VERSION 2.8)
 
+if(NOT DEFINED CMAKE_BUILD_TYPE)
+	set(CMAKE_BUILD_TYPE Release CACHE STRING "Build type")
+endif()
+
 project(libwebsockets C)
 
 set(PACKAGE "libwebsockets")
@@ -33,10 +37,19 @@ if(GIT_EXECUTABLE)
     message("Git commit hash: ${LWS_BUILD_HASH}")
 endif()
 
+set(LWS_USE_BUNDLED_ZLIB_DEFAULT OFF)
+if(WIN32)
+	set(LWS_USE_BUNDLED_ZLIB_DEFAULT ON)
+endif()
+
+option(LWS_WITH_STATIC "Build the static version of the library" ON)
+option(LWS_WITH_SHARED "Build the shared version of the library" ON)
 option(LWS_WITH_SSL "Include SSL support (default OpenSSL, CyaSSL if LWS_USE_CYASSL is set)" ON)
+option(LWS_USE_CYASSL "Use CyaSSL replacement for OpenSSL. When settings this, you also need to specify LWS_CYASSL_LIBRARIES and LWS_CYASSL_INCLUDE_DIRS" OFF)
+option(LWS_WITH_ZLIB "Include zlib support (required for extensions)" ON)
+option(LWS_WITH_LIBEV "Compile with support for libev" OFF)
+option(LWS_USE_BUNDLED_ZLIB "Use bundled zlib version (Windows only)" ${LWS_USE_BUNDLED_ZLIB_DEFAULT})
 option(LWS_SSL_CLIENT_USE_OS_CA_CERTS "SSL support should make use of OS installed CA root certs" ON)
-option(LWS_USE_EXTERNAL_ZLIB "Search the system for ZLib instead of using the included one (on Windows)" OFF)
-option(LWS_USE_CYASSL "Use CyaSSL replacement for OpenSSL. When settings this, you also need to specify LWS_CYASSL_LIB and LWS_CYASSL_INCLUDE_DIRS" OFF)
 option(LWS_WITHOUT_BUILTIN_GETIFADDRS "Don't use BSD getifaddrs implementation from libwebsockets if it is missing (this will result in a compilation error) ... Default is your libc provides it. On some systems such as uclibc it doesn't exist." OFF)
 option(LWS_WITHOUT_CLIENT "Don't build the client part of the library" OFF)
 option(LWS_WITHOUT_SERVER "Don't build the server part of the library" OFF)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,8 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.0)
```
JoakimSoderberg: Why change this? A lot of systems only have older versions. 2.8 is very common. Is there some new feature that's a must?

CMakeLists.txt
```diff
@@ -380,43 +410,38 @@ source_group("Sources"          FILES ${SOURCES})
...
-			${HDR_PRIVATE}
-			${HDR_PUBLIC}
-			${SOURCES})
+set(LWS_LIBRARIES)
+if (LWS_WITH_STATIC)
+	add_library(websockets STATIC
+				${HDR_PRIVATE}
+				${HDR_PUBLIC}
+				${SOURCES})
+	set(LWS_LIBRARIES ${LWS_LIBRARIES} websockets)
```
JoakimSoderberg: `list(APPEND LWS_LIBRARIES websockets)`

CMakeLists.txt
```diff
@@ -571,23 +589,29 @@ if (NOT LWS_WITHOUT_TESTAPPS)
 		source_group("Headers Private"   FILES ${TEST_HDR})
 		source_group("Sources"   FILES ${TEST_SRCS})
 		add_executable(${TEST_NAME} ${TEST_SRCS} ${TEST_HDR})
-		
+
 		if (LWS_LINK_TESTAPPS_DYNAMIC)
+			if (NOT LWS_WITH_SHARED)
+				message(FATAL_ERROR "Build of shared library is disabled, can't link the test apps to it")
```
JoakimSoderberg: change to `message(FATAL_ERROR "Build of shared library is disabled, can't link the test apps to it. LWS_LINK_TESTAPPS_DYNAMIC must be combined with LWS_WITH_SHARED")`

CMakeLists.txt
```diff
@@ -571,23 +589,29 @@ if (NOT LWS_WITHOUT_TESTAPPS)
...
+			endif()
 			target_link_libraries(${TEST_NAME} websockets_shared)
 			add_dependencies(${TEST_NAME} websockets_shared)
 		else(LWS_LINK_TESTAPPS_DYNAMIC)
+			if (NOT LWS_WITH_STATIC)
```
JoakimSoderberg: Same as above to be more helpful

CMakeLists.txt
```diff
@@ -891,13 +916,13 @@ if (RPMTools_FOUND)
...
 message("---------------------------------------------------------------------")
-message("  Settings:  (For more help do cmake -LH <srcpath>")
+message("  Settings:  (For more help do cmake -LH <srcpath>)")
 message("---------------------------------------------------------------------")
 message(" LWS_WITH_SSL = ${LWS_WITH_SSL}  (SSL Support)")
 message(" LWS_SSL_CLIENT_USE_OS_CA_CERTS = ${LWS_SSL_CLIENT_USE_OS_CA_CERTS}")
 message(" LWS_USE_CYASSL = ${LWS_USE_CYASSL} (CyaSSL replacement for OpenSSL)")
 if (LWS_USE_CYASSL)
-	message("   LWS_CYASSL_LIB = ${LWS_CYASSL_LIB}")
+	message("   LWS_CYASSL_LIBRARIES = ${LWS_CYASSL_LIBRARIES}")
```
JoakimSoderberg: You renamed this to `CYASSL_LIBRARIES` above

lib/libwebsockets.h
```diff
@@ -29,10 +29,9 @@ extern "C" {
 #else
 #include <stdarg.h>
 #endif
-	
-#ifdef CMAKE_BUILD
+
+/* That's a bad idea since it will leak all internal defines outside */
```
JoakimSoderberg: ?

CMakeLists.txt
JoakimSoderberg: Nevermind :)

JoakimSoderberg:
Test fails, since you changed CMake version. Don't do that

CMakeLists.txt
```diff
@@ -431,52 +456,46 @@ set(LIB_LIST)
...
-		# Create the library.
-		add_library(ZLIB STATIC ${ZLIB_SRCS})
-
-		# Set the same variables as find_package would.
-		set(ZLIB_INCLUDE_DIRS ${WIN32_ZLIB_PATH})
-		get_property(ZLIB_LIBRARIES TARGET ZLIB PROPERTY LOCATION)
-		set(ZLIB_FOUND 1)
-	else()
+if (LWS_WITH_ZLIB)
+	if (LWS_USE_BUNDLED_ZLIB)
```
JoakimSoderberg: You changed the logic here, so the default Windows build breaks.

On windows the default should be to use the bundled zlib

CMakeLists.txt
wonder-mice: I did it to avoid this warning:
```
CMake Warning (dev):
  Policy CMP0042 is not set: MACOSX_RPATH is enabled by default.  Run "cmake
  --help-policy CMP0042" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  MACOSX_RPATH is not specified for the following targets:

   websockets_shared
```
Also it's better to depend on latest libraries/tools, so when you need something from newer version nobody will say that this newer version is not available on some platforms :) Do you have an example of platform that doesn't have CMake 3.0?

warmcat: People build this on a lot of platforms, some funky and old like S390, ancient Debian stable, embedded platform with old rootfs basis who have no interest in making the latest and greatest available.  From that point of view "it's better" the minimum CMake version just reflects what it needs.... the warning is preferable.  Documenting it might be better.

JoakimSoderberg: Look at all the fail tests. Example of platform without latest CMake 3.0...

Depending on the latest isn't always best if you want as much portability as possible.

Solve this by doing `cmake_policy(SET CMP0042 OLD)` instead

lib/libwebsockets.h
wonder-mice: config.h defines a lot of generic names (like "inline", "_DEBUG", "VERSION", "HAVE_XXX", etc.). By including it in libwebsockets.h we expose all this defines to client code.

warmcat:
The patches except the big build change one all are good and applied thanks.

Basically I did the original autoblah build system, then Joakim contributed the much superior CMake one, which solved running sores like conflicting windows project file versions etc.  (and the autobuild pieces)  So for nontrivial changes related to build I just follow what he says.

PR #275, opened 2015-04-25
Re-indent source files (assuming tab size was 4)

wonder-mice:
Just did ":retab!" in vim for all files in "lib" folder. Probably some contributors had n-spaces for tab.
If my assumption about tab size (4) is wrong, I can re-indent with the correct value.


warmcat:
It's 8.  But the goal is follow Linux kernel coding style

https://git.kernel.org/cgit/linux/kernel/git/arm/arm-soc.git/tree/Documentation/CodingStyle

So simply replacing with hard tabs only isn't enough for the bits that are correct.  Any spaces that can be represented by a hard tab should be though.

Issue #264, opened 2015-04-17
libwebsockets calls libev's ev_break() on SIGINT


wonder-mice:
Hi,

libwebsockets is just a library and in my opinion it should behave properly and be a good citizen inside a hosting process. Obviously, event loop is used by other components.
It looks like libwebsockets was born as a standalone web server (or something) and only then decided to evolve into the library. That's understandable and appreciated. But once it decided to be a library it's better to minimize the interference with global process state, like calling ev_break(), setuid() or setgit() (I was shocked when discovered that the _transport_ library tries to call setuid()).

If you are agree with such thinking, I will provide patches to remove hostile calls.

warmcat:
"hostile calls"... I think you have gotten the wrong idea a bit too easily.

First it was a library from the very start, five years ago.

Setuid etc have to be opted into by the user code, they are provided as an optional convenience for the common case a low port < 1024 must be listened on, but otherwise root is not needed.  Formalizing that only enhances security and they are not called unless user code wants it.  So 'hostile' is completely backwards.

Libev support was contributed later, and it's clearly workable for his and others' usecases.  If you feel using a specific libev api is wrong for you, a patch to reduce it, make it smarter or select what to do as a runtime server option can be considered.  But unless it's an out-and-out bug please understand the library has been covering many different use-cases for five years and changes need to be smart enough not to disrupt every use-case but yours...

wonder-mice:
Ok, may bad I pulled setuid/setgid into the discussion. It's a separate topic. Just wanted to know your general line of thinking regarding such things.

Regarding ev_break(). I think that a transport library should not control the event loop. Event loop must be controled only by the app (or a specific library that provides process life cycle services). libwebsockets is just a transport library that is only a small part of some larger application. It doesn't know how to manage the event loop correctly any way.

I can come up with the only use case for that - smaller example applications. If you could explain where else it could be useful, I would appreciate it. But otherwise, I would like to ask you to consider removing ev_break() call. As I said, I can provide patch for that.

warmcat:
I originally wrote the library for headless embedded case.  So a device has no display, but it has network, then it can have a very rich UI via a browser.  That's why conveniences like the http serving logic and the optional daemonization are part of what it does.  You can link a small user code to this and get the rich UI in about 50KB or so of userland binary.

[...]

PR #277, opened 2015-04-27
Feature/ev sigint

andrew-canaday:
Resolves issue #264 in a backwards compatible way:

**_ONLY IF LWS_USE_LIBEV IS DEFINED_**

A new API function is provided:
```C
LWS_VISIBLE LWS_EXTERN int
libwebsocket_sigint_cfg(
    struct libwebsocket_context *context,
    int use_ev_sigint,
    lws_ev_signal_cb* cb);
```
It can (optionally) be invoked *after libwebsocket_create_context* and *before libwebsocket_initloop*.
 * To _prevent libwebsockets from setting the libev SIGINT handler entirely_: pass *use_ev_sigint* as *0*.
 * To instruct libwebsockets to install a _custom handler_, set *use_ev_sigint* to *1*, and pass your callback function as the third argument.
 * To instruct libwebsockets to use its _default handler_, set *use_ev_sigint* to *1* and set *cb* to ```NULL```.

If no action is taken, libwebsocket_sigint_cb will be installed, as usual to prevent breaking backwards compatibility.
For your convenience, the ev_signal callback has been typedef'ed for you as  lws_ev_signal_cb.

### DIFF
--- a/lib/context.c
+++ b/lib/context.c
@@ -314,6 +314,17 @@ libwebsocket_context_destroy(struct libwebsocket_context *context)
 		n--;
 	}
 
+#ifdef LWS_USE_LIBEV
+	/* (Issue #264) In order to *avoid breaking backwards compatibility*, we
+	 * enable libev mediated SIGINT handling with a default handler of
+	 * libwebsocket_sigint_cb. The handler can be overridden or disabled
+	 * by invoking libwebsocket_sigint_cfg after creating the context, but
+	 * before invoking libwebsocket_initloop:
+	 */
+	context->use_ev_sigint = 1;
+	context->lws_ev_sigint_cb = &libwebsocket_sigint_cb;
+#endif /* LWS_USE_LIBEV */
+
 	/*
 	 * give all extensions a chance to clean up any per-context
 	 * allocations they might have made
--- a/lib/context.c
+++ b/lib/context.c
@@ -126,6 +126,18 @@ libwebsocket_create_context(struct lws_context_creation_info *info)
 	context->ka_interval = info->ka_interval;
 	context->ka_probes = info->ka_probes;
 
+#ifdef LWS_USE_LIBEV
+	/* (Issue #264) In order to *avoid breaking backwards compatibility*, we
+	 * enable libev mediated SIGINT handling with a default handler of
+	 * libwebsocket_sigint_cb. The handler can be overridden or disabled
+	 * by invoking libwebsocket_sigint_cfg after creating the context, but
+	 * before invoking libwebsocket_initloop:
+	 */
+	context->use_ev_sigint = 1;
+	context->lws_ev_sigint_cb = &libwebsocket_sigint_cb;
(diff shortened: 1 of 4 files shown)


wonder-mice:
This is just a network library. It should not try to solve any problem that a hypothetical web server could have. This wiki article explains what I mean:
http://en.wikipedia.org/wiki/Separation_of_concerns

All this daemonization, setuid/setgid, SIGINT, etc. could be in a separate utility library (on top of libwebsocket) and libwebsocket will lose nothing, in contrary it will become simpler, more maintainable and will better serve its purpose.

I can't imagine any sane program that relies on SIGINT handling in *networking* library. It's like having thesaurus unicode support library (like ICU). "Why not" or "because it's a common pattern" are bad arguments for adding some unrelated things.

It's why Unix is so cool - because each tool does only one job and does it perfectly well. It's impossible to to do daemonization, setuid/setgid, SIGINT handling inside LWS correctly for any complex application. It is good only for examples and small demo applications. Please, don't increase library complexity with more useless runtime configurations.

andrew-canaday:
At the New York Times, we currently use libwebsockets +libev to dispatch _all_ browser notifications to end-users. The built-in sigint handling _is_ sanely incorporated into our application and nets us exactly the runtime behavior that is required for our services - though, I _understand your objection_. Philosophically, it makes sense. However, as Andy has pointed out, there are already _many_ users utilizing libwebsockets in a variety of production environments. **Not breaking the product for them is critically important.** My guess is that some people in the embedded world, in particular, _are_ leveraging the built-in daemonization and are happy about it.

_As was indicated earlier_, you can **opt out of any/all of the features you find offensive** - including ev_break utilizing either one of our patches.

warmcat:
Yes lws is now in, or on its way in several distros.  In the past we improved exported apis and so on in a noncompatible way a few times, but that's now very expensive for users / packagers to the point we're stuck with the apis we have.  And there's the related problem distros want to enable all the build options, so it's not enough to just make something a build option, it also has to be selectable at runtime.

I don't use libev personally but it looks like it does some housekeeping standalone libev users will like, it's in libev.c which means if you don't want to build libev support it costs nothing, it doesn't break what we had, and you are not forced to use it if you have a larger system using the event loop.

@wonder-mice Is there actually a practical problem for libev guys if I merge this?  It looks like people with sophisticated libev usage can continue to do the handling themselves.  And because the code is almost all local to libev.c I am not that worried it's bloaty.

wonder-mice:
> Is there actually a practical problem for libev guys if I merge this?

I can't tell for all "libev guys", but for me this discussion is enough to consider switching on wslay (which I trying to do right now).

andrew-canaday:
Well, best of luck to you, whatever you decide! :)

Ciao!

warmcat:
Yes foss is great like that.  If you think something else will fit your needs better, of course you should use that instead.  Thanks for your earlier patches and good luck.

warmcat:
This is finally pushed, thanks.

PR #278, opened 2015-04-29
update spec file for RPM generation

cejkato2:
Missing files that are installed, old name of shared object (missing
_shared in the name).

Additionally, for dynamic linkage, it was needed to update -l in the pc file.


warmcat:
I am sorry that change wasn't intentional, it has come from some contributed changes to CMake in a patchbomb a few days ago.

Rather than bend the spec file to comply with it I think we need to revert that change in CMakeLists, I'll study it later tonight.

warmcat:
I fixed it by reverting the change, at least for Linux.  Thanks for the patch anyway.

PR #286, opened 2015-05-05
Fix shared library name for all platforms.

ralight:
This is a further fix to f1b1254, building on 26fd0d2.


JoakimSoderberg:
Since the nsis windows installer does not use the CMake CPack stuff, it will not get the change name... So please rename it in https://github.com/warmcat/libwebsockets/blob/master/win32port/libwebsockets.nsi also

From appveyor:
```
File: "..\build\bin\Release\websockets_shared.dll" -> no files found.
```

Travis failed because the build timed out, so it should be fine if it is retriggered I think

PR #287, opened 2015-05-06
Fix shared library name for all platforms.

ralight:
This is a further fix to f1b1254, building on 26fd0d2.

Tidy up the LWS_WITH_SHARED and LWS_WITH_STATIC options.


warmcat:
Thanks, it's pushed.

PR #291, opened 2015-05-22
Winsock can return WSAEINVAL validly as return from non-blocking connect

bbuchanan-bcs:
I've encountered winsock non-blocking connect calls returning WSAEINVAL. It seems to only occur on 64 bit windows builds and I haven't been able to reproduce the issue inside the provided test apps. However, I've found multiple sets of documentation staging that WSAEINVAL can be returned by connect() in place of WSAEALREADY on occasion.

http://www.sockets.com/err_lst1.htm
Additional functions: Berkeley sockets connect() returns this error on subsequent calls, after an initial call on a non-blocking socket. However, some WinSocks fail with WSAEINVAL you call connect() a second time (or subsequent) on a non-blocking socket.

https://msdn.microsoft.com/en-us/library/windows/desktop/ms737625%28v=vs.85%29.aspx
A non-blocking connect call is in progress on the specified socket.
Note In order to preserve backward compatibility, this error is reported as WSAEINVAL to Windows Sockets 1.1 applications that link to either Winsock.dll or Wsock32.dll.

Using this patch in my own codebase resolves the issue. Usually the next connection attempt succeeds normally. You can also fix the issue bad adding brief delay between connect() calls that skips the time where WSAEINVAL is briefly returned however this is very brittle.


Frugality:
I had this exact same issue. Unfortunately I didn't see this until I had fixed the issue locally and made my pull request #315. Would be really nice to get this commited into master.

warmcat:
I took #315, thanks.

PR #294, opened 2015-06-08
Fix windows build errors

JoakimSoderberg:
NSI installera errors causing failed windows builds


warmcat:
I added it, but there's some confusion about the right name for, eg, _shared.dll or _static.lib.  I left both versions in since it's nonfatal.  But appveyor still dies on makensis

PR #288, opened 2015-05-10
Don't leak internal macros into public in lws_config.h.

ralight:

PR #296, opened 2015-06-24
Fix cmake define mistakes + memory leak / crash

JoakimSoderberg:
First of all, since `libwebsockets.h` includes `lws_config.h` we are exporting all our internal compile configuration macros to the world, which usually is not a good idea, something that someone already [pointed out](https://github.com/warmcat/libwebsockets/blob/master/lib/libwebsockets.h#L33) before.

Anyway, since we're doing that we should at least try to keep the names of all our exported macros prepended with `LWS_`.

Before I would get a bunch of warnings because of redefined macros when compiling libwebsockets together with other projects also containing `config.h` with the same `HAVE_` macros defined. This change fixes those warnings (and potential bugs caused by them).

Also remove invalid redefines of `pid_t`and `size_t`, this broke compilation of libwebsockets together with pthreads32 (which defines it for windows).


netrounds-joakim:
@warmcat This fixes the Windows build, please review/merge :)

I turned off creating the installer in the appveyor build, since as it stands it's very flaky. I will try to get time later to rewrite it so that the installers is created using CPack instead (this way it will only install the targets actually built, depending on the CMake options, instead of hacking it together like now).

JoakimSoderberg:
Added some memory leak and crash fixes as well @warmcat

JoakimSoderberg:
@ralight Ah right.... So many pull requests open, I realize I've looked at that one before now. I agree that's a better approach. I can incorporate that change in this PR instead.

@warmcat You should really close all the PRs that are merged or won't be merged, hard to keep track ... Maybe close all old ones and comment that people should reopen if they're still relevant maybe?

JoakimSoderberg:
@ralight Included your changes as well so #288 can be closed

JoakimSoderberg:
@warmcat poketi poke!

warmcat:
I think they're all okay, appveyer bailed on the third build with

nsis.install v2.46.0.20150406
 Downloading nsis.install 32 bit
   from 'http://sourceforge.net/projects/nsis/files/NSIS%202/2.46/nsis-2.46-setup.exe/download'
 Exception calling "GetResponse" with "0" argument(s): "Unable to connect to the
  remote server"

Seems some connectivity thing

PR #297, opened 2015-06-30
Pass the correct hostname when using a proxy

marc-groundctl:
The proxy doesn't modify the SNI request, so we need to pass the name of
the host we're trying to reach instead of the proxy we're reaching it
through.


warmcat:
Thanks, it's pushed.

PR #307, opened 2015-08-08
Move from CyaSSL to wolfSSL

ABruines:
This patch lets libwebsockets use the lastest version of wolfSSL (the new name for CyaSSL).
The reason for the patch is that allthough wolfSSL provides compatibility headers for (old) projects using CyaSSL, these are incomplete and do not work for libwebsockets.

The patch also fixes a typo in CMakeLists.txt where CYASSL_LIBRARIES was added to include_directories() instead of CYASSL_INCLUDE_DIRS.

Update: 
Allow the use of either LWS_USE_CYASSL or the new LWS_USE_WOLFSSL so we do not break CyaSSL support (see 3rd commit).


CMakeLists.txt
```diff
@@ -44,8 +44,8 @@ endif()
 
 option(LWS_WITH_STATIC "Build the static version of the library" ON)
 option(LWS_WITH_SHARED "Build the shared version of the library" ON)
-option(LWS_WITH_SSL "Include SSL support (default OpenSSL, CyaSSL if LWS_USE_CYASSL is set)" ON)
-option(LWS_USE_CYASSL "Use CyaSSL replacement for OpenSSL. When settings this, you also need to specify LWS_CYASSL_LIBRARIES and LWS_CYASSL_INCLUDE_DIRS" OFF)
+option(LWS_WITH_SSL "Include SSL support (default OpenSSL, wolfSSL if LWS_USE_WOLFSSL is set)" ON)
+option(LWS_USE_WOLFSSL "Use wolfSSL replacement for OpenSSL. When settings this, you also need to specify LWS_WOLFSSL_LIBRARIES and LWS_WOLFSSL_INCLUDE_DIRS" OFF)
```
JoakimSoderberg: In my opinion, I think it's better to add this option instead of renaming the old one. And then instead have a check:

```
if (LWS_USE_CYASSL)
    set(LWS_USE_WOLFSSL ON)
endif()
```

And same with the other two vars below. So you don't break peoples builds unnecessarily. But that's up to @warmcat to decide :)

ABruines: Perhaps you are right @JoakimSoderberg . Although I could argue that it might be better (to force) using the latest version of the SSL provider in this case. CyaSSL has had many updates since being renamed to wolfSSL ;)

ABruines: Fixed in the 3rd commit @JoakimSoderberg

PR #310, opened 2015-08-19
fix #309. Receive large messages over WSS

### LINKED ISSUE #309: Issue receiving large messages (> 4096 Bytes) with SSL enabled
Hi,

I'm using the latest libwebsocket version from git.
This issue is only present when WSS is used. Using plain WS everything is fine.

Incoming messages bigger than 4096 Bytes do not fire the protocol callback with reason `LWS_CALLBACK_RECEIVED`. 

The protocol `rx_buffer_size` is set to 65535.

It properly works if I manually set the macro `LWS_MAX_SOCKET_IO_BUF` to a value bigger than 4096.

This is the handshake log:

````
v13 hdrs completed
libwebsocket_parse sees parsing complete
checking sip
prot match 1
libwebsocket_ensure_user_space: 0x7fffbc001c40 protocol 0xbff9d0
lws_parse calling handshake_04
libwebsocket_ensure_user_space: 0x7fffbc001c40 protocol 0xbff9d0
libwebsocket_ensure_user_space: 0x7fffbc001c40 protocol pss 65552, user_space=-1140766752
WSI_TOKEN_EXTENSIONS = 'permessage-deflate; client_max_window_bits'
issuing resp pkt 158 len
callback_sip: LWS_CALLBACK_ESTABLISHED
Received network connect from 192.168.100.101:55817
    get  = /
    host: = 192.168.100.100:8000
    connection: = Upgrade
    upgrade: = websocket
    origin: = https://local-tryit.jssip.net
    sec-websocket-extensions: = permessage-deflate; client_max_window_bits
    sec-websocket-protocol: = sip
    http/1.1  = HTTP/1.1
    accept-encoding: = gzip, deflate, sdch
    accept-language: = en-US,en;q=0.8,es;q=0.6
    pragma: = no-cache
    cache-control: = no-cache
    sec-websocket-key: = x5gsAS5dhGtkrYg168gSnw==
    sec-websocket-version: = 13
    user-agent: = Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/46.0.2471.2 Safari/537.36
Allocating RX buffer 65557
accepted v13 connection
````
When the server receives a message bigger than 4096, the only log line I can see is:

````

libwebsocket_read: read_ok
```

Only when a second message is sent from the client, the protocol callback with reason `LWS_CALLBACK_RECEIVED` is fired and the two messages can be handled.

Any test I can do in order to give you more information, please just tell me and I'll provide it.

jmillan:
Read the full incoming TLS/SSL record at once in libwebsocket_service_fd().

SSL_read() is called until no more pending data for the current record is buffered in SSL.
SSL_read() is never requested more than the pending data size for the current record
to ensure that the fd is not read again for new data, which would be copied in the SSL buffer otherwise.

With this patch, the libwebsockets internal pending data handling (wsi->pending_read_list...) is not needed anymore. If the PR is accepted I could take care of removing it from the code.

### DIFF
--- a/lib/output.c
+++ b/lib/output.c
@@ -604,3 +604,9 @@ lws_ssl_capable_write_no_ssl(struct libwebsocket *wsi, unsigned char *buf, int l
 	lwsl_debug("ERROR writing len %d to skt %d\n", len, n);
 	return LWS_SSL_CAPABLE_ERROR;
 }
+
+LWS_VISIBLE int
+lws_ssl_pending_no_ssl(struct libwebsocket *wsi)
+{
+	return 0;
+}
--- a/lib/private-libwebsockets.h
+++ b/lib/private-libwebsockets.h
@@ -1111,6 +1111,7 @@ enum lws_ssl_capable_status {
 #define lws_context_init_http2_ssl(_a)
 #define lws_ssl_capable_read lws_ssl_capable_read_no_ssl
 #define lws_ssl_capable_write lws_ssl_capable_write_no_ssl
+#define lws_ssl_pending lws_ssl_pending_no_ssl
 #define lws_server_socket_service_ssl(_a, _b, _c, _d, _e) (0)
 #define lws_ssl_close(_a) (0)
 #define lws_ssl_context_destroy(_a)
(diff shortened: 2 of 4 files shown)


jwilm:
@jmillan nice work! I ran into the same issue, and this resolved it. Hopefully this will get merged.

warmcat:
Thanks, it's pushed.  If that's all that's needed, it's a lot simpler than the pending data stuff.

jmillan:
Thanks!

PR #323, opened 2015-10-10
added timeout reset in serve_http_file_fragment, to prevent timeout on big files.

Libco:
When sending big files (hundreds of MB over wifi), the connection always timed out after the AWAITING_TIMEOUT time, which is 5 seconds. So i added this line to reset it at each fragment sending and now I'm able to send the whole file.


warmcat:
Thanks it's pushed.

PR #324, opened 2015-10-11
Update appveyor to latest openssl.

ralight:
slproweb.com is as far as I can see a good trustworthy source of windows builds for openssl. They make a point of not providing out of date (most likely insecure) binaries, hence why there was a problem downloading it.

strcpy.net by comparison looks very dodgy.

This request updates to the most recent openssl installer from slproweb.com.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,6 @@
 environment:
+  SSL_VERSION: 1_0_2d
+
   matrix:
     - LWS_METHOD: default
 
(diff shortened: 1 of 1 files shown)


JoakimSoderberg:
This is fixed in #269 together with other appveyor fixes.

JoakimSoderberg:
correction, this is fixed in #296 together with other appveyor fixes ...

ralight:
Ah, it's been so long I was sure there must have been another openssl release... :)

I support #296, it is a good candidate for holiday merging.

PR #328, opened 2015-10-15
Fix getaddrinfo memory leak

hashok:
Tested with Valgrind.

### DIFF
--- a/lib/libwebsockets.c
+++ b/lib/libwebsockets.c
@@ -328,10 +328,11 @@ libwebsockets_get_addresses(struct libwebsocket_context *context,
 			    char *rip, int rip_len)
 {
 	struct addrinfo ai, *res;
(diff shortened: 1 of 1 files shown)


netrounds-joakim:
@hashok Did you test it with valgrind with one of the testapps? It would be nice to have a test program to catch stuff like this that can be run by the continous integration scripts... So if you have that, it would be nice to include.

hashok:
@netrounds-joakim No, I tested with my own "big" app which I can't share. But it should be easy to modify an existing testapp (if such exist) to catch this error - just call libwebsockets_get_peer_addresses() for IPv4 context and it will leak without this fix.

select-this:
Good catch; I just spotted that in the project I'm working on as well.

warmcat:
Great job, it's pushed thanks.

PR #335, opened 2015-10-23
Implemented fixes allowing libwebsockets to be built under Windows

stephaneberle9:
Hi,

I've tried to build libwebsockets under Windows using MinGW/MSYS, or more precisely

cmake -G "MSYS Makefiles" -DCMAKE_INSTALL_PREFIX=C:/MinGW .. 

This didn't work at the beginning due to a number of issues which I've investigated and fixed in a fork of the libwebsockets repo. As a result, the build runs through without no errors/warnings now.

As this could be interesting for other users too, I've created this pull request. It'd be great if you could check them and incorporate these changes into the original libwebsockets repo. Feel free to take them over as is or adapt them as appropriate. Let me know if you have any questions, I'll be happy to help.

Stephan


lws-team:
Can you review the travis and appveyor build errors your patches introduced please... eg on travis

/home/travis/build/warmcat/libwebsockets/test-server/test-echo.c:31:21: fatal error: process.h: No such file or directory

lib/private-libwebsockets.h
```diff
@@ -39,6 +39,8 @@
 #endif
 
 #if defined(WIN32) || defined(_WIN32)
+#define WINVER 0x0601
```
JoakimSoderberg: You're setting minimum version to windows 7. This is very restrictive. Why is this needed at all?

https://msdn.microsoft.com/en-us/library/windows/desktop/aa383745(v=vs.85).aspx#setting_winver_or__win32_winnt

lib/libwebsockets.h
```diff
@@ -234,15 +234,7 @@ struct libwebsocket_pollargs {
     int prev_events;   // the previous event mask
 };
 
-#ifdef _WIN32
```
JoakimSoderberg: You should not remove the normal Win32 support. All the MinGW support should have its own `#ifdef`s if they need something different

CMakeLists.txt
```diff
@@ -736,27 +738,10 @@ if (NOT LWS_WITHOUT_TESTAPPS)
 			set(TEST_SERVER_SSL_CERT "${PROJECT_BINARY_DIR}/libwebsockets-test-server.pem")
 
 			if (WIN32)
-				file(WRITE "${PROJECT_BINARY_DIR}/openssl_input.txt"
```
JoakimSoderberg: Don't remove the normal windows support. Instead you use use `if (MINGW)` instead, if its different from normal Win32

test-server/test-echo.c
```diff
@@ -33,6 +33,9 @@
 #include <syslog.h>
 #include <sys/time.h>
 #include <unistd.h>
+#else
+#include "../win32port/win32helpers/gettimeofday.h"
```
JoakimSoderberg: Simply include `"gettimeofday.h"`... The include path is added here: https://github.com/warmcat/libwebsockets/blob/master/CMakeLists.txt#L416

test-server/test-echo.c
```diff
@@ -208,7 +211,10 @@ int main(int argc, char **argv)
 	int disallow_selfsigned = 0;
 #endif
 
+#ifdef WIN32
```
JoakimSoderberg: use `#ifndef _WIN32` instead... always `_WIN32` as well not `WIN32`

test-server/test-echo.c
```diff
@@ -269,9 +275,12 @@ int main(int argc, char **argv)
 			rate_us = atoi(optarg) * 1000;
 			break;
 #endif
+#ifdef WIN32
```
JoakimSoderberg: `#ifndef _WIN32` and remove `#else`

test-server/test-echo.c
JoakimSoderberg: Second thought.... Why did you even put this in ifdef to begin with?

test-server/test-echo.c
JoakimSoderberg: Why ifdef at all?

test-server/test-echo.c
```diff
@@ -189,7 +192,7 @@ int main(int argc, char **argv)
 	struct libwebsocket_context *context;
 	int opts = 0;
 	char interface_name[128] = "";
-	const char *interface = NULL;
+	const char *_interface = NULL;
```
JoakimSoderberg: Why the name change?

test-server/test-ping.c
```diff
@@ -33,6 +33,8 @@
 #include <sys/ioctl.h>
 #include <poll.h>
 #include <unistd.h>
+#else
+#include "../win32port/win32helpers/gettimeofday.h"
```
JoakimSoderberg: See earlier comment on including this.

test-server/test-server.c
```diff
@@ -34,6 +34,7 @@
 #ifdef EXTERNAL_POLL
 #define poll WSAPoll
 #endif
+#include "../win32port/win32helpers/gettimeofday.h"
```
JoakimSoderberg: include properly

win32port/zlib/gzguts.h
```diff
@@ -12,7 +12,7 @@
 #  endif
 #endif
 
-#if ((__GNUC__-0) * 10 + __GNUC_MINOR__-0 >= 33) && !defined(NO_VIZ)
+#if ((__GNUC__-0) * 10 + __GNUC_MINOR__-0 >= 33) && !defined(NO_VIZ) && !defined(_WIN32)
```
JoakimSoderberg: This seems dodgy, having to change in zlib why is it needed? Are we using something internal to ZLIB somewhere?

CMakeLists.txt
```diff
@@ -543,6 +544,7 @@ if (LWS_WITH_ZLIB)
 			foreach (lib ${LWS_LIBRARIES})
 				add_dependencies(${lib} zlib_internal)
 			endforeach()
+			list(APPEND LWS_LIBRARIES zlib_internal)
```
JoakimSoderberg: This should not be needed. We're linking to zlib statically. `LWS_LIBRARIES` is only used for exporting the libwebsocket libraries to other cmake projects.

So this line can be removed

test-server/test-echo.c
JoakimSoderberg: Also... Why have you even added this include? None of your changes uses ?`gettimeofday`?

JoakimSoderberg: Basically none of the changes in this file have any relation to MinGW support, so remove them.

test-server/test-server.c
JoakimSoderberg: Correction, don't add this include at all

win32port/win32helpers/gettimeofday.c
```diff
@@ -1,5 +1,5 @@
 #include <time.h>
-#include <windows.h> //I've omitted context line
```
JoakimSoderberg: Since you change this. When building normal windows build, first the `timezone` struct is defined in the header, and then it's defined in here again. which breaks. This needs to be fixed so it works for both MINGW and normal

lib/private-libwebsockets.h
stephaneberle9: When left out, I get compiler warnings and linker errors due to absense of getnameinfo(), getaddrinfo(), and freeaddrinfo(). The reason for that is an #if (_WIN32_WINNT >= 0x0501) in the Ws2tcpip.h header included in the MinGW distribution. Given that, I agree that we could be less restrictive by going back to #define _WIN32_WINNT 0x0501. Nevertheless, MSDN tells that the minimum supported client for getnameinfo() is Windows 8.1, Windows Vista:

https://msdn.microsoft.com/en-us/library/windows/desktop/ms738532(v=vs.85).aspx

lib/libwebsockets.h
stephaneberle9: This is to avoid trouble like this:

w:/libwebsockets/test-server/test-server.c: In function 'main':
w:/libwebsockets/test-server/test-server.c:965:11: warning: passing argument 2 of 'libwebsocket_service_fd' from incompatible pointer type [enabled by default]
           &pollfds[n]) < 0)
           ^
In file included from w:/libwebsockets/test-server/test-server.c:44:0:
w:/libwebsockets/test-server/../lib/libwebsockets.h:1165:1: note: expected 'struct libwebsocket_pollfd *' but argument is of type 'struct pollfd *'
 libwebsocket_service_fd(struct libwebsocket_context *context,
 ^

resulting from the fact that pollfd is already defined in winsock2.h included by libwebsockets.h. Seems not to be anything specific to MinGW:

https://msdn.microsoft.com/en-us/library/windows/desktop/ms740094(v=vs.85).aspx

CMakeLists.txt
stephaneberle9: Agreed. I've changed that and committed/pushed it into my pull request.

test-server/test-echo.c
stephaneberle9: Yes, #include "gettimeofday.h" is much better and works out nicely.

The include as such is necessary though, otherwise I get the following warning:

w:/libwebsockets/test-server/test-echo.c: In function 'main':
w:/libwebsockets/test-server/test-echo.c:417:4: warning: implicit declaration of function 'gettimeofday' [-Wimplicit-function-declaration]
    gettimeofday(&tv, NULL);

test-server/test-echo.c
stephaneberle9: I introduced this to avoid the following warning:

w:/libwebsockets/test-server/test-echo.c: In function 'main':
w:/libwebsockets/test-server/test-echo.c:214:6: warning: unused variable 'debug_level' [-Wunused-variable]
  int debug_level = 7;
      ^
Wrt #ifdef/#ifndef WIN32/_WIN32: I've no problems to change that as suggested - but note that the same pattern is also used in line 275, 323, 332, and 431.

test-server/test-echo.c
stephaneberle9: The variable debug_level seems to be dead code under Windows. It is only defined (see first added #ifdef) and assigned with a command line argument (see second added #ifdef) but never actually used when being compiled with WIN32.

test-server/test-ping.c
stephaneberle9: This has been fixed as suggested.

test-server/test-echo.c
stephaneberle9: To avoid compile errors like those:

In file included from c:\mingw\include\rpcdce.h:10:0,
                 from c:\mingw\include\rpc.h:40,
                 from c:\mingw\include\windows.h:82,
                 from W:/libwebsockets/win32port/win32helpers/gettimeofday.h:12,
                 from w:/libwebsockets/test-server/test-echo.c:37:
w:/libwebsockets/test-server/test-echo.c: In function 'main':
w:/libwebsockets/test-server/test-echo.c:195:14: error: expected identifier or '(' before 'struct'
  const char *interface = NULL;
              ^
w:/libwebsockets/test-server/test-echo.c:296:14: error: expected '{' before '=' token
    interface = interface_name;
              ^
In file included from c:\mingw\include\rpcdce.h:10:0,
                 from c:\mingw\include\rpc.h:40,
                 from c:\mingw\include\windows.h:82,
                 from W:/libwebsockets/win32port/win32helpers/gettimeofday.h:12,
                 from w:/libwebsockets/test-server/test-echo.c:37:
w:/libwebsockets/test-server/test-echo.c:369:15: error: expected expression before 'struct'
  info.iface = interface;
               ^
make[2]: *** [CMakeFiles/test-echo.dir/test-server/test-echo.c.obj] Error 1
make[1]: *** [CMakeFiles/test-echo.dir/all] Error 2
make: *** [all] Error 2

test-server/test-server.c
stephaneberle9: No, because gettimeofday() is used in this module:

w:/libwebsockets/test-server/test-server.c: In function 'callback_http':
w:/libwebsockets/test-server/test-server.c:299:4: warning: implicit declaration of function 'gettimeofday' [-Wimplicit-function-declaration]
    gettimeofday(&tv, NULL);
    ^

win32port/zlib/gzguts.h
stephaneberle9: Without it, I run into the following warning:

In file included from w:/libwebsockets/win32port/zlib/gzio.c:13:0:
w:/libwebsockets/win32port/zlib/gzguts.h:16:0: warning: "ZLIB_INTERNAL" redefined [enabled by default]
 #  define ZLIB_INTERNAL __attribute__((visibility ("hidden")))
 ^
In file included from w:/libwebsockets/win32port/zlib/gzio.c:12:0:
w:/libwebsockets/win32port/zlib/zutil.h:19:0: note: this is the location of the previous definition
 #  define ZLIB_INTERNAL
 ^

Could be, that there is a better solution to fixing this issue.

CMakeLists.txt
stephaneberle9: I agree. Seems to be a left over from an earlier version of libwebsockets on which I had started to make all these changes some time ago.

win32port/win32helpers/gettimeofday.c
stephaneberle9: #include "gettimeofday.h" is required to avoid the following compile error:

w:/libwebsockets/win32port/win32helpers/gettimeofday.c:22:45: warning: 'struct timezone' declared inside parameter list [enabled by default]
 int gettimeofday(struct timeval *tv, struct timezone *tz)
                                             ^
w:/libwebsockets/win32port/win32helpers/gettimeofday.c:22:45: warning: its scope is only this definition or declaration, which is probably not what you want [enabled by default]
w:/libwebsockets/win32port/win32helpers/gettimeofday.c: In function 'gettimeofday':
w:/libwebsockets/win32port/win32helpers/gettimeofday.c:47:5: error: dereferencing pointer to incomplete type
   tz->tz_minuteswest = _timezone / 60;
     ^
w:/libwebsockets/win32port/win32helpers/gettimeofday.c:48:5: error: dereferencing pointer to incomplete type
   tz->tz_dsttime = _daylight;
     ^
make[2]: *** [CMakeFiles/test-client.dir/win32port/win32helpers/gettimeofday.c.obj] Error 1
make[1]: *** [CMakeFiles/test-client.dir/all] Error 2
make: *** [all] Error 2
[ 19%] Built target zlib_internal
[ 45%] Built target websockets

However, it is no problem to keep the #include <windows.h>. I'll commit that and check what the builds say.

test-server/test-echo.c
lws-team: This 'interface' is a msft perversion, they seem to #define it to 'struct' in the headers which is just as unfriendly as you'd expect from them.

So if we can't exclude the header that does the dirty, or #undef interface in private-libwebsockets.h,  changing 'interface' in the code is reasonable.

win32port/win32helpers/gettimeofday.c
stephaneberle9: Avoided breakage of build for normal Windows by cleaning up duplicate content in gettimeofday.c, removing header inclusions from gettimeofday.h and fixing include order in test-echo.c, test-ping.c and test-server.c

test-server/test-echo.c
JoakimSoderberg: Ok then it's fine, if it's simply to get rid of warnings

test-server/test-echo.c
JoakimSoderberg: Ah ok, well I see now. But I'd prefer you remove the `#ifdef` here, and instead move so that the `set_log_level` later is also used on windows (but not the syslog stuff). And while you're at it replace all  `WIN32` with `_WIN32`.

test-server/test-echo.c
JoakimSoderberg: Urgh, that's nasty... ok

win32port/zlib/gzguts.h
JoakimSoderberg: Ok I guess it's acceptable

build-msys.sh
```diff
@@ -0,0 +1,5 @@
+mkdir -p build
+cd build
+cmake -G "MSYS Makefiles" -DCMAKE_INSTALL_PREFIX=C:/MinGW -DCMAKE_BUILD_TYPE=DEBUG ..
+make
+make install
```
JoakimSoderberg: This seems a bit drastic for just building :|

JoakimSoderberg:
Had a few more comments, but looks good in general. Address those and I'll merge

test-server/test-echo.c
stephaneberle9: Fixed by removing #ifdef around definition and initialization of "debug_level" and moving the #endif that was behind the call to lws_set_log_level() in front of it (just like it already the case in test-server.c).

Also replaced all WIN32 occurrences by _WIN32 in test-echo.c, test-ping.c, and test-server.c as requested.

build-msys.sh
stephaneberle9: This is rather just a helper script for myself. We don't need to take that over into the actual repo. I think I'll convert it into an extra paragraph for the README.build.md and remove it then.

JoakimSoderberg: Yea I think that's a better idea to move it to the README.

stephaneberle9: Done.

JoakimSoderberg:
Squashed and merged in b820e2c2cc16620f6a664edf27377b3a7d9516a7

lws-team:
Nice job... @stephaneberle9 thanks a lot for dealing with the comments and doing the integration.

stephaneberle9:
Great! Thank you for reviewing and accepting this contribution. It is a pleasure to work with you.

PR #336, opened 2015-10-28
Define getdtablesize() for android

krrishnarraj:
getdtablesize() has been removed from headers since android-21.
Its value was 1024 in previous android api levels


JoakimSoderberg:
Thanks

PR #339, opened 2015-11-06
Fix for close ack sending

hashok:
This is related to a mailing list discussion with subject ['Close' frame with external message loop issue]


lws-team:
Looks good thanks, it's pushed.

PR #348, opened 2015-11-23
Terminate truncated header strings.

andrew-canaday:
Fixes a periodic segfault we've encountered wherein a header value gets truncated, but not null terminated.

### DIFF
--- a/lib/parsers.c
+++ b/lib/parsers.c
@@ -185,17 +185,24 @@ static int issue_char(struct libwebsocket *wsi, unsigned char c)
 		return -1;
 	}
 
-	if( wsi->u.hdr.ah->frags[wsi->u.hdr.ah->next_frag_index].len >= 
-		wsi->u.hdr.current_token_limit) {
-		lwsl_warn("header %i exceeds limit\n", wsi->u.hdr.parser_state);
-		return 1;
+	unsigned short frag_len = \
+		wsi->u.hdr.ah->frags[wsi->u.hdr.ah->next_frag_index].len;
+	/* If we haven't hit the token limit, just copy the character into
+	 * the header: */
+	if( frag_len < wsi->u.hdr.current_token_limit ) {
+		wsi->u.hdr.ah->data[wsi->u.hdr.ah->pos++] = c;
+		if (c)
+			wsi->u.hdr.ah->frags[wsi->u.hdr.ah->next_frag_index].len++;
+		return 0;
+	}
+	else {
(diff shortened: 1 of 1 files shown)


lib/parsers.c
```diff
@@ -185,17 +185,24 @@ static int issue_char(struct libwebsocket *wsi, unsigned char c)
 		return -1;
 	}
 
-	if( wsi->u.hdr.ah->frags[wsi->u.hdr.ah->next_frag_index].len >= 
-		wsi->u.hdr.current_token_limit) {
-		lwsl_warn("header %i exceeds limit\n", wsi->u.hdr.parser_state);
-		return 1;
+	unsigned short frag_len = \
```
JoakimSoderberg: Please declare any variables at the top of the scope to be C89 compatible (Microsoft compiler only support C89)

JoakimSoderberg: I mean older than what we're using currently on appveyor at least....

andrew-canaday: Certainly! I'll update momentarily. Thanks.

andrew-canaday:
As far as I can tell, the AppVeyor tests here (and on #349) are failing because the file at 'http://sourceforge.net/projects/nsis/files/NSIS%202/2.46/nsis-2.46-setup.exe/download' is not downloading consistently.
Are you able to confirm this, if you happen upon some time? If there are additional patches to be made, don't hesitate to let me know.

Ciao!

lws-team:
I mirrored it at https://libwebsockets.org/nsis-2.46-setup.exe, but the path to fetch it doesn't seem to appear in the yaml or appveyor project settings... if someone knows how to change it to dowloadfile like openssl and then cinst locally feel free to use the mirror url.

andrew-canaday:
Hey Andy,

   I have no way to test this, but I think perhaps changing
`cinst -y nsis`
to
`cinst -y nsis -source <some path to hosted source>`
Might work.

lws-team:
I tried a couple of variations on that but the builds failed miserably so I backed it out.

andrew-canaday:
Thanks for trying! Don't hesitate to ping if I can help.

lws-team:
Sorry I lost track of those for a few days, they're pushed now.

It's no problem but fyi the backslash at the end of the line is only needed in C when things absolutely must remain one blob, typically that's only preprocessor definitions and string continuations.  Anywhere whitespace comes between tokens, C doesn't care what that whitespace was or how many of them.

PR #349, opened 2015-11-23
Define 'daemonize' flag if LWS_NO_DAEMONIZE is not defined.

andrew-canaday:
This is required (for me) when building current master with: `-DLWS_WITHOUT_DAEMONIZE:BOOL=OFF`


lws-team:
Thanks this is also pushed.

PR #353, opened 2015-12-01
Allow zero-length pong to be received by server

hashok:
Client code already allows zero-length pongs


lws-team:
Thanks, it's pushed.

PR #354, opened 2015-12-02
Handle pending SSL reads which would otherwise not trigger a POLLIN.

andrew-canaday:
*To Reproduce:*
 * Using libwebsockets with libev backend
 * Setting max token length for headers to be some fixed number smaller than the server recv buffer (i.e. to avoid "excessive header content").
 * Send a request with an excess of read data

*Symptoms:*
 * libwebsockets appropriately handles TLS negotiation and reads data
 * SSL data exceeds receive buffer and is added to pending list
 * No additional POLLIN events are fired (as kernel has already notified us and SSL has consumed)
 * The pending list is never consulted.

*Changes:*
 * service.c: Add handle_pending label and jump if more content after lws_server_socket_service
 * ssl.c: set wsi->ssl = NULL on close (prevents potential segfault)

*Test:*
`curl -vk -H "Giant-header: $( for i in $( seq 5000 ); do echo -n 'a' ; done )" https://localhost:443/`

*Output*:
The request hangs after:
```
libwebsockets: insert_wsi_socket_into_fds: wsi=0x7f7f99f39350, sock=9, fds pos=1
libwebsockets: inserted SSL accept into fds, trying SSL_accept
libwebsockets: SSL_accept failed 2 / error:00000002:lib(0):func(0):system lib
libwebsockets: SSL_ERROR_WANT_READ
libwebsockets: SSL_accept failed 2 / error:00000002:lib(0):func(0):system lib
libwebsockets: SSL_ERROR_WANT_READ
libwebsockets: accepted new SSL conn
libwebsockets: libwebsocket_read: read_ok
```

### DIFF
--- a/lib/service.c
+++ b/lib/service.c
@@ -490,11 +490,12 @@ libwebsocket_service_fd(struct libwebsocket_context *context,
 		n = lws_server_socket_service(context, wsi, pollfd);
 		if (n < 0)
 			goto close_and_handled;
-		if( lws_ssl_pending(wsi) ) {
+		pending = lws_ssl_pending(wsi);
+		if( pending ) {
 			goto handle_pending;
 		}
 		else {
-		    goto handled;
+			goto handled;
 		};
 
 	case LWS_CONNMODE_WS_SERVING:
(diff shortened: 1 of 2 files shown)


lws-team:
Thanks... they're pushed.

It was a bit hard to see what the aggregated patches did, so I squashed them into one and cleaned it a little.  Then I split the ssl = NULL patch into his own patch.  But the end result is functionally the same.

andrew-canaday:
Cool. Thanks. :)

PR #361, opened 2015-12-08
Fixed to initialize sockaddr_in6 by zero

asuuma:
I found that sin6_scope_id was not set to zero in my machine (Mac OS X).
This cause me not able to open socket with IPv6 network.
I don't know which environment that doesn't have sin6_scope_id, 
but I think initializing by zero will fix this problem and doesn't affect to environment without sin6_scope_id.


lws-team:
ありがとございます。It's pushed.

asuuma:
Wow, I was surprised to hear Japanese!
You're welcome.

PR #362, opened 2015-12-08
Feature request: Access to SSL from LWS_CALLBACK_ESTABLISHED

ABruines:
A small feature request that will allow access to wsi->ssl at LWS_CALLBACK_ESTABLISHED.
As far as I can tell this patch can be applied without any side effects.
It calls LWS_CALLBACK_ESTABLISHED with wsi->ssl as the value for 'in' of the protocol callback.
This allows a user to do:

X509 *x509 = SSL_get_peer_certificate((SSL*)in);

With this the application can inspect the client's SSL certificate.
(This could also be done from LWS_CALLBACK_OPENSSL_PERFORM_CLIENT_CERT_VERIFICATION but the retrieved information can not be stored from here.)


lws-team:
It's pushed thanks.

PR #363, opened 2015-12-09
Custom file callbacks

SoapyMan:
Hi!
This patch adds file api callback support for built-in http file transfer, allowing applications to use their custom virtual file systems (archives, etc.)


lws-team:
Thanks for updating it... can I ask what the motivation was to abstract file access?  I can see it could be useful on, eg, mbed3.

SoapyMan:
I building a server app and I wanted to use html and image files from zip archive without unpacking them, I felt boring in implementing the custom http file transfer, so I made this.

lib/file-callbacks.c
```diff
@@ -0,0 +1,118 @@
...
+void compatible_file_close(void*fd){
+	close((int)fd)
+}
+
+unsigned long compatible_file_seek_cur(void* fd, long offset){
+	return lseek((int)fd, offset, SEEK_CUR);
+}
+
+void compatible_file_read(unsigned long amount, void* fd, unsigned char* buf, unsigned long* len){
+	*amount = read((int)fd, buf, len);
```
JoakimSoderberg: amount isn't a pointer? :|

test-server/test-server.c
```diff
@@ -296,6 +296,8 @@ int main(int argc, char **argv)
 		 * which also includes libwebsocket sockets
 		 */
 
+		pollfds->events = (0x0100 | 0x0200);
```
JoakimSoderberg: This should probably use defines?

JoakimSoderberg:
Please fix errors on Linux https://travis-ci.org/warmcat/libwebsockets/builds/95755136

JoakimSoderberg:
Nice, I like this! Maybe make the functions that shouldn't be exported `static` if there are any?

lws-team:
If we want to do it, all windows stuff should stay in plat-win.  So plat-* should contain the business end of those, and then no need for a new .c file.

As I mentioned it's also quite useful for mbed3, where there's no posix or filesystem, but we still need to serve blobs based on names from the uri.

test-server/test-server.c
SoapyMan: Uh, I'm a Windows user, I wanted to figure out why poll doesn't work

lws-team:
Soapyman I just pushed a modified version of what your patch was doing, please have a look at it

SoapyMan:
I can't build since windows  have _O_RDONLY instead of O_RDONLY in <fcntl.h>. And open for write will be implemented in future?

lws-team:
Okay the funny microsoft flag problem should be solved now.

Yes it should support write I think.  I added it in plat-unix but someone else will have to write the plat-win one.  Currently only read is used inside the library itself.

SoapyMan:
Shouldn't typedef lws_filefd_type be just void* for all platforms?

lws-team:
Nope.  Because on unix, a file descriptor is an int, which can't go in a void * without a cast.

Using a typedef lets the user code also hide the type cleanly in his code while still remaining typesafe.

SoapyMan:
Or maybe typedef intptr_t lws_filefd_type to be compatible with 64 bit pointers? I made changes in my fork, please have look at it

lws-team:
IIUI in windows HANDLE is an native opaque pointer type already.  There's no need to do anything about 64-bit compatibility.  If it was built for 64-bit environment the sizeof HANDLE will be 64-bit already.

Also if you just set any fops members you want to override directly, you can just override some, like the test server just overrides open.  And you can still use / call through to the lws platform versions by stashing a copy of the original fops after return from lws_create_context().  The code needed for this is smaller without anything related to external init in the library and there's no need to add to _info that way either.

If lws_filefd_type is an int on unix, and a HANDLE on windows, there's no need for any of the casts each time you use it.

lws-team:
I think I understood your meaning about the void *.... it's not about the file handle particularly but about providing context + state to the file request, including the file handle.

I pushed a patch that adapts the apis to take a wsi to provide context for the access, and another that provides an accessor API so you can get the user data pointer from the wsi alone.

lws-team:
This has probably gotten where it's going, please reopen if something else to patch.

PR #365, opened 2015-12-09
Fixed not to use IPv4-mapped address for IPv6 only node

asuuma:
Specifying AF_INET6 and AI_V4MAPPED for getaddrinfo, it will only return IPv6 addresses. If target host doesn't have IPv6 address, than it will return IPv4-mapped address. So this commit is both compatible with IPv6 only node and IPv6/IPv4 dual stack node. This also improve the code since you don't need to create IPv4-mapped address by self.
I tested both IPv6 only network and IPv4 network on my Mac and it worked, but I don't know about other OS like Windows. I believe this commit will work with any OS.


lws-team:
Thanks a lot, it's pushed.

PR #366, opened 2015-12-11
Revert using AI_V4MAPPED flag to getaddrinfo for Android

asuuma:
I found getaddrinfo will return error on Android when AI_V4MAPPED is specified. So we should use old implemntation only for Android. Android support of IPv6 is very poor, unfortunately.
It does not mean we should not use AI_V4MAPPED for other environment, because old implementation lack support of IPv6-only network.


lws-team:
わかりました。It's pushed thanks.

PR #368, opened 2015-12-13
Refactoring conditional directives for if wrappers.

wmarquesr:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

* [Linux kernel coding style](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892)
* [Living in the #ifdef Hell](https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/)

It might improve code understanding, maintainability and error-proneness.

### DIFF
--- a/lib/pollfd.c
+++ b/lib/pollfd.c
@@ -54,7 +54,7 @@ insert_wsi_socket_into_fds(struct lws_context *context,
 	wsi->position_in_fds_table = context->fds_count;
 	context->fds[context->fds_count].fd = wsi->sock;
 	context->fds[context->fds_count].events = LWS_POLLIN;
-	
+
 	lws_plat_insert_socket_into_fds(context, wsi);
 
 	/* external POLL support via protocol 0 */
@@ -107,7 +107,7 @@ remove_wsi_socket_from_fds(struct lws_context *context,
 	 * (still same fd pointing to same wsi)
 	 */
 	/* end guy's "position in fds table" changed */
-	wsi_from_fd(context,context->fds[context->fds_count].fd)-> 
+	wsi_from_fd(context,context->fds[context->fds_count].fd)->
 					position_in_fds_table = m;
 	/* deletion guy's lws_lookup entry needs nuking */
 	delete_from_fd(context,wsi->sock);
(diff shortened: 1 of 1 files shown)


lws-team:
I appreciate this cleans a lot of whitespace dirt but there's really no meaning to replacing

if (simple-conditional)

with a new variable holding the same conditional and then if (var).  It's making it harder to read if we did this everywhere.

Trust me, whatever you heard, the Linux kernel is chock full of if (simple-conditional).

wmarquesr:
Maybe, the idea is always to encompass only complete/entire statements/expressions with preprocessor conditional directives. It is something that is getting popular in many different projects (started mainly on the Linux Kernel).

We agree totally that breaking statements with directives influences code quality, i.e., understanding, maintainability and error proneness. We found some evidence based on interviews with developers and an online survey regarding this negative effect on quality.

*[http://www.cs.cmu.edu/~ckaestne/pdf/ecoop15.pdf](http://www.cs.cmu.edu/~ckaestne/pdf/ecoop15.pdf)

Don't you agree with that ?

lws-team:
Well, I did not understand that meaning from your patch.... I guess that's my fault, I looked at the diff but didn't see into it enough.

Yes it's true the preprocessor cutting across the code like that is usually a sign of hackery.

There have been several waves of this kind of thing over the years in lws, that plat- stuff came into existence to clean most of the Windows preprocessor conditionals out, and you can see in this code evidence of another wave around libev, lws_libev... was invented to remove preprocessor conditionals related to libev from the code.

LWS_POSIX is only a few weeks old since I added mbed3 support.  Although that works, it's still incomplete (server only, no extensions), so there will likely be more of it until we can see the whole scope of it.  So the time for the wave to wipe those out of the main code isn't ready yet.

Anyway your patch makes sense now I get your point so I pushed it.  Although as I say properly dealing with all LWS_POSIX will have to wait.  (BTW kernel style is also hard tabs, I fixed the +  if (... whitespace on your patch)

wmarquesr:
So you add my patch to code ?

lws-team:
Yeah that's what I meant by "I pushed it" "22 days ago"

https://github.com/warmcat/libwebsockets/commit/c718ff30d86ccd9e22b4e71eda5bfa8008f30aa8

wmarquesr:
Oh right, sorry for confusion and delay.

PR #372, opened 2015-12-15
Fix corruption of close reason code in close frame

hashok:
According to specification, close reason code is part of body


lws-team:
Thanks, it's pushed

PR #383, opened 2015-12-25
Recognize Debian GNU/kFreeBSD as FreeBSD-ish.

ppentchev:
Some of the FreeBSD-specific code in libwebsockets is related to
the FreeBSD kernel, not the general build environment.  Thus, it is
important to make this distinction, especially when building on
platforms that have a FreeBSD kernel and a non-FreeBSD userland build
environment, such as Debian GNU/kFreeBSD.

When checking for FreeBSD kernel features, also check for the newly
introduced __FreeBSD_kernel__ preprocessor constant; it is present in
the GNU/kFreeBSD kernel and also in FreeBSD itself since the 9.1 release
about three years ago.

### DIFF
--- a/lib/lws-plat-unix.c
+++ b/lib/lws-plat-unix.c
@@ -201,7 +201,9 @@ lws_plat_set_socket_options(struct lws_context *context, int fd)
 	int optval = 1;
 	socklen_t optlen = sizeof(optval);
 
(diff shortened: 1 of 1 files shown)


lws-team:
Thanks, it's pushed.

Issue #419, opened 2016-02-08
make libwebsockets_rpm is broken 


ezerbib:
make libwebsockets_rpm is broken 
....

> Requires: /usr/bin/pkg-config
> Processing files: libwebsockets-debuginfo-1.6.0-1.fc21.x86_64
> Checking for unpackaged file(s): /usr/lib/rpm/check-files /home/ezerbib/work/libwebsockets/libwebsockets/build/RPM/tmp
> error: Installed (but unpackaged) file(s) found:
>    /usr/bin/libwebsockets-test-fuzxy
> 
> 
> RPM build errors:
>     Installed (but unpackaged) file(s) found:
>    /usr/bin/libwebsockets-test-fuzxy
> CMakeFiles/libwebsockets_rpm.dir/build.make:49: recipe for target 'CMakeFiles/libwebsockets_rpm' failed
> make[3]: *** [CMakeFiles/libwebsockets_rpm] Error 1
> CMakeFiles/Makefile2:91: recipe for target 'CMakeFiles/libwebsockets_rpm.dir/all' failed
> make[2]: *** [CMakeFiles/libwebsockets_rpm.dir/all] Error 2
> CMakeFiles/Makefile2:99: recipe for target 'CMakeFiles/libwebsockets_rpm.dir/rule' failed
> make[1]: *** [CMakeFiles/libwebsockets_rpm.dir/rule] Error 2
> Makefile:184: recipe for target 'libwebsockets_rpm' failed
> make: *** [libwebsockets_rpm] Error 2
> 

need to add this line
`/usr/bin/libwebsockets-test-fuzxy`
 in %files section of the RPM/SPECS/libwebsockets.spec

lws-team:
Thanks it's pushed.

ezerbib:
thanks to you !!!
 for supporting the bugs

PR #420, opened 2016-02-09
Packaging fixes

acooks:
* update libwebsocket.spec file to address rpmlint and fedora-review issues. This .spec looks more like the one submitted for review in https://bugzilla.redhat.com/show_bug.cgi?id=1198498
* resolves gh issue #419 and others like it by using wildcard matching in the -devel package to install all libwebsockets* binaries.
* update version in CMakeLists.txt to 1.6.3


lws-team:
Thanks, it's pushed and v1.6.3 tag updated to point to this.

PR #425, opened 2016-02-12
Ask sysconf for Android's getdtablesize.

DanAlbert:
I don't see how to actually build/test this for Android... Will travis take care of it for me?

### DIFF
--- a/lib/libwebsockets.h
+++ b/lib/libwebsockets.h
@@ -163,7 +163,8 @@ struct sockaddr_in;
 #endif
 
 #if defined(__ANDROID__)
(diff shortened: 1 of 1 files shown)


lws-team:
Thanks, it's pushed.

Although lws works on a lot of platforms the ones that I work on are Linux and mbed3.  We have had travis and appveyor build CI for Linux, OSX, and Windows for something like a couple of years.  We know people use it on iOS and Android and more specialized platforms like OpenWRT.

But we don't have a direct way to build or do CI for Android (any contribution about that would be very welcome).

PR #431, opened 2016-02-20
Don't destroy ev/uv loops if they haven't been created yet

alexhultman:
This affects cases where the server is not able to create a context due to some error, like not being able to listen to the port. Before it would destroy uninitialized loops and most probably crash.


lws-team:
Thanks, it's pushed.

This looks like it fixes an error path, which is good.

But the mysterious (apparently one-per-session) 128 or 256 byte valgrind leak is still there with libuv on exit of the libuv test server.

PR #432, opened 2016-02-20
Make sure every handle is closed before destroying the uv loop

alexhultman:
I don't know why there are two commits but yeah. This closes every left handles by using uv_walk to iterate over them and uv_close to close them.

http://stackoverflow.com/questions/25615340/closing-libuv-handles-correctly

We could probably optimize this a little better in the future but for now it closes down with no Valgrind issues. It should be possible to debug what handles are still open by printing the address of the uv_handle_t and compare it to any handles we create and then solve it in other ways.


alexhultman:
Okay wait let me rebase it.

PR #434, opened 2016-02-20
Stop uv loop in default signal handler, clean-ups

alexhultman:
I think this is the last obvious things left.


lws-team:
Also pushed on master / v1.7-stable thanks.

PR #443, opened 2016-02-27
Don't include <sys/cdefs.h> for NetBSD

krytarowski:
We needed it for the BSD symbol to be defined, while __NetBSD__ is defined
with a compiler.

Thanks Andy Green for the initial fix.

Signed-off-by: Kamil Rytarowski <n54@gmx.com>


lws-team:
It's pushed thanks.

On platforms I can't test on, I am basically on a loser trying to fix things.  I appreciate the patch tested on the platform ^^

PR #445, opened 2016-02-29
WinCE compatibility

ondraco:
Added few changes which makes compiling on windows easier (v1.7-stable should be compilable with no problem).


lws-team:
Thanks, I squashed these into one along with by initial attempt, and replaced my original attempt.

https://github.com/warmcat/libwebsockets/commit/a3c419d5793dd7ceb1bba9e679d9c1475fc42003

PR #446, opened 2016-02-29
V1.7 stable - WinCE compatibility

ondraco:
Added few changes which makes compiling on windows easier (v1.7-stable should be compilable with no problem).

### DIFF
--- a/lib/libwebsockets.c
+++ b/lib/libwebsockets.c
@@ -946,13 +946,16 @@ LWS_VISIBLE void lwsl_emit_stderr(int level, const char *line)
 #endif
 	int n;
 
+#ifndef _WIN32_WCE
 #ifdef WIN32
 	ptm = localtime(&o_now);
 #else
 	if (localtime_r(&o_now, &tm))
 		ptm = &tm;
 #endif
+#endif
 	buf[0] = '\0';
+
 	for (n = 0; n < LLL_COUNT; n++) {
 		if (level != (1 << n))
 			continue;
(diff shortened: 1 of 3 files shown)


lib/libwebsockets.c
```diff
@@ -308,6 +308,9 @@ lws_close_free_wsi(struct lws *wsi, enum lws_close_status reason)
 		if (n)
 			lwsl_debug("closing: shutdown ret %d\n", LWS_ERRNO);
 
+// This causes problems with disconnection when the events are half closing connection
+// FD_READ | FD_CLOSE (33)
+#ifndef _WIN32_WCE
```
JoakimSoderberg: POSIX + WIN32_WCE ? :)

ondraco: Seems so it is just defined as:...

```
#ifdef MBED_OPERATORS
struct sockaddr_in;
#define LWS_POSIX 0
#else
#define LWS_POSIX 1
#endif
```

lws-team: No you are right, POSIX means that it's a platform that follows posix semantics for its apis, windows basically does that.

On the other hand mbed3 is a case where the operating system itself is a single global event loop.  It has some posix-sounding apis but they are fundamentally different.

lws-team:
Thanks.  Unless there's a big divergence in some contributed code, I can deal with adapting patches between master and v1.7-stable just as part of normal business.  I just need a patch on one or the other.

If I missed the point and there's something different here, let me know.

PR #449, opened 2016-03-02
libuv.c: Service fd with LWS_POLLHUP on poll errors

alexhultman:
This fixes my problems with stuck connections:

```
^C[alexhultman@localhost node-lws]$ node examples/echo.js
Running echo server on port 3000
Connections: 1
Connections: 2
^C[alexhultman@localhost node-lws]$ node examples/echo.js
Running echo server on port 3000
Connections: 1
Connections: 0
Connections: 1
Connections: 0
Connections: 1
Connections: 0
Connections: 1
Connections: 0
```

Instead of just returning on errors, we now service the fd with LWS_POLLHUP on all errors except for UV_EAGAIN (to show that it is possible to filter some errors). It also renames the wrongly named "accept" callback.

### DIFF
--- a/lib/libuv.c
+++ b/lib/libuv.c
@@ -31,26 +31,36 @@ lws_feature_status_libuv(struct lws_context_creation_info *info)
 }
 
 static void
-lws_accept_cb(uv_poll_t *watcher, int status, int revents)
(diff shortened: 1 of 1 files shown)


lws-team:
Thanks it's pushed.

PR #460, opened 2016-03-15
Well ,When I create a connection  no callback LWS_CALLBACK_WSI_CREATE

bjqiwei:
I added this callback, but I do not know whether or not in the right place


lws-team:
Thanks... it's pushed on both master and v1.7-master.

PR #462, opened 2016-03-16
Optimize payload exhaustion

alexhultman:
Hi.

This optimization bypasses the parser's per-byte strategy when in state LWS_RXPS_PAYLOAD_UNTIL_LENGTH_EXHAUSTED. This makes receiving large amount of data about 10x faster. We can test it more but it seems to work pretty okay. The state it handles really only does copying or copying and unmasking (if I understand it correctly?). I leave 1 byte left for the real parser so that proper state transition can happen using the old code.

### DIFF
--- a/lib/parsers.c
+++ b/lib/parsers.c
@@ -1455,3 +1455,53 @@ lws_remaining_packet_payload(struct lws *wsi)
 {
 	return wsi->u.ws.rx_packet_length;
 }
+
+/* bypass slow per-byte parsing of payload data when in
+ * parser state LWS_RXPS_PAYLOAD_UNTIL_LENGTH_EXHAUSTED */
+
+void
+lws_payload_until_length_exhausted(struct lws *wsi, unsigned char **buf, size_t *len)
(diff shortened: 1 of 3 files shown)


alexhultman:
Do you want it without restrict pointers or should we fix the building to use C11?

alexhultman:
I did a force update with the removed restrict pointers. It has roughly (if not exactly) the same performance. It could be optimized about 2x more but I didn't want to go ape crazy.

lws-team:
Thanks... I modified it a bit to do loop unrolling... it seems difficult to optimize it to use 32-bit reads and writes that would be faster because there's no alignment requirement on either buffer.  I confirmed it with the test server but please check I didn't break anything.

alexhultman:
Nice, it seems to work.

alexhultman:
That unrolling is really effective. Profiling shows a lot less CPU usage and *the benchmark* is the bottle neck now.

lws-team:
Great.

PR #463, opened 2016-03-17
uv: Treat all errors as hangup, except for UV_EAGAIN

alexhultman:
This fixes an error where sending a large buffer took a lot of time to finish. I didn't properly handle UV_EAGAIN.

Edit: Wow, wait - I think this is wrong.

### DIFF
--- a/lib/libuv.c
+++ b/lib/libuv.c
@@ -42,14 +42,8 @@ lws_io_cb(uv_poll_t *watcher, int status, int revents)
 	eventfd.events = 0;
 	eventfd.revents = 0;
 
-	if (status < 0) {
-		/* at this point status will be an UV error, like UV_EBADF,
-		we treat all errors as LWS_POLLHUP */
-
-		/* you might want to return; instead of servicing the fd in some cases */
-		if (status == UV_EAGAIN)
-			return;
-
+	/* treat all errors as LWS_POLLHUP, except for UV_EAGAIN */
+	if (status < 0 && status != UV_EAGAIN) {
 		eventfd.events |= LWS_POLLHUP;
 		eventfd.revents |= LWS_POLLHUP;
 	} else {


alexhultman:
Need to fix this some more.

alexhultman:
Never mind this. UV_EAGAIN never happens it seems.

lws-team:
Is the problem you're seeing maybe related to generic lws timeouts?  You can try increasing the timeout value (at the context creation info .timeout_secs) and see if it changes the behaviour.

alexhultman:
I have two problems: sending a large buffer (100 mb) doesn't work at all when using perMessageDeflate (client disconnects), it does work without it but I'm pretty sure I have seen (sometimes) the sending time out (60 seconds). I should debug it and see what's up.

alexhultman:
The client connects, sends a 100 mb message with permessage deflate on and it gets disconnected. It works without permessage deflate. Is there any limit?

alexhultman:
Regarding the sending without permessage deflate, I found what makes it work and not work. If I send a large buffer with (any) set rx_buffer_size it works. If it is unset, it simply times out.

lws-team:
No the pm deflate stuff works on streams for both inflate and deflate; it can pass large dataset tests with Autobahn using pm deflate so I think it's pretty stable generally.

Not setting the rx buffer size chooses 4KB for you... what are you setting it to when you get different behaviour?

alexhultman:
For me, anything below 35 kb takes 63 seconds while having 35 or larger rx_buffer_size sends the buffer back in less than a second.

lws-team:
I guess this is on 127.0.0.1 where the network interface will accept large things in one step.

But I don't have much idea where the 63s is coming from.

I guess this is not related to pm deflate, because by default, the inflate and deflate stream buffers are set to 1K

		priv->args[PMD_RX_BUF_PWR2] = 10; /* ie, 1024 */
		priv->args[PMD_TX_BUF_PWR2] = 10; /* ie, 1024 */

you can override these options from user code.

alexhultman:
The pm deflate thing is probably a different issues yes, it might just be invalid settings or something I don't know much about that. The client just gets disconnected directly when it tries to send.

But this problem where sending a large buffer seems to be an issue because it really does take about 60 second each time the rx_buffer_size is too small and the client actually does get the full message by that time.

I could solve it by only sending fragments but the lws_write should accept large buffers, right? It is on localhost.

alexhultman:
I'm basically requesting a writable callback, using lws_write to pass it a 100 mb buffer and then I don't do anything else in my code. I do not get any uv errors (which is what I tried to "solve" in this pull request but I was totally wrong because there is no errors happening in this case at all).

lws-team:
lws_write() is a wrapper for send() on the socketfd.

Actually it's up to the kernel how much data he will accept on that connection, he can accept nothing, something or all of it.  If less than the requested amount was accepted, lws will malloc up a buffer and hold the rest, returning immediately.

If so, in the background he disables the WRITABLE callback and sends the buffered data transparently, allowing WRITABLE again when it is done.

If you pass it 100MB buffer in one step, the kernel will never accept that from send().  So you are into this malloc / memcpy thing, a huge one too.  And this will never work reasonably outside of 127.0.0.1.

You should send a chunk at a time where the chunk has a good chance of being accepted wholly by send(), typ 2K or 4KB if the connection is being kept with data in flight that is only relatively slowly being Acked, and you get signalled it's WRITEABLE as soon as anything was acked to top it back up again.  That way you avoid this "malloc / send what the kernel didn't accept in the background" thing.

alexhultman:
Yep I saw you did that change a couple of months or weeks ago. But I just tried this with the ev backend and I got the exact same behavior. So there is nothing wrong with the uv backend I think. Ofc, this is just an insane test on localhost but it really does seem to be a flaw in the lws_write since I get the same behavior with both backends.

I do have a solution that chops the message up in fragments and sends it using the fragmented message support but I wanted to make sure the lws_write worked properly as an alternative solution.

lws-team:
Yes you should use the fragmented message stuff, this isn't about the event loop it's about send() and the kernel.  lws_write() is basically send() with this emergency recovery scheme wrapped around it... you couldn't get anything good to happen trying to do that with send() directly either.

The basic problem under this is we only get signalled that the connection may be written to, there is no indication of how much send() will accept, and it varies dynamically with memory pressure.  So you can get good results if all or most send()s are fully accepted by the kernel, and deal with unexpected incomplete send()s automatically.  Using the fragmentation scheme lets you do that nicely while also allowing interleaved control packets etc on the connection.

alexhultman:
Yep, I get the fact that the internal lws_write buffering is bad (it requires a massive copy, etc). Personally I would *always* use fragmented messages, but I opened a bug for this because even if it is not recommended, the internal buffering fallback seems to be broken.

lws-team:
With the test server leaf.jpg, I can force it to do buffering with this

```
diff --git a/lib/output.c b/lib/output.c
index 0d850eb..72b1be9 100644
--- a/lib/output.c
+++ b/lib/output.c
@@ -172,7 +172,7 @@ handle_truncated_send:
         * Newly truncated send.  Buffer the remainder (it will get
         * first priority next time the socket is writable)
         */
-       lwsl_info("%p new partial sent %d from %d total\n", wsi, n, real_len);
+       lwsl_err("%p new partial sent %d from %d total\n", wsi, n, real_len);
 
        /*
         *  - if we still have a suitable malloc lying around, use it
diff --git a/test-server/test-server-http.c b/test-server/test-server-http.c
index 046f390..1c03749 100644
--- a/test-server/test-server-http.c
+++ b/test-server/test-server-http.c
@@ -118,7 +118,7 @@ int callback_http(struct lws *wsi, enum lws_callback_reasons reason, void *user,
 {
        struct per_session_data__http *pss =
                        (struct per_session_data__http *)user;
-       unsigned char buffer[4096 + LWS_PRE];
+       unsigned char buffer[40960 + LWS_PRE];
        unsigned long amount, file_len, sent;
        char leaf_path[1024];
[agreen@warmcat build]$ git diff | cat
diff --git a/lib/output.c b/lib/output.c
index 0d850eb..72b1be9 100644
--- a/lib/output.c
+++ b/lib/output.c
@@ -172,7 +172,7 @@ handle_truncated_send:
 	 * Newly truncated send.  Buffer the remainder (it will get
 	 * first priority next time the socket is writable)
 	 */
-	lwsl_info("%p new partial sent %d from %d total\n", wsi, n, real_len);
+	lwsl_err("%p new partial sent %d from %d total\n", wsi, n, real_len);
 
 	/*
 	 *  - if we still have a suitable malloc lying around, use it
diff --git a/test-server/test-server-http.c b/test-server/test-server-http.c
index 046f390..1c03749 100644
--- a/test-server/test-server-http.c
+++ b/test-server/test-server-http.c
@@ -118,7 +118,7 @@ int callback_http(struct lws *wsi, enum lws_callback_reasons reason, void *user,
 {
 	struct per_session_data__http *pss =
 			(struct per_session_data__http *)user;
-	unsigned char buffer[4096 + LWS_PRE];
+	unsigned char buffer[40960 + LWS_PRE];
 	unsigned long amount, file_len, sent;
 	char leaf_path[1024];
 	const char *mimetype;
```

and see it work fine between two machines (and see it actually did trigger the buffering)

alexhultman:
Yep it seems to be a non-issue for a few people. But I'll write more as I debug in the issue. #464

Issue #464, opened 2016-03-18
Sending a large buffer with lws_write takes 63 seconds in some cases


alexhultman:
I think this needs investigation. Like I mentioned, both the ev and uv backends behave identical in this regard and that is when sending a 100mb buffer with lws_write, it can take less than a second in the case of rx_buffer_size being large enough (larger than or equal to 33 kb in my specific case). In cases where rx_buffer_size is smaller, or unset (4kb) it takes 63 seconds to complete.

alexhultman:
The fact that lws_write now makes internal buffering instead of refusing to write more than what the kernel allows makes it hard to solve by using fragmented sends. When I pass 100 mb to lws_write it returns that it did accept all bytes:

lws_write accepted: 104857600

Still, it doesn't actually manage to send these bytes in a reasonable time.

alexhultman:
When it fails, I get bursts of lws_service_fd_tsi calls, then a long wait, then a new burst, etc. lws_service_fd_tsi is not being called as often as it should have.

lws-team:
I don't really understand why the rx buffer size is involved.  However as I wrote in #463 lws_write() is a wrapper around send().  You cannot call send() with a 100MB buffer one time either and expect it to get anywhere.

The malloc() thing is a safety-net for the case that sometimes the kernel won't accept all of a reasonable buffer size like 4KB, it's not intended to enable this 100MB usage thing.

It'd help me a lot if you can hack the test case on the normal test server so I can hopefully see what you're seeing.

alexhultman:
Do you have any plans on adding a flag to lws_write so that one can explicitly say "no, do not buffer things up if you couldn't fit in the kernel buffer". Because currently I have no way of knowing how much I can send without triggering the built-in buffering. I want to explicitly disable it so that I can be sure my fragments are directly sent to the kernel, and I do not want to guess.

alexhultman:
If you look at the old documentation: https://libwebsockets.org/libwebsockets-api-doc.html

lws_write is said to return the number of bytes *actually* sent. Now, since the internal buffering change (which I personally cannot get to work reliably) there is no way of knowing how many bytes were written, since it now accepts all bytes and does the internal buffering.

I want to disable the internal buffering so that I can get the old behavior where you manually have to do the fragmented send.

alexhultman:
[...]

PR #465, opened 2016-03-18
Do not send longer fragments than rx_buffer_size

alexhultman:
I'm just posting this one, you can test it and merge later if it works.


alexhultman:
Solves #464

Issue #469, opened 2016-03-22
libuv initloop signal handling


dosvald:
Hi,

in the libwebsockets-libuv code, there is a function `lws_uv_sigint_cfg` which should allow disabling signal handler registration (via `use_uv_sigint` parameter) as well as custom signal handler parameter. I tried those and they didn't work.

Digging through lib/libuv.c i spotted a couple related problems with libuv signal handling code:

1. the `lws_uv_initloop` doesn't check the `use_ev_sigint` (= `use_uv_sigint`) flag as the libev counterpart does, before setting signal handlers. This impacts the use of libwebsockets in existing/foreign loop.

2. the `lws_uv_initloop` function takes a callback parameter itself, and does not use the one set by `lws_uv_sigint_cfg`, making the latter function currently useless as far as I could tell. The callback parameter is of different type, is not needed and sigint_cfg should be used for that.

3. there is a typedef `lws_uv_signal_cb_t` which is used in above functions, but I don't see its purpose. First, it has a `int revents` parameter for which I don't see how it can be related to a signal like SIGINT. Second, it has parameters uv_loop_t and uv_signal_t. These can be collapsed into just uv_signal_t, since it is a handle, and has a public member pointing to its loop: http://docs.libuv.org/en/v1.x/handle.html#c.uv_handle_t.loop. The typedef then becomes identical to libuv's `uv_signal_cb` and is thus not of any use.

4. there is a function lws_uv_sigint_cb which is set as the default sigint action (probably meant to be used unless overridden by `lws_uv_sigint_cfg` function). It conforms to the typedef in the previous item, and just runs uv_stop on the loop. Most likely it should run lws_libuv_stop instead., but I'm not certain about this. If lws_libuv_stop is the right one, the lws_context should be passed as a user pointer (member named `data` in signal `uv_handle_t`) to it.

Aside from changing (in my opinion) broken behavior, the proposed fixes would break API:
- `lws_uv_initloop` loses the callback parameter (should be set via _cfg function)
- the typedef is changed to be plain `uv_signal_cb`, or removed in favor of it

On the other hand, the API can be maintained (keeping the function signatures and typedef):
- for the initloop function. ignore the parameter
- for the redundant loop argument in typedef: a libuv-compatible callback wrapper has to be made, and a context structure would have to be allocated for each signal handler (per-thread) which would just store two pointers: lws_context, and the wrapped callback. The wrapper would call the wrapped callback with redunntant parameters...
[...]

PR #470, opened 2016-03-22
uv: fix sigint handling

dosvald:
See https://github.com/warmcat/libwebsockets/issues/469

First commit fixes behavior while keeping function signatures same (add cb wrapper to preserve API)
Second commit removes callback wrapper, cleaning up API and removing wrapper.

### DIFF
--- a/lib/libuv.c
+++ b/lib/libuv.c
@@ -66,7 +66,7 @@ lws_io_cb(uv_poll_t *watcher, int status, int revents)
 }
 
 LWS_VISIBLE void
-lws_uv_sigint_cb(uv_loop_t *loop, uv_signal_t *watcher, int signum)
+lws_uv_sigint_cb(uv_signal_t *watcher, int signum)
 {
 	lwsl_info("internal signal handler caught signal %d\n", signum);
 	lws_libuv_stop(watcher->data);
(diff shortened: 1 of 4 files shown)


dosvald:
The appveyor build seems to have failed due to network error:
````
Downloading nsis.install 32 bit
   from 'http://sourceforge.net/projects/nsis/files/NSIS%202/2.50/nsis-2.50-setup.exe/download'
 Exception calling "GetResponse" with "0" argument(s): "Unable to connect to the
  remote server"
````

Please try restarting if possible.

**Comment by lws-team** [comment]
Thanks... I have checked them against the libuv test server (especially ^C there...) and they're fine so pushed them on master.  I looked at them from v1.7-stable point of view but there are some type changes, I think we have to leave that.

@alexhultman please reopen if it makes trouble for you.

**Comment by dosvald** [comment]
> I looked at them from v1.7-stable point of view but there are some type
changes, I think we have to leave that.

For 1.7, have you checked the first commit in the PR, ? That one doesn't
change type signatures, so maybe cherry pick that one for 1.7 ?

**Comment by lws-team** [comment]
This kind of change in the libuv test server -->

```
-void signal_cb(uv_signal_t *watcher, int revents)
+void signal_cb(uv_loop_t *loop, uv_signal_t *watcher, int signum)
 {
        lwsl_err("Signal %d caught, exiting...\n", watcher->signum);
        switch (watcher->signum) {
@@ -287,7 +287,9 @@ int main(int argc, char **argv)
                return -1;
        }
 
-       lws_uv_initloop(context, NULL, signal_cb, 0);
+       lws_uv_sigint_cfg(context, 1, signal_cb);
```

requires an soname bump in the library, since we're saying the user application has to do different things depending on whehter it meets <= 1.7.4 or afterwards: all the v1.7-stable are defined to have the same soname.  So it doesn't fit on v1.7-stable.

But there will be a new release based off master in the next few weeks including the changes, so it shouldn't be a problem for long.

dosvald:
Hmm, you are right, thanks!

PR #476, opened 2016-03-29
Fix build with musl libc

joerg-krause:
Fix building libwebsockets with the musl C libary.

<sys/cdefs.h> is an internal glibc header and should be avoided in user code.

__P() was used for compatibility with some old K&R C compilers, when there were
no prototypes (which were introduced to C with C89). As supporting legacy
non-ANSI compilers is nowadays not necessary anymore get rid of the unnecessary
function prototype using __P().


lws-team:
Thanks... yes both just seemed to be cruft.  It's pushed on v1.7-stable and master.

Issue #359, opened 2015-12-04
UNIX domain socket support


bprodoehl:
I'm adding support for UNIX domain sockets, so that the server can listen on a sock file rather than a network port.  Would something like this be welcome upstream?  If so, are there any preferences on how that should be implemented?  The options in my mind would be to either add a field for a sock file path to lws_context_creation_info, or to add a field there for an already-created server socket file descriptor.  The difference there being that in the former, the job of creating the UNIX domain socket lives in libwebsockets, and in the latter, that job belongs to the server implementation.

lws-team:
Yes it'd be welcome thanks.

Client may also want to do this eventually so passing the path is better I think.  The guy listening on it should create it, unless I missed the point somewhere.

If it sounds okay to you, you can declare a new enum lws_context_options flag to tell you want unix domain, and re-use the existing info struct member const char *iface for the path -- unix domain has no network interface concept.

One quirk you should protect the implementation with #if LWS_POSIX since we also support mbed3 now that builds the server bits but has no use for this.

lws-team:
Please reopen this or send a PR when you have something.

bprodoehl:
@lws-team I've been working on this here: https://github.com/bprodoehl/libwebsockets/commits/unix_domain_sockets

Not ready for a PR, but I'll send one once I feel it is.  Thanks for the advice on how to integrate it.

PR #490, opened 2016-04-12
Unix domain socket support

yjroot:
This PR adds an option to create a server to UNIX domain socket.
It also supports abstract namespace, by prepending an '@' symbol path name.(just linux)
Refs #359


lws-team:
Very nice... it's pushed thanks.

PR #491, opened 2016-04-13
Add support for client certificates and Certificate Revocation List to the test applications.

ABruines:
As a followup to issue #489 I have written this patch for the server and client test applications.
For the test server:
 - Added option --ssl-verify-client (-v) to enable verification of client certificates.
 - Added option --ssl-crl (-R) to load a CRL to check for revoked client certs (if LWS_OPENSSL_SUPPORT is defined).

For the test client:
 - Added option --ssl-cert (-C) so a client certificate can be specified.
 - Added option --ssl-key (-K) so the client certificate key can be specified.
 - Added option --ssl-ca (-A) to load a CA certificate
 - Added option --ssl-crl (-R) to load a CRL to check for a revoked server cert (if LWS_OPENSSL_SUPPORT is defined).

### DIFF
--- a/test-server/test-client.c
+++ b/test-server/test-client.c
@@ -35,10 +35,17 @@
 
 #include "../lib/libwebsockets.h"
 
+#ifdef LWS_OPENSSL_SUPPORT
+#include <openssl/err.h>
+#endif
+
 static int deny_deflate, deny_mux, longlived, mirror_lifetime;
 static struct lws *wsi_dumb, *wsi_mirror;
 static volatile int force_exit;
 static unsigned int opts;
+#ifdef LWS_OPENSSL_SUPPORT
+static char crl_path[1024] = "";
+#endif
 
 /*
  * This demo shows how to connect multiple websockets simultaneously to a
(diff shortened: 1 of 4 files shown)


lws-team:
Thanks.  I fixed the travis / mac build problem in cmake (his openssl does not have an api you are using) and pushed it.

ABruines:
Very minor fix: `LWS_CALLBACK_OPENSSL_PERFORM_CLIENT_CERT_VERIFICATION` does not need `LWS_HAVE_SSL_CTX_set1_param` in test-server.http
```
diff --git a/test-server/test-server-http.c b/test-server/test-server-http.c
index 56e2206..8f02d04 100644
--- a/test-server/test-server-http.c
+++ b/test-server/test-server-http.c
@@ -648,7 +648,8 @@ bail:
 			}
 		}
 		break;
-
+#endif
+#if defined(LWS_OPENSSL_SUPPORT)
 	case LWS_CALLBACK_OPENSSL_PERFORM_CLIENT_CERT_VERIFICATION:
 		/* Verify the client certificate */
 		if (!len || (SSL_get_verify_result((SSL*)in) != X509_V_OK)) {

```

lws-team:
Thanks... I fixed it in the patch itself by swapping those two cases and doing it with nested #if defined(... like the others, since it's easy to see the relationship that way.

PR #492, opened 2016-04-17
Fix inet_ntop linking error on Windows

lucasterra:
This fixes libwebsockets not building correctly on Windows, because it couldn't link to inet_ntop


lws-team:
Thanks, it's pushed.

PR #494, opened 2016-04-18
Script and config to make tracking ABI/API changes easy.

ralight:
This lets you very easily generate a timeline of how the library ABI has changed - useful for users wanting to know how to update their code and for developers to check they haven't broken compatibility.

Example output is at http://atchoo.org/abi/timeline/libwebsockets/

### DIFF
--- a/contrib/abi/README.md
+++ b/contrib/abi/README.md
@@ -1,0 +1,48 @@
+ABI Compatility Tracking
+========================
+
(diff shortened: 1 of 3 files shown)


lws-team:
Thanks... that's a pretty cool project.  I edited a couple of things and pushed it.

I also integrated it on a cronjob at the libwebsockets.org server, it's here

https://libwebsockets.org/abi/timeline/libwebsockets/index.html

PR #495, opened 2016-04-18
Conditionally define snprintf on Windows

lucasterra:
snprintf is already defined in MSVC 2015. If you redefine it <stdio.h> will cry about it and abort the compilation.


lws-team:
Thanks, it's pushed.

PR #498, opened 2016-04-21
Fix build on FreeBSD

hgrundy:
This just adds FreeBSD to the list that requires the netinet/in.h include, so it'll build.


lws-team:
Thanks, it's pushed

PR #505, opened 2016-04-26
Fixes for building with MinGW

stephaneberle9:
Added fixes to avoid compile errors and warnings when building under
Windows using MinGW

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,7 @@ if(GIT_EXECUTABLE)
     OUTPUT_VARIABLE GIT_HOST
     OUTPUT_STRIP_TRAILING_WHITESPACE
     )
+	string(REGEX REPLACE "([^\\])[\\]([^\\])" "\\1\\\\\\\\\\2" GIT_USER ${GIT_USER})
     set(LWS_BUILD_HASH ${GIT_USER}@${GIT_HOST}-${GIT_HASH})
     message("Git commit hash: ${LWS_BUILD_HASH}")
 endif()
--- a/lib/lws-plat-win.c
+++ b/lib/lws-plat-win.c
@@ -541,7 +541,6 @@ _lws_plat_file_read(struct lws *wsi, lws_filefd_type fd, unsigned long *amount,
 {
 	DWORD _amount;
 
-	(void *)wsi;
 	if (!ReadFile((HANDLE)fd, buf, (DWORD)len, &_amount, NULL)) {
 		*amount = 0;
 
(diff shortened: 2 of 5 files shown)


lws-team:
Thanks, they're all pushed.

Actually I have a question about MinGW... I may have to provide lws under Windows myself, if so, avoiding dealing with msvc for what's basically a couple of headless services sounds good.

I was looking at MinGW yesterday from that point of view and I saw it has split into mingw-w64 and the original mingw.  Do you have any advice about choosing between them, or the difficulty level of trying to avoid msvc in windows-land (for ancilliary things like installer)?

kangjoni76:
Hi, sorry bothering here. the mingw build download you can pick here https://qa.debian.org/watch/sf.php/mingw-w64 it is alternative from Debian maintener rather than directly using sf.net site adware bloated. the posix thread is mostly downloaded because of <thread.h> compat and use dwarf 32 bit and seh exception 64 bit. for msvc using monthly build you can download only 200mb~ complete cl.exe toolset via nuget here https://blogs.msdn.microsoft.com/vcblog/2016/04/26/stay-up-to-date-with-the-visual-c-tools-on-nuget/

lws-team:
Thanks... Fedora that I use also has mingw32 + 64 packages and related mingwX-built packages.  So I think it can be done.

I am really more asking is that realistic to expect to be able to provide normal installable (headless) executables on windows using those tools.  Since I can do almost everything else with almost no windows-specific code, I am wondering if I can get through the whole thing by almost completely ignoring windows....

 - no msvc
 - build on Linux using Fedora mingwX
 - end up with installable services for windows (ie, lwsws)
 - test them on VM

It seems like it should be OK but actually I have no idea about how to, eg, wrap them in an installer via mingwX.

lws-team:
Hm there's mingw32-nsiswrapper + mingw32-nsis in Fedora too.  I saw we use them as an installer in Appveyor somehow.

lws-team:
Closing as the PR itself is done, I'll try the Fedora/mingw32/nsis road if it comes to it.

stephaneberle9:
Hi,

Thanks a lot for accepting and merging my changes.

Regarding your question about the Windows installer: I wasn't aware of the mingw32-nsiswrapper + mingw32-nsis so far. In the mbeddr project (https://github.com/mbeddr), where I'm active, we have separate repo (https://github.com/mbeddr/mbeddr.installer.win) which is dedicated to providing a Windows installer for the mbeddr binaries and all required third-party dependencies (including MinGW). This is basically a traditional NSIS installer script. We build it on a TeamCity server (https://build.mbeddr.com/project.html?projectId=Mbeddr2_Mbeddr_Package_Win&tab=projectOverview) that runs this build on a Windows machine. Nevertheless, the build result, i.e., the win installer package, is published via GitHub (https://github.com/mbeddr/mbeddr.core/releases).

Btw, my interest in libwebsockets is also related to mbeddr. I'm actually working on an mbeddr wrapper for libwebsockets that enables users to integrate libwebsockets in mbeddr applications and program against it through mbeddr's advanced components/interface language.

PR #507, opened 2016-04-29
Add warn_unused_result check, attribute only supported by GCC 3.4 or …

pfarrell85:
…later.

warn_unused_result was introduced in GCC version 3.4.

Change-Id: I6c2cc938d2b868ddfe0889cc41d7fa9d70e1b907


lws-team:
Thanks, it's pushed on v1.7-stable and master.

PR #511, opened 2016-05-05
Win CE support

ondraco:
Hello, I have just updated to LWS 2.0 and since I have to support also WinCE i added few modifications. 

1. lib/lws-plat-win.c 
`time_t time(time_t *t)`
This was surely a bug which was causing a crash when new context was created and `time(NULL)` was called.

2. lib/libwebsockets.h
There is no `<stdint.h>` on WCE but it seems to me that it is not needed at all (i may be possibly wrong).  On WCE and in MSVC older than 2015 there is only `_stricmp`.

3. private-libwebsockets.h
 On WCE and in MSVC older than 2015 there is only `_vsnprintf`.

4. server.c
 Here is a problem that WCE does not have `struct stat` at all. This is not much of a fix just a workaround that I use to avoid the problem. I would like to ask if there is any problem with this part of code disabled on WCE. If i understood it right it takes care of serving symlinks and http expiration. If i personally don't need those features it should not affect anything else right?

### DIFF
--- a/lib/libwebsockets.h
+++ b/lib/libwebsockets.h
@@ -114,7 +114,6 @@ struct sockaddr_in;
 #include <stddef.h>
 #include <basetsd.h>
 #ifndef _WIN32_WCE
-#include <stdint.h>
 #include <fcntl.h>
 #else
 #define _O_RDONLY	0x0000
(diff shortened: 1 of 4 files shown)


lws-team:
Thanks... but removing stdint.h kills test-ping in Appveyor

```
C:\projects\libwebsockets\test-server\test-ping.c(81): error C2061: syntax error: identifier 'uint64_t' [C:\projects\libwebsockets\build\test-ping.vcxproj]
```

lws-team:
I simply skipped the stdint.h removal patch since it's actually already protected by #ifndef WCE, and pushed the rest on master + v2.0-stable.  Thanks again.

PR #512, opened 2016-05-05
lib/ssl.c: fix libre- and boringssl

Gottox:
This PR fixes support for libreSSL and BoringSSL.


lws-team:
Thanks, it's pushed on master + v2.0-stable.

PR #513, opened 2016-05-05
lib/server.c: fix ipv6 support

Gottox:
fixes compile error when IPv6 is enabled.


lws-team:
Thanks, it's pushed on master + v2.0-stable.

PR #527, opened 2016-05-13
Fix leak caused by undestroyed pthread mutex

sterlingjensen:
On freebsd valgrind reports 80 bytes definitely lost by lws_pt_mutex_init, this fixes that.


lws-team:
Thanks, it's pushed on master, v2.0-stable and v1.7-stable.

PR #543, opened 2016-05-28
Android test client

ABruines:
Hi, I would like to submit this code to the project.
It is an example of how libwebsockets can be used with Android and
provides a test application that can connect to the libwebsockets test server


lws-team:
Thanks a lot, I am sure people will find that helpful.

It's pushed on master.

PR #549, opened 2016-06-01
revert changes in daemonize.c from commit 22d6f39e7fb270d0c011bae820d…

mdrohmann:
Lock file creation stops working without the given changes.

The master process still returned with exit code 0 and the daemon process existed, but the lockfile was never written to disk...

Thanks,
Martin


lws-team:
Thanks.... I took too much out when getting it working as a systemd service.  It's pushed on master + v2.0-stable

Issue #553, opened 2016-06-03
IPv6 doesn't bind to interface address


dosvald:
In `lws_socket_bind` the IPv6 path binds to `in6addr_any` instead of the address and doesn't use the `iface` argument at all.

lws-team:
Bro do we really need three issues for the same problem?

dosvald:
This might be as easy as calling `interface_to_sa` in the IPv6 case of `lws_socket_bind` which, although it gets a `sockaddr_in` with ipv4 address, it uses it as `sockaddr_in6` if ipv6 enabled, so it might work. But I don't see that code path being tested anywhere so it might not work after all.

lws-team:
I have pushed what should be the per-vhost ipv6 disable for server.

But it's 21:30 here... you'll find the rest quicker to solve yourself and send me a patch.

dosvald:
I'll be happy to do so. Thanks again!

Issue #552, opened 2016-06-03
IPv6 support should be per-vhost, not per-context


dosvald:
Let's say I want to make this scenario with 1 context and couple of vhosts:

- vhost A wants to listen on 0.0.0.0:4444, doesn't want to touch IPv6
- vhost B wants to listen on *:6464, using both IPv6 and IPv4 (via mapping), or using IPv6 only

Of course, if  LWS_IPV6=OFF compile-time, vhost A is only possible option, and works as expected by default (no options need to be set), and vhost B can't be supported.
But when having IPV6 on compile-time, how to achieve the vhost cases like above?

I've tried and failed. and I now think that IPv6 handling is wrong at least in vhost / server part, or at least It isn't obvious what to do. I've tried the logical stuff, disabling IPv6 on some vhost but that didn't work.

Then I saw in server.c `lws_context_init_server` ( called from create vhost function ):

```c
#ifdef LWS_USE_IPV6
	  if (LWS_IPV6_ENABLED(vhost->context))
 		sockfd = socket(AF_INET6, SOCK_STREAM, 0);
 	else
#endif
		sockfd = socket(AF_INET, SOCK_STREAM, 0);
```

That looks like it checks if **context** enables IPv6 when doing vhost init, but since socket belongs to vhost, it should check that instead. That means that currently one lws context can make all vhosts either type A above (4 only, if ctx info disables IPv6) or type B (if ctx info doesn't enable IPV6), and I believe that's a defect.

So, the vhost creating part should check the vhost ipv6 support when making socket.

Then user code would do something like this:

```c
// don't disable IPv6, since some vhosts need it (analog to SSL, need to enable if some vhosts need it)
ctxinfo->options &= ~LWS_SERVER_OPTION_DISABLE_IPV6;
...
// For vhost A disable IPv6
vhinfoA->options = LWS_SERVER_OPTION_DISABLE_IPV6
....
// For vhost B don't disable IPv6
vhinfoB->options &= ~LWS_SERVER_OPTION_DISABLE_IPV6;
....
```

The global context option for IPv6 might not be needed though, it might be okay if that flag is only looked at for vhosts.

dosvald:
> I can migrate the ipv6-fulness flag to vhost by myself easily but is that really going to solve your problem? 

If IPv6 is checked per-vhost that'll solve this one. I have no idea how client context works, but I think for server + vhosts part, checking IPv6 per vhost is enough

lws-team:
[...]

PR #556, opened 2016-06-03
IPv6 fixes for bind, IPV6ONLY

### LINKED ISSUE #551: Bind vhost to IPv6 without IPv4
Regarding the IPv6 support and IPv4 mapping that lws uses, I think there might be room for improvement here. As it is, there is no occurrence of `setsockopt` with `IPV6_V6ONLY` in lws, which means that depending on the value of `/proc/sys/net/ipv6/bindv6only` the IPV6-capable sockets of lws will sometimes be v6only and sometimes not. That means that two vhosts on same port that are ipv4 and ipv6 can't be made at the same time. Perhaps lws could adopt some options per socket (vhost) which would mean either: disable v6only (use mapping), enable v6only (don't use mapping) or use OS default.

I'm not sure how other projects handle the v6only and sockets, but this option would enable to have separate vhosts for ipv4 port 80 and ipv6 port 80.

I'm guessing that would entail adding two bits of creation info `option` flags, or a separate field.

dosvald:
Should fix #551  and #552 and #553

### DIFF
--- a/lib/client-handshake.c
+++ b/lib/client-handshake.c
@@ -35,7 +35,7 @@ lws_client_connect_2(struct lws *wsi)
 		ads = wsi->vhost->http_proxy_address;
 
 #ifdef LWS_USE_IPV6
-		if (LWS_IPV6_ENABLED(context)) {
+		if (LWS_IPV6_ENABLED(wsi->vhost)) {
 			memset(&server_addr6, 0, sizeof(struct sockaddr_in6));
 			server_addr6.sin6_port = htons(wsi->vhost->http_proxy_port);
 		} else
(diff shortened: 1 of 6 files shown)


lws-team:
Thanks... I pushed them.  Thanks for fixing the missing \ in the define.

I left the client-handshake ones deliberately, on the basis there's no binding of the client socket to a vhost... but actually I added that recently.  So thanks for doing that as well.

I added lwjp-conf entries for them and related docs.

BTW I dunno if it's interesting but also recently the JSON config part of lwsws has moved into the library, you can enable it by itself at cmake with `-DLWS_WITH_LEJP=1 -DLWS_WITH_LEJP_CONF=1`

It means you can prepare the context creation and do the vhost creations using the same config / JSON as lwsws, without using lwsws.  If you're using multi vhosts, that might simplify things (or might not, depending on what exactly you're doing with it).  To use it you prepare your context create info as usual but call these

```
	if (lwsws_get_config_globals(&info, config_dir, &cs, &cs_len))
		goto init_failed;

	if (context == NULL) {
		lwsl_err("libwebsocket init failed\n");
		goto init_failed;
	}

	/*
	 * then create the vhosts... protocols are entirely coming from
	 * plugins, so we leave it NULL
	 */

	info.extensions = exts;

	if (!lwsws_get_config_vhosts(context, &info, config_dir,
				     &cs, &cs_len)) {
...
```

The two new apis in there set the context creation info from JSON config and loop creating whatever vhosts are found in the conf.d/  You can also pass it your own protocols[] and not use plugins if that's how you're set up.

dosvald:
Will any of these be going into 2.0.x branch?

lws-team:
No, because they add things to the api.

However I intend to release v2,1 when the generic-sessions plugin plus a plugin to demo it with is ready, which is not far off.

PR #560, opened 2016-06-06
Fix deprecated NDK tools for Android test-client

ABruines:
The latest Android SDK updates deprecate the shell script used to create a standalone toolchain. This patch updates the Makefile for the Android test-client to use the new ~~Perl~~ Python script to create the standalone toolchains.


ABruines:
Oeps. The 'Perl' script I mention in the commit is a actually a Python script...

lws-team:
Thanks... I did /Perl/Python/ in the commitlog and pushed it on master.

PR #565, opened 2016-06-14
Win con fix

ondraco:
Hello,

I have added a small "dumb" back up for windows in case that there is more than 64 connections.
The solution is mostly to avoid server failures in case of a seldom connection peaks. 

It just splits the events into groups of max 64 events and processes them in a sequence.
It is not really nice or fast or fair but windows does not give us much options here (at least while staying single-threaded, and supporting embedded).

Other possible solutions:
1. Start separate thread for every 64 connections
2. Use IOCP (not supported on embedded)

Best regards,
Ondrej Copak


lws-team:
Thanks, it's pushed.

I dunno if it helps but libuv support also works on windows.  At least on Linux, that's a way to get epoll() so maybe on windows it uses some other api.

PR #577, opened 2016-07-01
Suggestion for doing events one-by-one not a final implementaion

mjmessina:
Not a final implementation.  A starting point for the implementation.
Needs to handle timeouts, or have this be a new polling method.

### DIFF
--- a/lib/lws-plat-win.c
+++ b/lib/lws-plat-win.c
@@ -206,41 +206,50 @@ lws_plat_service_tsi(struct lws_context *context, int timeout_ms, int tsi)
 	/* if we know something needs service already, don't wait in poll */
 	timeout_ms = lws_service_adjust_timeout(context, timeout_ms, tsi);
 
-	ev = WSAWaitForMultipleEvents(pt->fds_count + 1, pt->events,
-				      FALSE, timeout_ms, FALSE);
-	context->service_tid = 0;
-
-	if (ev == WSA_WAIT_TIMEOUT) {
-		lws_service_fd(context, NULL);
-		return 0;
-	}
-
-	if (ev == WSA_WAIT_EVENT_0) {
+    ev = WSAWaitForMultipleEvents( 1,  pt->events , FALSE, 0, FALSE);
+    if (ev == WSA_WAIT_EVENT_0) {
 		WSAResetEvent(pt->events[0]);
 		return 0;
 	}
 
-	if (ev < WSA_WAIT_EVENT_0 || ev > WSA_WAIT_EVENT_0 + pt->fds_count)
-		return -1;
+    unsigned int indexToService =  -1;
+    for(unsigned int eIdx = 0; eIdx < pt->fds_count; ++eIdx)
+    {
+        ev = WSAWaitForMultipleEvents( 1,  &(pt->events[eIdx+1]) , FALSE, 0, FALSE);
+        if( ev ==  WSA_WAIT_FAILED )
+        {
+            lwsl_err("WSAWaitForMultipleEvents() failed with error %d\n", LWS_ERRNO);
+            continue;
+        }
+        if( ev == WSA_WAIT_EVENT_0 )
+        {
+            if (WSAEnumNetworkEvents(pt->fds[eIdx].fd,  pt->events[eIdx+1], &networkevents) == SOCKET_ERROR) {
+                lwsl_err("WSAEnumNetworkEvents() failed with error %d\n", LWS_ERRNO);
+                return -1;
+            }
+            if( indexToService == -1 )
+                indexToService = eIdx;
 
-	pfd = &pt->fds[ev - WSA_WAIT_EVENT_0 - 1];
+            pfd = &pt->fds[eIdx];
+            pfd->revents = (short)networkevents.lNetworkEvents;
 
-	/* eh... is one event at a time the best windows can do? */
+            if (pfd->revents & LWS_POLLOUT) {
+                wsi = wsi_from_fd(context, pfd->fd);
+                if (wsi)
+                    wsi->sock_send_blocking = 0;
+            }
 
-	if (WSAEnumNetworkEvents(pfd->fd, pt->events[ev - WSA_WAIT_EVENT_0],
-				 &networkevents) == SOCKET_ERROR) {
-		lwsl_err("WSAEnumNetworkEvents() failed with error %d\n",
-								     LWS_ERRNO);
-		return -1;
-	}
+            lws_service_fd_tsi(context, pfd, tsi);
+        }
+    }
+	context->service_tid = 0;
 
-	pfd->revents = (short)networkevents.lNetworkEvents;
+    if( timeout_ms >= 0 && indexToService == -1)
+    {
+        lws_service_fd(context, NULL);
+        return 0;
+    }
 
-	if (pfd->revents & LWS_POLLOUT) {
-		wsi = wsi_from_fd(context, pfd->fd);
-		if (wsi)
-			wsi->sock_send_blocking = 0;
-	}
 
 faked_service:
 
@@ -262,12 +271,7 @@ faked_service:
 		return 0;
 	}
 
-	if (timeout_ms < 0)
-		return 0;
-
-	/* otherwise just do the one... must be a way to improve that... */
-
-	return lws_service_fd_tsi(context, pfd, tsi);
+	return 0;
 }
 
 LWS_VISIBLE int
--- a/lib/lws-plat-win.c
+++ b/lib/lws-plat-win.c
@@ -212,39 +212,33 @@ lws_plat_service_tsi(struct lws_context *context, int timeout_ms, int tsi)
 		return 0;
 	}
 
-    unsigned int indexToService =  -1;
+    int servicedFd = 0;
+
     for(unsigned int eIdx = 0; eIdx < pt->fds_count; ++eIdx)
     {
-        ev = WSAWaitForMultipleEvents( 1,  &(pt->events[eIdx+1]) , FALSE, 0, FALSE);
-        if( ev ==  WSA_WAIT_FAILED )
-        {
-            lwsl_err("WSAWaitForMultipleEvents() failed with error %d\n", LWS_ERRNO);
-            continue;
+        if (WSAEnumNetworkEvents(pt->fds[eIdx].fd,  pt->events[eIdx+1], &networkevents) == SOCKET_ERROR) {
+            lwsl_err("WSAEnumNetworkEvents() failed with error %d\n", LWS_ERRNO);
+            return -1;
         }
-        if( ev == WSA_WAIT_EVENT_0 )
+
+        pfd = &pt->fds[eIdx];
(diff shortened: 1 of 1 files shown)


mjmessina:
@lws-team 
Here is my PR.  Not a final implementation.  Just a staring point.

lws-team:
Thanks... on master, this has code has changed via a patch from @dosvald ... dosvald, this solution seems better, can you try integrate it to master?  I think it's maybe not so hard...   Or any other comments on it?

dosvald:
I'd be happy to help, but I haven't worked with the lws-plat-win or the WSA
windows stuff unfortunately. Perhaps you meant to ping some other
contributor to lws-plat-win.c ?

lws-team:
Argh... sorry... @dosvald ... I meant @ondraco ... this PR seems to more or less solve what you were working around in https://github.com/warmcat/libwebsockets/commit/528adbde7f069e767a1418e8e9fc90a0e6c9283a + https://github.com/warmcat/libwebsockets/commit/f1bdb0fefe895d2598c2dc33e7e7097a58f9a5a2

could you take a look especially in terms of what might be missing.  Thanks.

mjmessina:
The obvious hole will be Its missing a timeout implementation.
I always pass 0 for a timeout, so not an issue for me.

ondraco:
The missing timeout implementation is big problem right? Anyone who uses non zero value and runs service in loop (on its own thread) will now get a 100% cpu usage.

lws-team:
Looking at the diff, am I missing the point or we can just use the computed timeout and add the check back in?

```
-	ev = WSAWaitForMultipleEvents(pt->fds_count + 1, pt->events,
 -				      FALSE, timeout_ms, FALSE);
 -	context->service_tid = 0;
 -
 -	if (ev == WSA_WAIT_TIMEOUT) {
 -		lws_service_fd(context, NULL);
 -		return 0;
 -	}
 -
 -	if (ev == WSA_WAIT_EVENT_0) {
 +    ev = WSAWaitForMultipleEvents( 1,  pt->events , FALSE, 0, FALSE);
 +    if (ev == WSA_WAIT_EVENT_0) {
```

ondraco:
For one i think there is no reason to use any alternative unless there is more than 64 handlers . Also I think it that it may be slower to call 64 times WSAWaitForMultipleEvents than call it once for 64. 

There really should be a working timeout or it can break many existing (and also new) implementations .

lws-team:
I tried reading about that api

https://msdn.microsoft.com/en-us/library/windows/desktop/ms742219(v=vs.85).aspx

It seems always wait each one in turn is really inefficient.

Actually the api returns the lowest index in the wait list that has an event.

I think ondraco's meaning is you shouldn't test against WSA_WAIT_EVENT_0 with 1 handle each time.

Instead you can first wait on all events, the return will be WSA_WAIT_EVENT_0 + idx if any events remain... service that one and then loop wait on events n + 1 ... event_count, etc until no events left.  That way instead of the max times around the loop every wait, if there are only n sockets with events we only loop n + 1 times... does it make sense?

Also the timeout value must be used, at least the first time.  Because 0 timeout for that api means no wait == 100% cpu all the time.

Can you maybe modify the patch for these points and check again on ab?  Performance should stay the same on ab because he's artificially busy.  But for real loads cpu efficiency will radically improve.

ondraco:
What is the actual motivation in doing it one by one? The previous solution left it alone unless it was needed (only for cases when more than 64 events). And when there was more than 64 events it did something like this but with chunks of 64 events and with manually computed timeout.

mjmessina:
There was something wrong with the implementation.  I would get into a state where the server would not respond.  I think it was something with the timeouts being set and the events being signaled where not in the set of 64 currently being looked at.

I did it one-by-one just to get something going for me.

It would seem counter intuitive, but It is actually more responsive because it pumps thru each event that is signaled in one loop.  The loop is just inverted (rather than the caller calling).  Either way you look at it, you are looping thru each event.

mjmessina:
One implementation I thought of (but did not try yet), was to do 64 at a time (like you did) and the use WSAEnumNetworkEvents to go thru each of the events in the 64 to see if anymore where signaled. 

Of just to use WSAEnumNetworkEvents in a loop.  

Either way the timeout handing sucks because you are going to get caught in a wait while other events are signed.

lws-team:
No you only have to handle the timeout on the first wait.

If you come out of that with the timedout return code you don't do anything more, because it means nobody had events.

If you kept the existing code structure the subsequent partial waits wouldn't have a timeout, and remain at 0.

Ondraco's 'backup solution' also spins (100% cpu) it seems....

If this is difficult with these nasty win32 apis, maybe we should just make windows users use libuv.

ondraco:
Actually it was done as a compromise between spin and fairness. It should take almost no CPU because it was first checking all the handlers with 0 timeout but the last group had a timeout set (with maximum of 20ms). And this was running in a loop until event was triggered or all the time was wasted.

The problem is that any other solution either requires additional threads or wont work on older platforms (for example Windows embedded).

mjmessina:
Just an FYI, I ditched the WSAWaitForMultipleEvents and did a loop like  below (much better).

Would you guys be willing to have two paths?  One to support those who don't want to spin and one that does?  I would rather spin and minimally put a sleep(1) to get better performance.

`for(unsigned int eIdx = 0; eIdx < pt->fds_count; ++eIdx)
    {
        if (WSAEnumNetworkEvents(pt->fds[eIdx].fd,  pt->events[eIdx+1], &networkevents) == SOCKET_ERROR) {
            lwsl_err("WSAEnumNetworkEvents() failed with error %d\n", LWS_ERRNO);
            return -1;
        }

        pfd = &pt->fds[eIdx];
        pfd->revents = (short)networkevents.lNetworkEvents;

        if (pfd->revents & LWS_POLLOUT) {
            wsi = wsi_from_fd(context, pfd->fd);
            if (wsi)
                wsi->sock_send_blocking = 0;
        }

        if(pfd->revents  != 0 )
        {
            lws_service_fd_tsi(context, pfd, tsi);
            ++servicedFd;
        }
    }`

ondraco:
I tested it (with hundreds of connections) and it was working quite well but for sure It is possible that there was some bug but it should be possible to fix it.

It was working like this (T is timeout set by user):

1. If there is less than 64 handlers -> do nothing special just use WSAWaitForMultipleEvents()
2. If there is more than 64 handlers:

     (A)  Split all handlers in groups by 64 (example: we got 80 so we have group X with 64 and Y with 16)
     (B)  Wait for all groups but last with 0 timeout 
     (C)  Wait for last group with timeout calculated as MAX(T, 20). I used 20 as just some random coefficient the lower it is the more CPU is taken and the bigger it is the less fair it is.
     (D): If we got triggered event or we used all time T -> quit and return the event ID or               WSA_WAIT_TIMEOUT else goto (A)

mjmessina:
With the last commit (ditching the wait for multiple events).  
You get a really tight mean times.  Note the results below.

$ ./ab -n 15000 -c 800 127.0.0.1:8080/test
This is ApacheBench, Version 2.3 <$Revision: 1706008 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient)
Completed 1500 requests
Completed 3000 requests
Completed 4500 requests
Completed 6000 requests
Completed 7500 requests
Completed 9000 requests
Completed 10500 requests
Completed 12000 requests
Completed 13500 requests
Completed 15000 requests
Finished 15000 requests


Server Software:
Server Hostname:        127.0.0.1
Server Port:            8080

Document Path:          /test
Document Length:        49 bytes

Concurrency Level:      800
Time taken for tests:   11.811 seconds
Complete requests:      15000
Failed requests:        0
Total transferred:      2040000 bytes
HTML transferred:       735000 bytes
Requests per second:    1270.01 [#/sec] (mean)
Time per request:       629.919 [ms] (mean)
Time per request:       0.787 [ms] (mean, across all concurrent requests)
Transfer rate:          168.67 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   1.3      0      16
Processing:    99  615  73.1    633     670
Waiting:       98  614  73.1    633     669
Total:         99  615  73.1    633     670

Percentage of the requests served within a certain time (ms)
  50%    633
  66%    635
  75%    639
  80%    639
  90%    649
  95%    651
  98%    655
  99%    655
 100%    670 (longest request)

ondraco:
But this still ignores the timeouts right? So if you run it will just spin CPU on 100% all the time. For example I use the LWS on windows embedded devices with close to no processor power and 1 core. This would just fry them :-).

mjmessina:
Correct, it ignores timeouts.
That's why I am requesting a special poll call.
This is good for servers with many CPUS and good cooling.

lws-team:
Let me sum it up...

1) LWS_WITH_LIBUV and select LIBUV at context creation time bypasses all this.  I get it for even crapper versions of windows that may not be simple enough to build or whatever.

2) sleep()s and 100% cpu spins won't fly.  Something is broken in the patch actually if it can't go idle when it is idle.

3) Ondraco says the master implementation also passes ab, mjmessina was not testing or patching that actually but the v2.0.x implementation before.  So we are basically fighting ghosts.

4) Ondraco says his loop is idle when things are idle.  So it seems already reasonable on master.

5)  Master is all we should worry about since it will soon become 2.1 and the other branches are stuck with what they got (which has no sleeps, goes idle properly but due to windows limits has trouble at 64 connections).

If that correctly sums it up we should stick with what we have.  @mjmessina how about just try master branch as it is for reference?

mjmessina:
I'll try the master again.
If it locks up I'll try and debug it.

mjmessina:
@ondraco
I think I see my issue with the master,  I use a timeout of 0.

The back up solution doesn’t do anything because of the while timeout > 0.

I'll verify tomorrow.

lws-team:
Thanks... just to clarify timeout of zero for lws generally makes no sense.

The timeout is only triggered if every socket was completely idle for the whole wait.

As soon as somebody had an event, the wait exits, even if the timeout was ten years it exits right away.

So there is normally no reason to make it spin over and over with a zero timeout.

mjmessina:
I do some other checks in the poll using that thread.
I don't want to wait. I know how to get around it now.

mjmessina:
You should change your documentation, it states 0  a legitimate value.

You can close my question.  I know why we couldn't handle more than 64 connections. 

Thanks so much for your time!
Love this library its been useful.

lws-team:
OK if I understood it, everyone is happy with current master.  Thanks a lot for the testing.

ondraco:
I think I understand the problem now. I don't have any usable computer where i could fix/test it but I believe that just changing the:

`(line 173) while(timeout_left > 0)` 

with 

`do { ... } while(timeout_left > 0)` should make it work correctly also when timeout is 0. We just need to make sure the loop always runs at least once.

lws-team:
I pushed a patch to clean the style and make this a do-while.

mjmessina:
Great! Thanks Andy.

PR #580, opened 2016-07-06
Use one event for all sockets to avoid 64 limit and fix the single dispatch issue

mjmessina:
@lws-team 

Not asking you to take this PR, just something to consider for the future as the overall performance is better.  The AB test shows much tighter and lower mean numbers.  I will be testing this in our server this week.

I did some playing around with just using a single event.   The timeout value is now what you want.
You don't have to do 20mSec steps.
@ondraco  - this may be better for your single processor when you hit more than 64 fds.


mjmessina:
@lws-team - Andy for you consideration. Just trying to help out since your lib has help me tremendously! 

I have deployed this fork in a prod environment.  Works great for my use case.  Wasn't sure if I broke the fake_service.

lws-team:
Faked service needs to work like it works atm.

Outside of this platform-specific function, lws may decide for a few reasons to force signalling that a wsi behind a pollfd needs service by forcing revents.  For example, it may have been forced to buffer RX data (that is no longer visible in the socket, because it has been read) and needs further service because lws had to return to the event loop before it used it all up.

It seems your patch thinks it ought to use WSAEnumNetworkEvents instead, but that is not what is going on with this faked service concept... you can't detect what needs doing any more at the network level.

lws-team:
In case that's not clear, the related code is quite clear

https://github.com/warmcat/libwebsockets/blob/master/lib/service.c#L424

```
/*
 * guys that need POLLIN service again without waiting for network action
 * can force POLLIN here if not flowcontrolled, so they will get service.
 *
 * Return nonzero if anybody got their POLLIN faked
 */
int
lws_service_flag_pending(struct lws_context *context, int tsi)
{
	struct lws_context_per_thread *pt = &context->pt[tsi];
#ifdef LWS_OPENSSL_SUPPORT
	struct lws *wsi_next;
#endif
	struct lws *wsi;
	int forced = 0;
	int n;

	/* POLLIN faking */

	/*
	 * 1) For all guys with already-available ext data to drain, if they are
	 * not flowcontrolled, fake their POLLIN status
	 */
	wsi = pt->rx_draining_ext_list;
	while (wsi) {
		pt->fds[wsi->position_in_fds_table].revents |=
			pt->fds[wsi->position_in_fds_table].events & LWS_POLLIN;
		if (pt->fds[wsi->position_in_fds_table].revents &
		    LWS_POLLIN)
			forced = 1;
		wsi = wsi->u.ws.rx_draining_ext_list;
	}

#ifdef LWS_OPENSSL_SUPPORT
	/*
	 * 2) For all guys with buffered SSL read data already saved up, if they
	 * are not flowcontrolled, fake their POLLIN status so they'll get
	 * service to use up the buffered incoming data, even though their
	 * network socket may have nothing
	 */
	wsi = pt->pending_read_list;
	while (wsi) {
		wsi_next = wsi->pending_read_list_next;
		pt->fds[wsi->position_in_fds_table].revents |=
			pt->fds[wsi->position_in_fds_table].events & LWS_POLLIN;
		if (pt->fds[wsi->position_in_fds_table].revents & LWS_POLLIN) {
			forced = 1;
			/*
			 * he's going to get serviced now, take him off the
			 * list of guys with buffered SSL.  If he still has some
			 * at the end of the service, he'll get put back on the
			 * list then.
			 */
			lws_ssl_remove_wsi_from_buffered_list(wsi);
		}

		wsi = wsi_next;
	}
#endif
	/*
	 * 3) For any wsi who have an ah with pending RX who did not
	 * complete their current headers, and are not flowcontrolled,
	 * fake their POLLIN status so they will be able to drain the
	 * rx buffered in the ah
	 */
	for (n = 0; n < context->max_http_header_pool; n++)
		if (pt->ah_pool[n].rxpos != pt->ah_pool[n].rxlen &&
		    !pt->ah_pool[n].wsi->hdr_parsing_completed) {
			pt->fds[pt->ah_pool[n].wsi->position_in_fds_table].revents |=
				pt->fds[pt->ah_pool[n].wsi->position_in_fds_table].events &
					LWS_POLLIN;
			if (pt->fds[pt->ah_pool[n].wsi->position_in_fds_table].revents &
			    LWS_POLLIN)
				forced = 1;
		}

	return forced;
}
```

mjmessina:
What I wasn't sure about is that a  (timeout_ms < 0) implies the faked service.
Previously there was a check and a goto.  I just moved the code into the if statement.
So if faked_service == (timeout < 0).  Then all should be good.  If that is not true, then I broke something.

mjmessina:
If you use the split view diff, you can see what I did a little better.

lws-team:
I see, you moved it around... I was looking at this from the diff

```
-	if (lws_service_flag_pending(context, tsi)) {
-		/* any socket with events to service? */
-		for (n = 0; n < (int)pt->fds_count; n++) {
-			if (!pt->fds[n].revents)
-				continue;
+			pfd = &pt->fds[eIdx];
+			pfd->revents = (short)networkevents.lNetworkEvents;
```

Actually you kept it but moved it

```
+			if (lws_service_flag_pending(context, tsi)) {
+			/* any socket with events to service? */
+			for (n = 0; n < (int)pt->fds_count; n++) {
+				if (!pt->fds[n].revents)
+					continue;
+
+				m = lws_service_fd_tsi(context, &pt->fds[n], tsi);
```

lws-team:
I pushed it on master let's see what happens.

PR #584, opened 2016-07-15
V2.0 stable

rollen1960:
It possible to use lws_serve_http_file for sending more files. I must send css style file + html file + javascript file. How is possible ?

Thank you

Mario


lws-team:
I dunno why you sent a pull request pushing v2.0-stable on master... that won't do anything.  You can open an issue if you just have a question.  So I will close the PR since it seems unintentional.

There are three ways to serve files with lws 2.0...

1) Do it like test-server/test-server.c.... basically deal with each file by hand in the code.

2) Do it like test-server/test-server-v2.0.c ... he "mounts" a filesystem directory on the vhost.  He will automatically serve any file he can figure the mimetype for that is in the filesystem directory, including subdirectories.  This is the recommended way now.

3) Use lwsws, it's a generic webserver built on lws that you configure using JSON config files.  You can set up mounts in JSON without having to write any code.

You can see lwsws in action on https://libwebsockets.org... and you can see his vhosts and mounts from the live server status page https://libwebsockets.org/server-status/

PR #597, opened 2016-08-08
esp8266 initial support

bettersTsai:


lws-team:
I'll be very happy to get a PR about ESP8266, but d6f367f you pushed here is old master from a few days ago.  Update against current master (you can do like git fetch https://github.com/warmcat/libwebsockets +master:m1 ; git reset --hard m1 to force your current branch to current master) and try create the PR again if something to look at.

PR #603, opened 2016-08-13
Allow serving files without known mimetype

paroga:
RFC2616 only says that any HTTP/1.1 message containing an entity-body
SHOULD include a Content-Type header field defining the media type of
that body.
RFC2119 defines SHOULD as: This word mean that there may exist valid
reasons in particular circumstances to ignore a particular item, but
the full implications must be understood and carefully weighed before
choosing a different course.

### DIFF
--- a/lib/server.c
+++ b/lib/server.c
@@ -428,10 +428,8 @@ lws_http_serve(struct lws *wsi, char *uri, const char *origin,
 #endif
 
 	mimetype = lws_get_mimetype(path, m);
-	if (!mimetype) {
-		lwsl_err("unknown mimetype for %s\n", path);
-		goto bail;
-	}
+	if (!mimetype)
+		lwsl_debug("sending no mimetype for %s\n", path);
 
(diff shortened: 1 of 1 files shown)


lws-team:
I think probably in most cases, it's better to supply the mimetype, in the mount struct there is a place to provide a linked list of extra mimetype ".xyz": "text/whatever" type mappings, and it's supported in lwsws config already.

The reason it's using a whitelist by default is I don't want to find out I made a flaw in the directory traversal stuff and I have been giving out /etc/passwd or /etc/shadow (let's ignore you shouldn't be running as root), this mimetype requirement means various other sensitive files also with unusual (or no) suffixes won't be served without greater effort at widening the (hopefully entirely theoretical) hole.

Anyway I adapted the patch to allow giving a wildcard extra-mimetype "*": "" that will match anything and send it without a Content-Type, and pushed it on master.

paroga:
THX. That does the job even better. :-)

PR #604, opened 2016-08-14
Remove unused context from lws_context_per_thread

paroga:
The variable is never read. So it's safe to remove it.


lws-team:
As Travis points out, it is read in the libuv support.

https://travis-ci.org/warmcat/libwebsockets/jobs/152178699

PR #606, opened 2016-08-14
Add error handling for SSL_new() of clients

paroga:
Do not access wsi->ssl if SSL_new() failed and log the error.


lws-team:
Thanks... it's pushed on v2.0-stable and master

PR #607, opened 2016-08-14
Remove context from lws_context_per_thread for non-libuv builds

paroga:
The variable is never read when built without libuv.


lws-team:

PR #612, opened 2016-08-25
Remove unneeded #include <stdint.h>

paroga:
This fixes the build for Visual Studio 2008.

### DIFF
--- a/lib/libwebsockets.h
+++ b/lib/libwebsockets.h
@@ -116,7 +116,6 @@ struct sockaddr_in;
 #include <stddef.h>
 #include <basetsd.h>
 #ifndef _WIN32_WCE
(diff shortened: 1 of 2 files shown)


lws-team:
It may not be needed on vs2008, but it is needed on whatever vs appveyor uses (2014 or such), as you can see from the appveyor fail log on the link above.  So it is not useless cruft per se.

Patching around it is fine, but it will need to be made conditional on, eg, > VS2008.

lws-team:
Thanks, that works too: it's pushed.

paroga:
Can you push it on the stable branch too?

lws-team:
Sure... it's done thanks.

paroga:
thx!

PR #635, opened 2016-09-26
fix build problem on systems without X509_VERIFY_PARAM type defined

shifty35:
Had an issue building against wolfssl on an ARM based system, this patch seems to resolve it.


lws-team:

PR #636, opened 2016-09-26
Added option to build the static library with PIC

arc6011:
I needed PIC in the static library for building a shared object plugin.

Of course, this adds some dependencies on the version of the loader (and perhaps other tools in the chain). I'm sure that it should be "OFF" by default and only "ON" if the user knows his/her build environment.

I'd be happy to port this to other versions as well. My project is still on 1.7 for now.


lws-team:
Thanks, I pushed it on v1.7-stable and ported and pushed it on v2.0-stable and master.

Issue #638, opened 2016-09-28
Server failing to process pipelined calls


dosvald:
When a HTTP/1.1 client talks with a libwebsockets-powered server and pipelines multiple HTTP requests in a single TCP connection, depending on the network and/or OS conditions of the two peers and/or hops between them, libwebsockets will fail to process the pipelined requests beyond the first one. The client will then hang on to the HTTP connection, waiting for more pipelined responses, but they will never arrive.

The problem is present when, due to speed of incoming traffic, a single `recvfrom` call results in data containing multiple (pipelined) HTTP requests, in which case lws only processes the first one. There may be other similar cases triggerint the behavior too, however. Using SSL I was not able to reproduce it.

The issue is easiest to reproduce with the tool httperf:
```
$ httperf --server libwebsockets.org --port 80 --uri /favicon.ico --max-connections 1 --num-calls 20 --burst-length 2 --timeout 5

httperf --timeout=5 --client=0/1 --server=libwebsockets.org --port=80 --uri=/favicon.ico --max-connections=1 --send-buffer=4096 --recv-buffer=16384 --num-conns=1 --num-calls=20 --burst-length=2
httperf: warning: open file limit > FD_SETSIZE; limiting max. # of open files to FD_SETSIZE
Maximum connect burst length: 0

Total: connections 1 requests 4 replies 3 test-duration 5.629 s

Connection rate: 0.2 conn/s (5629.3 ms/conn, <=1 concurrent connections)
Connection time [ms]: min 5629.3 avg 5629.3 max 5629.3 median 5629.5 stddev 0.0
Connection time [ms]: connect 313.5
Connection length [replies/conn]: 3.000

Request rate: 0.7 req/s (1407.3 ms/req)
Request size [B]: 81.0

Reply rate [replies/s]: min 0.6 avg 0.6 max 0.6 stddev 0.0 (1 samples)
Reply time [ms]: response 315.1 transfer 0.0
Reply size [B]: header 186.0 content 0.0 footer 0.0 (total 186.0)
Reply status: 1xx=0 2xx=0 3xx=3 4xx=0 5xx=0

CPU time [s]: user 1.18 system 4.45 (user 20.9% system 79.1% total 100.0%)
Net I/O: 0.2 KB/s (0.0*10^6 bps)

Errors: total 1 client-timo 1 socket-timo 0 connrefused 0 connreset 0
Errors: fd-unavail 0 addrunavail 0 ftab-full 0 other 0
```

[...]

PR #642, opened 2016-10-07
test-server-extpoll: check for forced service before looping again

dosvald:
Addresses https://github.com/warmcat/libwebsockets/issues/638

Signed-off-by: Denis Osvald <denis.osvald@sartura.hr>


lws-team:
I wondered if that was the case.... anyway thanks for the patch it's on v2.1-stable and master.

PR #643, opened 2016-10-07
add timeout servicing to extpoll and API doc

dosvald:


lws-team:
Thanks they're also pushed on master + v2.1-stable.  Of the last 3 patches, the first 2 also fit on v2.0-stable so I pushed those there too.

dosvald:
Great, but you've backported the wrong commit to v2.0-stable - test-server-extpoll can't compile (it requires the exposed API only in v2.1).

Perhaps you meant to backport api doc commit and timeout servicing commit, and not forced servicing one?

lws-team:
Ugh I got rid of them on v2.0 stable, sorry.

When I looked at them I thought it was doing the expoll forced service work in the library, I missed seeing they were patching the test server.

dosvald:
Unfortunately I'm not sure how extpoll forced service would be done in the library. At least not in a stable branch without breakage and/or general changes to meaning of `lws_service_*` functions, unless you have any ideas?

dosvald:
Perhaps a new issue can be opened for this and to make known that v2.0 has that problem that isn't yet fixable.

lws-team:
I think people who need extpoll will need to use v2.1 or master.

We can play games around adding a new meaning to -1 timeout_secs in the existing apis.  But I don't think it makes sense to fragment the support further, eg, v2.1 release won't support the new trick.

dosvald:
Okay then. I believe making a known-bug note in release notes would be a good idea, or announcement on the mailing list, saying that v2.0 sometimes breaks pipelining when using external poll.

lws-team:
Yeah I agree.  I added a note to README.coding.md on v2.0-stable at the top of the external poll section explaining the situation and telling people to use v2.1 or master for this feature to work reliably.

PR #645, opened 2016-10-10
Correct the library name of LIBHUBBUB_LIBRARIES

mapx:
It should be `hubbub` in `find_library` to make the function work.


lws-team:
Thanks, it applies to v2.0-stable, v2.1-stable and master, it's pushed on all three.

PR #660, opened 2016-11-08
Better support for MINGW32

rpoisel:
When (cross-)compiling using the MINGW32 toolchains there are some warnings and issues that could be fixed easily. This commit is a proposal in this context, fixing a few issues I had when cross-compiling from Linux to 32-bit Windows.


lws-team:
Thanks, it's pushed on v2.1-stable and master.

PR #667, opened 2016-11-11
Added flag to allow expired certificates.

fancycode:
I have a use case where I'm connecting to a device which on first boot creates a certificate and might have a wrong clock at that time, so the certificate might not yet / no longer be valid when checked with a "correct" clock.
So I need to accept expired certificates until the certificate has been replaced with a valid one.


lws-team:
Thanks, I pushed it on master.  But this will only work for client.

fancycode:
Thanks! For my use case, the client part is sufficient.

PR #679, opened 2016-11-21
update path variable to larger size

sjames1958gm:
I wasn't sure if there was a constant or better value to set this to.


lws-team:
Thanks, since it's a private struct I pushed it on v2.0-stable, v2.1-stable and master.

PR #694, opened 2016-11-29
Update CMakeLists.txt

Bablooos:
Fixing build failure of libwebsockets-test-fraggle  on FreeBSD when LWS_WITH_PLUGINS. 
Solution: FreeBSD has no libdl


lws-team:
Thanks: I was wrong it still links libdl in the CMake file as you said... that is the actual problem this solves then I think.

It's pushed on v2.0-stable, v2.1-stable and master.

PR #696, opened 2016-11-29
ability to creating vhosts after libuv loop started

Bablooos:
I added private function int lws_uv_initvhost(struct lws_vhost* vh, struct lws* wsi) to libuv.c that makes initialization of vhost (from lws_uv_initloow - while( vh ) {...}) when libuv loop already started only. It called for every vhost creation and for each created vhost in lws_uv_initloop. When libuv loop is not started does nothing. Check if initialization on vhost already done.


lws-team:
I modified this a bit and pushed it on master

https://libwebsockets.org/pipermail/libwebsockets/2016-November/002968.html

PR #726, opened 2016-12-20
V2.1 stable aconway adopt ssl

alanconway:
These 3 commits enable SSL for a server that uses external polling and socket adoption. All minor, no new functionality, just exposing features that were already there but not accessible via public API. 

Testing: test-server, manual test via browser with/without SSL, test-client, autobhan.sh script. All reported OK.

### DIFF
--- a/lib/libwebsockets.h
+++ b/lib/libwebsockets.h
@@ -3514,6 +3514,7 @@ lws_remaining_packet_payload(struct lws *wsi);
 
 /**
  * lws_adopt_socket() - adopt foreign socket as if listen socket accepted it
+ * for the default vhost of context.
  * \param context: lws context
  * \param accept_fd: fd of already-accepted socket to adopt
  *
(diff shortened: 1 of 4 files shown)


lws-team:
Thanks... I pushed them on master... the ssl error fix should be helpful to everyone.

Although they're based on v2.1-stable on your branch, I can't push them there since they change the api while the soname is the same.

What I suggest is wait a little bit to see if there's any more api breaking changes needed and then release master as v2.2 with an soname bump.

alanconway:
On Tue, 2016-12-20 at 17:37 -0800, Andy Green wrote:
> Thanks... I pushed them on master... the ssl error fix should be
> helpful to everyone.
> Although they're based on v2.1-stable on your branch, I can't push
> them there since they change the api while the soname is the same.
> What I suggest is wait a little bit to see if there's any more api
> breaking changes needed and then release master as v2.2 with an
> soname bump.

OK by me - I've tested with master and it all works fine. Thanks for
taking the patches!

PR #735, opened 2017-01-02
server: check listen(2) return value

dosvald:
The `listen` call can fail with EADDRINUSE after bind() succeeds, for
example because another process called listen on that port in the
meantime, or under some circumstances with IPv6-mapped-IPv4. This was
causing EINVAL on accept, with an infinite loop in case of libuv.

A reproducible example was to run nc -l -p 5555 ( OpenBSD netcat (Debian
patchlevel 1)) before starting test-server on port 5555

Signed-off-by: Denis Osvald <denis.osvald@sartura.hr>

### DIFF
--- a/lib/server.c
+++ b/lib/server.c
@@ -160,7 +160,15 @@ lws_context_init_server(struct lws_context_creation_info *info,
 	vhost->lserv_wsi = wsi;
 
 #if LWS_POSIX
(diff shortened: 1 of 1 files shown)


lws-team:
Thanks... I pushed it on v2.0-stable, v2.1-stable and with a little adaptation, master.

PR #748, opened 2017-01-13
undefined reference of 'dlopen,dlclose, dlsys' fixes while linking on ubuntu 64bit system

dumganhar:


lws-team:
Thanks... however this isn't what it seems... the underlying issue is default libuv on ubuntu is ancient, and doesn't have libuv dl support lws wants.

If you install `libuv1` package instead, this problem will go away, and lws will actually be able to work.

PR #749, opened 2017-01-16
ignore leading spaces when checking for a suitable subprotocol

tobiasgraf:
My Browsers send subprotocols e.g. chat, superchat, mySubprotocol (with spaces after the comma ). Libwebsockets now checked if `' mySubprotocol'` (mind the space) was equal to `'mySubprotocol'` which failed. With this fix the leading space is ignored and ins this case uses 'mySubprotocol' for comparision.


lws-team:
Thanks... it's pushed on master / v2.1-stable / v2.0-stable.

PR #760, opened 2017-01-23
SSL expose as public API

dosvald:
Some programs may want to use SSL-related capabilities that libwebsockets doesn't implement yet, or that it may never implement or expose. Rather than implement all possible SSL features into lws, this might be the easier route by just exposing it to applications and let them deal with the implementation specifics themselves.


lws-team:

PR #761, opened 2017-01-23
wsi remove unused 'upgraded' boolean field

dosvald:
While grepping something I ran into this field that seems unused. Please double check, but I don't see this used anywhere. Perhaps it was meant to be used / checked somewhere and removing it is a fix in the wrong direction?


lws-team:
Thanks.  Yes it's just cruft... introduced it while fighting something and then solved it a better way and forgot about it.  I pushed it on master.

PR #769, opened 2017-02-02
adding gethostbyname for ipv4

andSpace:
Fallback for ipv4 to use gethostbyname in case getaddrinfo returns EAI_SYSTEM

### DIFF
--- a/lib/client-handshake.c
+++ b/lib/client-handshake.c
@@ -113,6 +113,8 @@ lws_client_connect_2(struct lws *wsi)
 	} else
 #endif
 	{
+		int addr_rv;
+		struct hostent *host;
 		struct addrinfo ai, *res, *result = NULL;
 		void *p = NULL;
 		unsigned char *p1;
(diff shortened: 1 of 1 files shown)


lws-team:
The problem I am having with this is why would getaddrinfo reply EAI_SYSTEM but gethostbyname can work?

IPv6 never does that or you are just not using it?

What platform was this on?

andSpace:
This is for a low-powered chipset kind of like raspberry-pi but worse, and the vendor provided kernel while relatively modern (~3.0.6) has a decent amount of features turned off. Unfortunately I do not have their config or the kernel so I can't confirm which options are enabled -- but what does happen is ping/curl are able to resolve addresses while using getaddrinfo throws EAI_SYSTEM. 

The kernel in question definitely doesn't have  IPV6 support, which makes me think this wouldn't even be an issue if it was. (Additionally what makes me think this is a slimmed down kernel is that inotify returns not implemented system errors)

lws-team:
This doesn't really fix any problem in lws... I added it, but behind a default-off CMake variable.

I also saw it would read-after-free the getaddrinfo attempt the way you did it, so I fixed that while I was about it.

It's pushed on master.

andSpace:
Awesome, thanks for adding this -- one note in the modified version it doesn't compile since the return value of getaddrinfo is never saved.

```
error: ‘addr_rv’ undeclared (first use in this function)
   } else if (addr_rv == EAI_SYSTEM) {
```

lws-team:
Sorry, that's my 'improvement'... I pushed an updated version of the patch on master that should sort it.

PR #772, opened 2017-02-04
Make GCC check format strings

mmilata:
Adds the `format` [function attribute](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html) when compiled by gcc and fixes whatever it found on my system.

* I'm assuming not all platforms the library supports have `%z` so I casted `size_t` to `(unsigned long)` where needed.
* `ERR_get_error` seems to return `unsigned long` so the commit changes the type of couple of variables.

There's also small docstring fix.


lws-team:
Thanks... that found a lot of nits... I built here with more options and in DEBUG mode, it found more, so I added those on a third patch on master.  I also backported them to v2.1-stable.

PR #777, opened 2017-02-07
Fixes a conversion error in the test client

iUltimateLP:
This fixes a conversion error from a void pointer to a string, hope I'm not wrong here or didn't set something up in the project settings since I'm migrating this code file to a smaller test project.


lws-team:
Thanks... most compilers will accept a void * as any pointer at the same level.

I changed the cast in your patch to const char * and pushed it on master.

PR #793, opened 2017-02-14
changes to build with VS2015

jwinarske:
Changes to enable building with VS2015, and running lwsws on Windows.


lws-team:

PR #807, opened 2017-02-21
SSL config for http client

jwinarske:


namowen:
Hi, what is the motivation of this pull request? With your changes in ssl-client.c you will break any existing code used til your pull request ;-)
I'm not using plugins at all, but I have the impression you want to solve something in this area, what is ok, but pls. don't break the code of the whole thing.

lws-team:
Ah it's okay... sorry for late reply I only just read it.

I asked for this hacked code to save me time implement it normally shortly.... basically make sure I implement the right things.    It was told it was a hack initially on the discussion on the ml... @jwinarske thanks.  @namowen don't worry it's just for reference.

lws-team:
Thanks that was a bigger help than I expected, thanks a lot.  It's pushed on master.

I used it directly except the last bit, I adapted it to continue to use the server ssl config if the new client members are NULL.  If anyone was relying on that and would have been surprised by the change, that should make it invisible if they don't care and were happy the old way.  But obviously the new way allowing them to be controlled separately is a lot more useful.

PR #812, opened 2017-02-21
gcc format strings: couple more

mmilata:
Found with gcc-4.7 on mipsel.

See #811 for master branch (the one in handshake.c is already included).


lws-team:
The handshake.c one is already in current master... I added the first PR in permessage-deflate only.

mmilata:
Sorry, forgot to mention that this one is against `v2.1-stable`.

lws-team:
Ah I see... it's also pushed on v2.1-stable now, thanks.

mmilata:
Thank you!

PR #814, opened 2017-02-22
Add client support for relative reference redirects

skyhisi:
This adds client support for relative redirects, the client will now support redirects of the form:
* `/absolute/path`
* `relative/path`
* `http://full.uri/with/path`


lws-team:
Great stuff, thanks.

I squashed the first and second and pushed "all three" on master.

PR #829, opened 2017-03-09
fix log compile error

honjane:


lws-team:
Thanks, I squashed the two patched into one and pushed it on master + v2.2-stable.

PR #839, opened 2017-03-18
added missing file to %files section of spec file

stv0g:
The .spec file is missing a file in the %files section.

The causes rpmbuild to fail.


lws-team:
Thanks a lot, it's pushed on master + v2.2-stable.

PR #843, opened 2017-03-21
[Libevent] Initial Support

adityata:
* Added libevent support functionality into lib/libevent.c
* Added test-server-libevent for testing


lws-team:
Thanks a lot... nice job... it's pushed on master.

PR #849, opened 2017-03-28
Modified CMakeLists.txt to fix issue when used as a cmake subdirectory

JonnyPtn:
Hi, I'm trying to include libwebsockets in my cmake build process, by cloning the repository and then adding it as a subdirectory in my cmakelists. Currently I can't do this, as it fails when copying libwebsockets.h and lws_config.h into the include folder.

It seems changing the paths shown here fixes the issue, and has no negative effects when building the project by itself.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -773,13 +773,13 @@ if (LWS_WITH_STATIC)
 	endif()
 	add_custom_command(
 		      TARGET websockets
-		      COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_SOURCE_DIR}/lib/libwebsockets.h
-		     					 ${CMAKE_BINARY_DIR}/include/libwebsockets.h
+		      COMMAND ${CMAKE_COMMAND} -E copy ${PROJECT_SOURCE_DIR}/lib/libwebsockets.h
+		     					 ${PROJECT_BINARY_DIR}/include/libwebsockets.h
 	)
 	add_custom_command(
 		      TARGET websockets
-		      COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_BINARY_DIR}/lws_config.h
-		     					 ${CMAKE_BINARY_DIR}/include/lws_config.h
+		      COMMAND ${CMAKE_COMMAND} -E copy ${PROJECT_BINARY_DIR}/lws_config.h
+		     					 ${PROJECT_BINARY_DIR}/include/lws_config.h
 	)
 
 endif()


JonnyPtn:
Sorry, I'm going to close this. After investigating further it seems that this change doesn't fix the problem and is a red herring. I'll open an issue to discuss the problem

JoakimSoderberg:
Hello...

Now I have not tested this with the absolute latest version of libwebsockets, so it might be broken.

But 3 different ways to link with libwebsockets (with options).

1. Use CMake to download libwebsockets via git, build it and then link with the resulting binaries.

2. Add libwebsockets as a submodule to the project (or just dump the files if that floats your boat) into the project, and then point to that location for the CMake find_package command. Note this requires that you manually build the version first

3. Find the system version of libwebsockets and link with that.

(4. Another option would be to use the CMake add_subdirectory command, but not sure if the libwebsockets cmake project can handle that. Maybe this was what you wanted to do?)

```cmake
option(AUTO_LWS "Automatically download and build dependencies" OFF)
option(INTERNAL_LWS "Use the included version of Libwebsockets" ON)

# So in the example below you would build LWS using CMake and output the binaries to this path.
# (used for option 2, when INTERNAL_LWS is ON and AUTO_LWS is OFF)
set(LIBWEBSOCKETS_PATH "${PROJECT_SOURCE_DIR}/build_3rdparty/libwebsockets" CACHE PATH "Location of libwebsockets build tree")

# In my project I fetch the same commit as specified in the submodule for libwebsocket here
# when using option 2. But this can be hard coded to some commit or tag as well of course.
# (used for option 1, when AUTO_LWS is ON)
set(LIBWEBSOCKETS_TAG "2.2" CACHE STRING "Libwebsockets commit hash to build)

....

if (AUTO_LWS)
   # This method I use when cross compiling usually, since then all dependencies
   # can be built in one go using the exact same CMake toolchain file
   # that is passed here to the subproject...

   if (CMAKE_TOOLCHAIN_FILE)
    set(EXTERNAL_TOOLCHAIN "-DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}")
   endif()

    include(ExternalProject)

    # This makes CMake checkout the project using git, and then build it
    # using the same cmake generator as used to build the main project
    ExternalProject_Add(libwebsockets-git
        GIT_REPOSITORY "https://github.com/warmcat/libwebsockets.git"
        GIT_TAG "${LIBWEBSOCKETS_TAG}"
        CMAKE_GENERATOR "${CMAKE_GENERATOR}"
        CMAKE_ARGS ${EXTERNAL_TOOLCHAIN} ${EXTRA_EXTERNAL_ARGS} -DLWS_SSL_CLIENT_USE_OS_CA_CERTS=ON -DLWS_WITHOUT_TESTAPPS=ON -DCMAKE_BUILD_TYPE=MinSizeRel
        INSTALL_COMMAND ""
        )

    add_dependencies(theprogram libwebsockets-git)

    ExternalProject_Get_Property(libwebsockets-git SOURCE_DIR)
    ExternalProject_Get_Property(libwebsockets-git BINARY_DIR)

    set(LIBWEBSOCKETS_INCLUDE_DIRS
        "${SOURCE_DIR}/lib/"
        "${BINARY_DIR}")

    find_package(ZLIB REQUIRED)

    set(LIBWEBSOCKETS_STATIC_LIBRARIES "${BINARY_DIR}/lib/${CMAKE_CFG_INTDIR}/libwebsockets.a" ${ZLIB_LIBRARIES})

elseif (INTERNAL_LWS)
    # Find the version built in the project dir.
    find_package(Libwebsockets REQUIRED CONFIG NO_DEFAULT_PATH NO_CMAKE_PACKAGE_REGISTRY
                PATHS ${LIBWEBSOCKETS_PATH})

    # TODO: Move these to Libwebsockets itself.
    set(LIBWEBSOCKETS_STATIC_LIBRARIES websockets)
    set(LIBWEBSOCKETS_DYNAMIC_LIBRARIES websockets_shared)
else()
    # Find system version
    find_package(Libwebsockets REQUIRED)
endif()

# Finally we can use it.
include_directories(${LIBWEBSOCKETS_INCLUDE_DIRS})
target_link_libraries(theprogram ${LIBWEBSOCKETS_STATIC_LIBRARIES})
```

PR #861, opened 2017-04-06
ssl: fix OpenSSL client method detection

mmilata:
Commit 00081a2 breaks libwebsockets on my system that has no `TLS_client_method` function. Please check the ordering in ssl-client.c if it's what you intended.


lws-team:
Thanks... the change was caused by updating my build machine to Fedora 26 Alpha whith OpenSSL 1.1 now... unfortunately the wrong logic I managed to put in worked on that, Appveyor and Travis so I was none the wiser.

It's pushed on master. v2.1-stable and v2.2-stable (seemed the same patch could do on all... did I miss something else?)

mmilata:
I think that is all. Thanks!

PR #865, opened 2017-04-06
Spinning on poll() when closing SSL connection

### LINKED ISSUE #831: SSL shutdown may cause poll to constantly trigger
I'm seeing an issue with the SSL shutdown, where the client enters a situation where the poll is always triggering with POLLIN (1).

```
[ CLIENT ]DEBUG : 0 : LWS : LWS_CALLBACK_RECEIVE_CLIENT_HTTP closed it
[ CLIENT ]DEBUG : 0 : LWS : Close and handled
[ CLIENT ]LwsCallback: wsi=0x8985e68, reason=LWS_CALLBACK_CLOSED_CLIENT_HTTP, user=0x81eec00, in=(nil), len=0
[ CLIENT ]INFO : 0 : LWS : lws_close_free_wsi: shutting down SSL connection: 0x8985e68 (ssl 0x8986190, sock 6, state 6)
[ CLIENT ]DEBUG : 0 : LWS : lws_set_timeout: 0x8985e68: 20 secs
[2017/03/09 10:52:54:8812] INFO: lws_http_transaction_completed: wsi 0x88ac0b8
[2017/03/09 10:52:54:8816] INFO: lws_http_transaction_completed: 0x88ac0b8: close connection
[2017/03/09 10:52:54:8816] INFO: writeable_fail
[2017/03/09 10:52:54:8816] INFO: lws_close_free_wsi: shutting down SSL connection: 0x88ac0b8 (ssl 0x88ac268, sock 8, state 3)
[ CLIENT ]DEBUG : 0 : LWS : fd=6, revents=1
[ CLIENT ]DEBUG : 0 : LWS : fd=6, revents=1
[ CLIENT ]DEBUG : 0 : LWS : fd=6, revents=1
[ CLIENT ]DEBUG : 0 : LWS : fd=6, revents=1
[ CLIENT ]DEBUG : 0 : LWS : fd=6, revents=1
[ CLIENT ]DEBUG : 0 : LWS : fd=6, revents=1
[ CLIENT ]DEBUG : 0 : LWS : fd=6, revents=1
[ CLIENT ]DEBUG : 0 : LWS : fd=6, revents=1
[ CLIENT ]DEBUG : 0 : LWS : fd=6, revents=1
[ CLIENT ]DEBUG : 0 : LWS : fd=6, revents=1
```

This can be "fixed" by not using the `SSL_shutdown`, and just calling `shutdown` in `lws_close_free_wsi`; I'm not yet sure what the proper fix is.

I'm only seeing sporadically when running this on a slower PC.

I suspect that it is due to the non-blocking sockets, as `SSL_shutdown` may fail if the read or write would block and return an error code, expecting to be called again.

mmilata:
Unfortunately the commit 4198c20 that is supposed to fix #831 actually causes the poll() spinning issue in my use case. What seems to be happening is that LWS sends connection close notification in the SSL layer and the other side closes the connection immediately without sending a reply (`SSL_shutdown` [returns -1](https://wiki.openssl.org/index.php/Manual:SSL_shutdown(3)#RETURN_VALUES) with `SSL_get_error` returning `SSL_ERROR_SYSCALL`). Using strace I can see these two syscalls alternating in a busy loop (until 20s timeout):
```
poll([{fd=6, events=POLLIN}], 1, 500) = 1 ([{fd=6, revents=POLLIN}])
read(6, "", 5)                        = 0
```

Here's a helpful explanation how `SSL_shutdown` should be used in non-blocking mode: http://stackoverflow.com/questions/28056056/handling-ssl-shutdown-correctly/28056464#28056464

This pull request improves the behavior in my use case but I think the code is still not correct wrt the post above and probably shouldn't be merged as it is now. I don't really understand the high-level control flow inside of LWS so I'll appreciate any hints in this direction.

### DIFF
--- a/lib/libwebsockets.c
+++ b/lib/libwebsockets.c
@@ -510,14 +510,28 @@ just_kill_connection:
 	    !wsi->socket_is_permanently_unusable) {
 #ifdef LWS_OPENSSL_SUPPORT
 		if (lws_is_ssl(wsi) && wsi->ssl)
-               {
-                       lwsl_info("%s: shutting down SSL connection: %p (ssl %p, sock %d, state %d)\n", __func__, wsi, wsi->ssl, (int)(long)wsi->desc.sockfd, wsi->state);
+		{
+			lwsl_info("%s: shutting down SSL connection: %p (ssl %p, sock %d, state %d)\n", __func__, wsi, wsi->ssl, (int)(long)wsi->desc.sockfd, wsi->state);
 			n = SSL_shutdown(wsi->ssl);
-                       if (n == 1) /* If finished the SSL shutdown, then do socket shutdown, else need to retry SSL shutdown */
-                               n = shutdown(wsi->desc.sockfd, SHUT_WR);
-                       else
-                               lws_change_pollfd(wsi, LWS_POLLOUT, LWS_POLLIN);
-               }
+			if (n == 1) /* If finished the SSL shutdown, then do socket shutdown, else need to retry SSL shutdown */
+				n = shutdown(wsi->desc.sockfd, SHUT_WR);
+			else if (n == 0)
+				lws_change_pollfd(wsi, LWS_POLLOUT, LWS_POLLIN);
+			else /* n < 0 */
+			{
+				int shutdown_error = SSL_get_error(wsi->ssl, n);
+				lwsl_debug("SSL_shutdown returned %d, SSL_get_error: %d\n", n, shutdown_error);
+				if (shutdown_error == SSL_ERROR_WANT_READ) {
+					lws_change_pollfd(wsi, LWS_POLLOUT, LWS_POLLIN);
+					n = 0;
+				} else if (shutdown_error == SSL_ERROR_WANT_WRITE) {
+					lws_change_pollfd(wsi, LWS_POLLOUT, LWS_POLLOUT);
+					n = 0;
+				} else { // actual error occurred, just close the connection
+					n = shutdown(wsi->desc.sockfd, SHUT_WR);
+				}
+			}
+		}
 		else
 #endif
 		{
--- a/lib/service.c
+++ b/lib/service.c
@@ -911,22 +911,40 @@ lws_service_fd_tsi(struct lws_context *context, struct lws_pollfd *pollfd, int t
 
 
 #ifdef LWS_OPENSSL_SUPPORT
-       if ((wsi->state == LWSS_SHUTDOWN) && lws_is_ssl(wsi) && wsi->ssl)
-       {
-               n = SSL_shutdown(wsi->ssl);
-               lwsl_debug("SSH_shutdown=%d for fd %d\n", n, wsi->desc.sockfd);
-               if (n == 1)
-               {
-                       n = shutdown(wsi->desc.sockfd, SHUT_WR);
-                       goto close_and_handled;
-               }
-               else
-               {
-                       lws_change_pollfd(wsi, LWS_POLLOUT, LWS_POLLIN);
-                       n = 0;
-                       goto handled;
-               }
-       }
+	if ((wsi->state == LWSS_SHUTDOWN) && lws_is_ssl(wsi) && wsi->ssl)
+	{
+		n = SSL_shutdown(wsi->ssl);
+		lwsl_debug("SSL_shutdown=%d for fd %d\n", n, wsi->desc.sockfd);
+		if (n == 1)
+		{
+			n = shutdown(wsi->desc.sockfd, SHUT_WR);
+			goto close_and_handled;
+		}
+		else if (n == 0)
+		{
+			lws_change_pollfd(wsi, LWS_POLLOUT, LWS_POLLIN);
+			n = 0;
+			goto handled;
+		}
+		else /* n < 0 */
+		{
+			int shutdown_error = SSL_get_error(wsi->ssl, n);
+			lwsl_debug("SSL_shutdown returned %d, SSL_get_error: %d\n", n, shutdown_error);
+			if (shutdown_error == SSL_ERROR_WANT_READ) {
+				lws_change_pollfd(wsi, LWS_POLLOUT, LWS_POLLIN);
+				n = 0;
+				goto handled;
+			} else if (shutdown_error == SSL_ERROR_WANT_WRITE) {
+				lws_change_pollfd(wsi, LWS_POLLOUT, LWS_POLLOUT);
+				n = 0;
+				goto handled;
+			}
+
+			// actual error occurred, just close the connection
+			n = shutdown(wsi->desc.sockfd, SHUT_WR);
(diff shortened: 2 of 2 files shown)


lib/libwebsockets.c
```diff
@@ -510,14 +510,28 @@ lws_close_free_wsi(struct lws *wsi, enum lws_close_status reason)
...
 			n = SSL_shutdown(wsi->ssl);
-                       if (n == 1) /* If finished the SSL shutdown, then do socket shutdown, else need to retry SSL shutdown */
-                               n = shutdown(wsi->desc.sockfd, SHUT_WR);
-                       else
-                               lws_change_pollfd(wsi, LWS_POLLOUT, LWS_POLLIN);
-               }
+			if (n == 1) /* If finished the SSL shutdown, then do socket shutdown, else need to retry SSL shutdown */
+				n = shutdown(wsi->desc.sockfd, SHUT_WR);
+			else if (n == 0)
+				lws_change_pollfd(wsi, LWS_POLLOUT, LWS_POLLIN);
```
mmilata: Not really sure what to do in the `n == 1` case. If we want to wait for the other side to close connection we need to call `SSL_shutdown` again. However we probably shouldn't just add `LWS_POLLIN` as we do here.

lws-team:
First just noting although the spinning is bad, this manages to close OK at the timeout.  So the symptom is just the spinning.

If it was waiting for, eg, WANT_WRITE, assuming we come back and do the SSL_shutdown() in the loop, that gives it many opportunities while spinning for the timeout period to discover the connection became writeable.  But it doesn't, so the problem either isn't WANT_WRITE and / or we don't redo the SSL_shutdown() inside the loop.

If it was waiting for WANT_READ, it will evidently never receive anything after getting in this situation, although it is signalling POLLIN.  I think POLLIN with 0 length read is indicating the remote peer has closed the connection in this case.  Then, the root cause is a false signalling of POLLIN continuously due to the closed condition of the connection.

If that's true the first time we can see that happened we should forcibly close the connection and destroy the wsi.  But since it's not our code doing the read(), we somehow have to identify that has happened from the openssl return code.

So maybe try

1) confirm once we enter the SHUTDOWN state, we are spinning calling SSL_shutdown() and not SSL_read() we would normally do

2) If we see SYSCALL as the error, check errno for something uniquely identifying the situation the remote peer closed

3) Try checking for SYSCALL / errno showing it closed and then set a flag indicating it happened and do the close processing with no further delays

lib/libwebsockets.c
skyhisi: The OpenSSL manual says that a return code of 1 means that both sides have done the SSL "close notify". I don't think you need to call `SSL_shutdown` again.

skyhisi:
I haven't tried this patch yet, but the changes look reasonable; if I have some time I'll see if they work for me.

In the code I'm working on, as a work around for these sorts of bugs, I've added a "libwebsockets throttle" to ensure that if one of these busy loops is triggered it won't cause CPU starvation for 20 seconds. Basically every call to call to `lws_service` increments a counter that is reset once a second, when the counter is over 50, after doing the service it'll do a sleep for 100ms on my lws handling thread.

lib/libwebsockets.c
mmilata: Right, sorry, meant the `n == 0` case.

skyhisi: When we call `SSL_shutdown` and the other end hasn't sent an SSL close message, we will hit the `n == 0` case, so we are now in the state of having sent an SSL close message and we are giving the other side a chance to acknowledge the close with their own SSL close. This is why we start waiting for the reply, we can't do the second `SSL_shutdown` call in the `n == 0` case here, as the sockets are non-blocking, so there won't have been time for the message to be sent and a reply received.
When the reply comes in, this will trigger the code in `lws_service_fd_tsi` to do the next `SSL_shutdown` call which should then return 1.

lws-team: Guys going by the description in this issue, the root cause is we don't detect the peer has gone (by SYSCALL / errno).  All other sequencing is meaningless  that situation if that's the case.

If it's not that the peer has dropped, it won't busyloop, because it's the false POLLIN indicating the peer dropped that drives that.

If it's not interesting to confirm SYSCALL with a specific errno let me know how to reproduce the problem in the issue with the test server and I'll look at it.

mmilata:
> I think POLLIN with 0 length read is indicating the remote peer has closed the connection in this case. Then, the root cause is a false signalling of POLLIN continuously due to the closed condition of the connection.

I think that's correct.

> confirm once we enter the SHUTDOWN state, we are spinning calling SSL_shutdown() and not SSL_read() we would normally do

The following log message gets logged, that suggests we're indeed spinning on `SSL_shutdown`: `SSH_shutdown=-1 for fd 6` (sic). I think the main issue here is that the original code only checks whether the return value from `SSL_shutdown` is 1 and if it is not it assumes it to be 0. There's no error handling.

> If we see SYSCALL as the error, check errno for something uniquely identifying the situation the remote peer closed

Tried to check errno but it is 0 in this case. I think `read` does not set errno when it indicates end of stream by returning 0. But then again, when SSL_ERROR_SYSCALL is returned can we do anything besides closing the connection?

lib/libwebsockets.c
mmilata: I have a minimal client for reproducing this. I'll try to put together the server side as well.

lws-team:
OK... the docs say SYSCALL may set errno.  But since WANT_ which may inside OpenSSL have initially presented it self as EAGAIN or whatever is already handled separateby by OpenSSL and lws, the approach to try SYSCALL alone as indicating the peer closed and just closing our side should clear the spinning up.

Since we know now we will logically come back to SSL_shutdown, we just need to respond to WANT_WRITE and WANT_READ properly... it looks like that is what your patch is already doing that.  If we add the SYSCALL close stuff on top of your patch, what symptom is left?

lib/libwebsockets.c
mmilata: @lws-team If you want to look at it, you can use [https://github.com/mmilata/libwebsockets/commits/close-connection-after-ssl-shutdown](this branch) to build a server that closes socket right after it receives ssl shutdown notification. The server spins too, though that might be caused by my changes. Also note that the server I've originally run into this problem with is based on node.js.

Here are instructions to reproduce the issue: https://gist.github.com/mmilata/0a054ef99e7235a7954a1a05c9006267

mmilata:
That sounds good, I think there'd be no symptom left.

mmilata:
@lws-team I see that the patch is already included in master, thanks! Is there anything else that needs to be done here?

lws-team:
This open issue sounds like it should be related... https://github.com/warmcat/libwebsockets/issues/875 but I can't reproduce his symptom...

lws-team:
Thanks for the comment on the issue... and the patch...

PR #868, opened 2017-04-10
report back http errors >= 400 with HTTP: prefix

technicianted:
If server returns HTTP error code before WS upgrade, call back LWS_CALLBACK_CLIENT_CONNECTION_ERROR will only report an indicative "HS: no accept". In most cases the client would need to know the HTTP error. For example in case of 403.

This patch adds an early check for HTTP error codes and reports the error with prefix HTTP: so client can differentiate it and parse it out.

### DIFF
--- a/lib/client.c
+++ b/lib/client.c
@@ -415,7 +415,9 @@ lws_client_interpret_server_handshake(struct lws *wsi)
 	int n, len, okay = 0, port = 0, ssl = 0;
 	int close_reason = LWS_CLOSE_STATUS_PROTOCOL_ERR;
 	struct lws_context *context = wsi->context;
-	const char *pc, *prot, *ads = NULL, *path, *cce = NULL;
+	const char *pc, *prot, *ads = NULL, *path;
+	char *cce = NULL;
+	int free_cce = 0;
 	struct allocated_headers *ah = NULL;
 	char *p, *q;
 	char new_path[300];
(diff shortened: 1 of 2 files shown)


lws-team:
asprintf is a gnu-ism... but there should be a better way to handle this.

This API already exists

```
/**
 * lws_http_client_http_response() - get last HTTP response code
 *
 * \param wsi: client connection
 *
 * Returns the last server response code, eg, 200 for client http connections.
 */
LWS_VISIBLE LWS_EXTERN unsigned int
lws_http_client_http_response(struct lws *wsi);
```

He just does this

```
LWS_VISIBLE LWS_EXTERN unsigned int
lws_http_client_http_response(struct lws *wsi)
{
	if (!wsi->u.http.ah)
		return 0;

	return wsi->u.http.ah->http_response;
}
```

We already set that on client path in client.c:478

```
	if (ah)
		ah->http_response = n;
```

If something still needs doing that way would be the way to do it.

technicianted:
`ah` is only set if it is an HTTP request, not WS:

> 

        if (!wsi->do_ws) {
                /* we are being an http client...
                 */
                ah = wsi->u.hdr.ah;
                lws_union_transition(wsi, LWSCM_HTTP_CLIENT_ACCEPTED);
                wsi->state = LWSS_CLIENT_HTTP_ESTABLISHED;
                wsi->u.http.ah = ah;
                ah->http_response = 0;
        }

so it is never set to `n`.
Why is this code path only executed when `!wsi->do_ws` ?

lws-team:
It's because he sees he wants to be an HTTP mode client, not a ws mode client.  So he changes the union and wsi mode to be suitable for that... something you would only do if you are not going to upgrade to ws.

The ah is set before this code... what it is doing in there is lws_union_transition() will clear down the whole union area to zero, including the ah pointer.  So it is saving it to ah before then and putting it back after the union transition; it will detach it by hand later and zero the pointer in the union.

It should be OK to move the `ah = wsi->u.hdr.ah;` before the if()...

technicianted:
abandoning and taking suggest route.

PR #869, opened 2017-04-10
Fix capturing of HTTP errors for WS clients before upgrade

technicianted:
Save copy of ah pointer even with WS client so that HTTP error can be captured by calling lws_http_client_http_response.


lws-team:
Thanks a lot... it's pushed on v2.1-stable, v2.2-stable and master.

PR #895, opened 2017-05-16
client: added socks5 proxy support

andy-ning:
This patch added socks5 proxy support for client. A client can specify socks proxy as following:

struct lws_context *context;
struct lws_context_creation_info context_ci;
struct lws_client_connect_info client_ci;

context_ci.socks_proxy_address = address;  /* socks5 proxy host, could be IP or FQDN */
context_ci.socks_proxy_port = proxy_info.port;  /* port of the proxy server */

context = lws_create_context( &context_ci );
client_ci.context = context;

lws = lws_client_connect_via_info( &client_ci );

### DIFF
--- a/lib/client-handshake.c
+++ b/lib/client-handshake.c
@@ -26,6 +26,7 @@ lws_client_connect_2(struct lws *wsi)
 
 	/* proxy? */
 
+	/* http proxy */
 	if (wsi->vhost->http_proxy_port) {
 		plen = sprintf((char *)pt->serv_buf,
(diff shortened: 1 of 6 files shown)


andy-ning:
It supports no authentication and basic username/password authentication.

In case of username/password authentication, the address passed to context_ci.socks_proxy_address is in the format of username:password@proxyserver<:port>

lws-team:
Thanks a lot, it looks good and is pushed on master.

I made some small changes but mainly added control of whether it builds in CMake, with LWS_WITH_SOCKS5, defaulting to OFF since it's quite a bit of code.

PR #902, opened 2017-05-23
libuv: Fix closing handle multiple times

pparadzik:
Sometimes "Assertion failed: !uv__is_closing(handle)" happens because
handle is being closed multiple times. To fix this, "uv_is_closing"
is added before calling "uv_close".


lws-team:
Thanks... I haven't seen this happen in lwsws, but it costs almost nothing to protect against it as you're doing here.  And if you are sharing the loop with other code it might be a danger.  It's pushed on master and v2.2-stable.

PR #905, opened 2017-05-29
Fixed build in scope of a project (add_subdirectory)

ksergey:
This patch fix CMake build in scope of a project


lws-team:
Thanks... can you explain what problem is this actually fixing?  For eg, esp32 lws already builds two levels down inside another project with its own non-cmake build system.  It's to do with if the outer build system is also cmake somehow?

ksergey:
CMAKE_SOURCE_DIR points to directory of parent project, so lib/libwebsockets.h not exists.  Same for CMAKE_BINARY_DIR.
CMake output for my project:
```
Error copying file "/home/ksergey/dev/bitfinex_feed_handler/lib/libwebsockets.h" to "/home/ksergey/dev/bitfinex_feed_handler/b/include/libwebsockets.h".
```
CMAKE_CURRENT_SOURCE_DIR points to directory where libwebsockets actually placed

ksergey:
Also CMake should automatic import include directories of sub project to parent project in case of parent project linked with target of sub project, but It's not happen in case of libwebsockets. Workaround is include_directories(${LIBWEBSOCKETS_INCLUDE_DIRS}) in parent project.

lws-team:
Thanks... I pushed the first one on v2.2-stable nd master, and the second on on v2.1-stable as well.

PR #907, opened 2017-05-30
client: fix compile error on Visual Studio 12 2013

andy-ning:
Use LWS_INLINE instead of inline in libwebsockets.h to
make it compatible on both Linux and Windows.

Signed-off-by: Andy Ning <andy.ning@windriver.com>


lws-team:
Thanks... it's pushed on master.

PR #911, opened 2017-06-02
Support to bind to device on Linux

leonardorufino:
I need to bind the IP address and interface on Linux.
So I created a flag called _bind_iface_ in order to enable it.

### DIFF
--- a/lib/context.c
+++ b/lib/context.c
@@ -355,6 +355,10 @@ lws_create_vhost(struct lws_context *context,
 		vh->name = info->vhost_name;
 
 	vh->iface = info->iface;
+#if !defined(LWS_WITH_ESP8266) && !defined(LWS_WITH_ESP32) && !defined(OPTEE_TA) && !defined(WIN32)
+	vh->bind_iface = info->bind_iface;
+#endif
+
 	for (vh->count_protocols = 0;
 	     info->protocols[vh->count_protocols].callback;
 	     vh->count_protocols++)
--- a/lib/libwebsockets.h
+++ b/lib/libwebsockets.h
@@ -1787,6 +1787,11 @@ struct lws_context_creation_info {
 	 * the pathname of a UNIX domain socket. you can use the UNIX domain
 	 * sockets in abstract namespace, by prepending an at symbol to the
 	 * socket name. */
+#if !defined(LWS_WITH_ESP8266) && !defined(LWS_WITH_ESP32) && !defined(OPTEE_TA) && !defined(WIN32)
+	int bind_iface;
+	/**< VHOST: 1 to bind to the interface name from iface value (eg, "eth2"),
+	 0 otherwise. */
+#endif
 	const struct lws_protocols *protocols;
 	/**< VHOST: Array of structures listing supported protocols and a protocol-
 	 * specific callback for each one.  The list is ended with an
--- a/lib/lws-plat-unix.c
+++ b/lib/lws-plat-unix.c
@@ -255,6 +255,14 @@ lws_plat_set_socket_options(struct lws_vhost *vhost, int fd)
 #endif
 	}
 
+	if (vhost->bind_iface) {
+		if (setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, vhost->iface,
+				strlen(vhost->iface)) < 0) {
+			lwsl_warn("Failed to bind to device %s\n", vhost->iface);
+			return 1;
+		}
+	}
+
 	/* Disable Nagle */
 	optval = 1;
 #if defined (__sun)
--- a/lib/private-libwebsockets.h
+++ b/lib/private-libwebsockets.h
@@ -854,6 +854,9 @@ struct lws_vhost {
 	struct lws *lserv_wsi;
 	const char *name;
 	const char *iface;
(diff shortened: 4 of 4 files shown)


lws-team:
What's wrong with the existing arrangements via `.iface`?  On my Fedoa box (checked with `$ sudo netstat -pltn | grep 7681` )

```
$ libwebsockets-test-server
...
tcp        0      0 0.0.0.0:7681            0.0.0.0:*               LISTEN      4436/libwebsockets- 
```
```
$ libwebsockets-test-server -i eno1
...
tcp        0      0 192.168.2.229:7681      0.0.0.0:*               LISTEN      4331/libwebsockets-
```
```
$ libwebsockets-test-server -i lo
...
tcp        0      0 127.0.0.1:7681          0.0.0.0:*               LISTEN      4395/libwebsockets- 
```

You can also give it the interface IP in -i for the same result.

leonardorufino:
The problem of not using SO_BINDTODEVICE is that Linux can send and receive packets to/from the right IP, but to/from a different interface.
A simple example is if you connect in loopback (in your case: 192.168.2.229:7681 connects to 192.168.2.229:7681).
Assuming this IP is in eth1, Linux will send over a wrong interface (lo) and you will receive the packet, unless you set SO_BINDTODEVICE.
There are other more complex cases about SO_BINDTODEVICE, but this simple example explains my point.

lws-team:
Yes the existing arrangements just control which interface(s) we listen on.  That is the intention.

When you send a packet, where that goes is dependent on your routing table.  If we set it to listen on `eno1` in my case, in the 192.168.2/24 network, there's never going to be a route in the table to send that out on `lo` == 127/8 or vice-versa (maybe I just miss your point).  And indeed if I use -i lo, I can run the tests in a browser at 127.0.0.1:7681, and with -i eno1, no problem with 192.168.2.229:7681.

How could I reproduce the situation with the problem you're solving?

leonardorufino:
The case is that when you define _-i eno1_, you just bind IP and port 192.168.2.229:7681, not the interface (SO_BINDTODEVICE does this).
Maybe you can see it connecting your browser (running in 192.168.2.229) to 192.168.2.229:7681 and looking in Wireshark by which interface the packets were sent and received (I believe it will show the interface _lo_ and not _eno1_).
The expected result (using SO_BINDTODEVICE) is not receiving the packets and therefore not completing the connection.

lws-team:
OK I will try it later.

But -i binds the **listen socket** to an interface, when the vhost inits.  That works fine because I can see it in netstat; without -i / `.iface` the listen socket accepts from any interface.  At that point, for that purpose, to say it binds the listen socket 'to the interface' or 'to the interface's IP' doesn't have any distinction.

You seem to be talking about binding the accepted socket to restrict its routing possibilities.  AFAIK only 127/8 can travel on lo but I am not certain how IP loopback actually works, so maybe I will learn something.

lws-team:
Well I learned something... according to tcpdump looped-back packets for other interfaces does travel on lo in linux as you say.

If I implement your patch, and run the server as root so it has permissions to apply the SO_BINDTODEVICE on the listen socket... it 'works' in the sense that I can't connect to 192.168.2.229 via lo any more.  Which is the idea I guess.  Binding it to lo with this flag and then coming on 127.0.0.1:7681 is still ok, since it all occurs on lo that it is bound to.

I pushed it on master with several modifications including adding -k switch on test server, and because it needs root, moved the code applying it to only operate only on listen sockets (which are started before we drop root privs), not accepted sockets too which would need permanent root.

Please check it still does what you originally needed it to do.

leonardorufino:
The changes are correct except that I need to set SO_BINDTODEVICE on client sockets as well.
This is necessary, for example, when I have multiple IPs on different interfaces and some of them can access external addresses.
If I try to send a message to an unknown external address, the default gateway will be used and this can be a problem because the IP and the interface of the packet sent may be different.
So, configuring SO_BINDTODEVICE forces the use of the right interface.

lws-team:
Well... running the whole server as root as a fix would be considered an unreasonable tradeoff by most people.

Stuff shouldn't go down the default gateway if there's an explicit route for the network on another local network interface.

Since this is Linux-only though, it looks like there's also a Linux-only security mostly-solution... SO_BINDTODEVICE only seems to need RAW Linux capability.  So if you acquire the capability at listen-socket creation time if any vhost has your flag set, while you are root, and keep it acquired even after the privs are dropped, this should let you run the server as an unprivileged user subsequently and still be able to bind the accepted sockets too.  Can you give it a go?

leonardorufino:
Ok then!

leonardorufino:
Andy, I decided to call SO_BINDTODEVICE on my callback function when LWS_CALLBACK_ADD_POLL_FD arrives.
Then I can bind all sockets (client, server listen and server accept).
You can close this pull request and decide if you revert the 6761180 commit or not.
Thanks.

lws-team:
^^ Adding capabilities was less simple than I was imagining.  But it's generally useful for lws so I added the ability to request generic caps to be inherited by the unprivileged user from the context creation struct, currently it's at https://github.com/warmcat/libwebsockets/commit/b795e79fa7a51cb4442756efb15fcdedca1ac0ab   CMake automatically enables lws support for it if it finds sys/capabilities.h and libcap lying around.

For your patch I adapted the -k switch on the test server to use the capabilities support and confirmed BINDTODEVICE is happy running under an unprivileged user by the capabilities support, even after putting the BINDTODEVICE back where it was so i also applies to accepted sockets.

PR #915, opened 2017-06-05
Support to bind to a specific IPv6 address on Windows

leonardorufino:
On Windows, we could only bind to [::].
This change allows to bind to a specific IPv6 configured in iface.


leonardorufino:
What version of Windows is running on the AppVeyor check?

lws-team:
Thanks... I squashed this into one patch and applied on master.  Since we expect further changes to #911 I kept that as the last patch.

PR #922, opened 2017-06-09
Reduce memory size when client context is created. V1.7 stable

dulguunbatmunkh:
Hello @lws-team, 

`lws_context_init_server_ssl` and `lws_context_init_server` allocates noticeable amount of memory and it can be skipped when the context is created as a client.

Regards,
Dulguun

### DIFF
--- a/lib/context.c
+++ b/lib/context.c
@@ -290,16 +290,20 @@ lws_create_context(struct lws_context_creation_info *info)
 			lws_set_proxy(context, p);
 #endif
 	}
-
-	if (lws_context_init_server_ssl(info, context))
-		goto bail;
-
+	
+	if (info->port != CONTEXT_PORT_NO_LISTEN) {
+		if (lws_context_init_server_ssl(info, context))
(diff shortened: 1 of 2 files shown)


lws-team:
Thanks... only the second of the four commits would be needed.

However there's already a better way to do this, `-DLWS_WITHOUT_SERVER=1` at cmake.

That not only makes those calls a NOP already in private-libwebsockets.h but also snips out most server-related code and data from the library compilation, removes related constant strings etc.

Is it doing the same job?

lws-team:
Well I think LWS_WITHOUT_SERVER=1 does this already, but I am happy to look again if I miss the point.

dulguunbatmunkh:
@lws-team, 

Sorry for the late response on this. 
We have been testing it with our devices. And yes, compiling with -DLWS_WITHOUT_SERVER=1 reduced the memory. 

As always thanks for your kind help.

Best Regards,
Dulguun

Issue #923, opened 2017-06-09
error: ‘mount_protocols’ defined but not used


ksergey:
Error in file "libwebsockets/lib/context.c:43:27"
Build with option(LWS_WITH_NO_LOGS "" ON) and CMAKE_BUILD_TYPE=Release
gcc version 6.2.1 20160916 (Red Hat 6.2.1-3) (GCC)

Could you please fix?

lws-team:
Thanks it should be fixed on v2.2-stable and master.

ksergey:
Looks like not solved, now I got error "error: ‘mount_protocols":
```
/home/ksergey/dev/bitfinex_feed_handler/deps/libwebsockets/lib/context.c: In function ‘lws_create_vhost’:
/home/ksergey/dev/bitfinex_feed_handler/deps/libwebsockets/lib/context.c:444:5: error: ‘mount_protocols’ undeclared (first use in this function)
     mount_protocols[mounts->origin_protocol],
```
Build with option(LWS_WITH_NO_LOGS "" ON) and CMAKE_BUILD_TYPE=Debug

lws-team:
Okie... but can you send me a patch?  Evidently my efforts to fix it from your description aren't getting the job done.

ksergey:
The patch tested on Linux+gcc with flags:
-DCMAKE_BUILD_TYPE=Release -DLWS_WITH_NO_LOGS=ON
-DCMAKE_BUILD_TYPE=Debug -DLWS_WITH_NO_LOGS=ON
-DCMAKE_BUILD_TYPE=Release -DLWS_WITH_NO_LOGS=OFF
-DCMAKE_BUILD_TYPE=Debug -DLWS_WITH_NO_LOGS=OFF

PR #931, opened 2017-06-13
Added LWS_UNUSED

ksergey:
Fix for issue #923


lws-team:
Thanks... I pushed a different patch on master than fixes the problem with (void) references to the three variables that are unused if lwsl_notice() doesn't reference them.  This is a C language feature that will work on all platforms.

PR #940, opened 2017-06-21
Update client-handshake.c

wujunjenny:
Resolve the bug that http client redirect timeout using libuv

### DIFF
--- a/lib/client-handshake.c
+++ b/lib/client-handshake.c
@@ -264,6 +264,9 @@ lws_client_connect_2(struct lws *wsi)
 			}
 		}
 	}
+	else {
+		lws_change_pollfd(wsi, 0, LWS_POLLIN);	
+	}
 
 #ifdef LWS_USE_IPV6
 	if (wsi->ipv6)


lws-team:
Thanks... I am not certain this is the right fix though,

The `else` belongs to

```
	if (!lws_socket_is_valid(wsi->desc.sockfd)) {
```

the patch is saying if we came back there, and the socket fd is already valid, force POLLIN active.

But at the time we made the socketfd valid (the other branch of the same `if` above) we set POLLIN active.

There are three ways to decativate it as far as I can see.

Are you using SSL?

wujunjenny:
I have test with http.
I have not used SSL testing yet .
The fuction name is lws_client_connect_2,I think wsi->desc.sockfd should be deatived.

wujunjenny:
The bug is when lws_socket_is_valid(wsi->desc.sockfd) is true ,"connect"  action will lose the POLLIN and lead to timeout。

lws-team:
Right but I would like to understand why lws removed the POLLIN.

There are only a few reasons lws will do it, one is that it could not find a free ah to parse the returned http headers with (there is a pool of preallocated ah, and if the server part is busy, they may all be in use).  If that is what happened, the wsi joins an ah "waiting list", and when a free ah is available, it is attached to the wsi and POLLIN is enabled again.  wsi who need an ah but can't get one yet stay silent with POLLIN off until they can get one.

If that is what is happening, we don't want to force POLLIN on.

Maybe I can put it another way, how can I reproduce this problem that POLLIN is off?  Preferably with the test client.

wujunjenny:
may be libuv cause the bug

lws-team:
Doubt it's libuv... it would be making trouble everywhere with this kind of problem... can you give me an idea how to reproduce it?

You mentioned no SSL, and you're using libuv which is useful to know.  Are you running as a server as well as client?  The client connection works if there is no redirect?

wujunjenny:
The client works fine with no redirect
and it works fine with redirect not using libuv

wujunjenny:
I build project under windows 7

wujunjenny:
The client works fine with no redirect
and it works fine with redirect not using libuv

lws-team:
Thanks... that info should help a lot.  I'll have a go at reproducing it tomorrow.

lws-team:
Does the patch mentioned above help?

lws-team:
Can you let me know if that helped?

wujunjenny:
Today I will test to see if it works

lws-team:
Thanks... I had another look at the patch I sent before... I think it's the right idea, but I added a little more code to handle the case libuv didn't close the watcher before we retry the connection.  I am not sure if we trigger that in reality or not.  Anyway please let me know what it does, if it crashes please send a backtrace.

PR #959, opened 2017-07-06
libuv: fix casting from pointer to integer of different size

pparadzik:
When building for platforms where sizeof 'long long' is 8, but sizeof
pointer is 4 bytes, we get 'cast to pointer from integer of different size
[-Werror=int-to-pointer-cast]' and vice versa. Instead of 'long long'
type, 'intptr_t' is used while casting.

### DIFF
--- a/lib/libuv.c
+++ b/lib/libuv.c
@@ -529,7 +529,7 @@ lws_libuv_closehandle(struct lws *wsi)
 static void
 lws_libuv_closewsi_m(uv_handle_t* handle)
 {
-	lws_sockfd_type sockfd = (lws_sockfd_type)(long long)handle->data;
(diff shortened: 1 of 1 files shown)


lws-team:
Thanks... I generalized this to a) check intptr_t exists in cmake, if not then stick with unsigned long long, and b) apply the new lws_intptr_t to everywhere (hopefully) that does ptr - int conversions.

It has needed this for a while since at various times I thought `long` was enough (it is in the Linux kernel and most platforms), `size_t` was enough (not on all  windows...) and most recently `long long`... it is enough but now some compilers complain it is too much.

You may have to delete build/CMakeCache.txt to get the cmake test to be run.

PR #966, opened 2017-07-17
Eliminated 'unused variable' compiler warning generated with -DLWS_WI…

bopwerks:
…TH_NO_LOGS=ON.

The unused variable was only declared for use in a log macro that's
compiled out with the above compiler switch. I removed the declaration
and casted the variable at each use in the block.


lws-team:
Thanks... I modified it to use a `(void)` cast on the var that's only used with logging on.

And I added another test to travis with this option enabled, so I will notice most of these first from now on.

PR #987, opened 2017-08-12
send content-type when LWS_WITH_RANGES=OFF

lnmx:
With the RANGES feature disabled, lws_serve_http_file would
not add the content-type header to the response.


lws-team:
Thanks... I modified it slightly to reduce the amount of code duplicated in the #if, and pushed on v2.2-stable / v2.3-stable / master.

PR #990, opened 2017-08-16
Mismatched lws_zalloc, free => lws_zalloc, lws_free.

unprompted:
This causes problems if lws_set_allocator is used.  lwsp is allocated on line 488.


lws-team:
Thanks... it's pushed on v2.2-stable, v2.3-stable and master.

PR #992, opened 2017-08-17
Add getifaddrs/freeifaddrs include headers

fathyb:
If not, LLVM will generate "warning: implicit declaration of function 'getifaddrs' is invalid in C99" on Android with `LWS_HAVE_GETIFADDRS=1`

### DIFF
--- a/lib/lws-plat-unix.c
+++ b/lib/lws-plat-unix.c
@@ -1,4 +1,5 @@
 #include "private-libwebsockets.h"
+#include "getifaddrs.h"
 
 #include <pwd.h>
 #include <grp.h>


JoakimSoderberg:
That does not make much sense.

(Note I have never compiled on Android myself)....

h1. Why it does not make sense

From CMakeLists.txt (This does system introspection on the system to check if `getifaddrs` exists natively on the system):
```cmake
...
CHECK_FUNCTION_EXISTS(getifaddrs LWS_HAVE_GETIFADDRS)
...
```

Later this is done:
```cmake
if (UNIX)
	if (NOT LWS_HAVE_GETIFADDRS)
		list(APPEND HDR_PRIVATE lib/getifaddrs.h)
		list(APPEND SOURCES lib/getifaddrs.c)
	endif()
endif()
```
https://github.com/warmcat/libwebsockets/blob/4b24369d6401efa6bb9dfdeedfb76abf3f390cea/CMakeLists.txt#L723-L728

So does `getifaddrs` exist natively on Android or not? If it does exist, `getifaddrs.*` should not be compiled or included. If it does not exist you need it.

In any case that include should not be there for all platforms, so you should have it in an `ifdef`. But I'm assuming this is being included somewhere else already?

```c
#ifndef LWS_HAVE_GETIFADDRS
#include "getifaddrs.h"
#endif
```

h1. This should work looking at Android.mk

Ah but ok... It turns out the `Android.mk` includes `getifaddrs.c` and therefore should include the proper header as is in `lib/private-libwebsockets.h`:
```c
#if defined(LWS_BUILTIN_GETIFADDRS)
 #include <getifaddrs.h>
#else
```
https://github.com/warmcat/libwebsockets/blob/ba45f7cf9f29030155970b3ad002e62e7da0e504/lib/private-libwebsockets.h#L188-L190
`getifaddrs.c` in Makefile
https://github.com/warmcat/libwebsockets/blob/4b24369d6401efa6bb9dfdeedfb76abf3f390cea/Android.mk#L21

`Android.mk` sets this define and therefore it should be included:
```make
LOCAL_CFLAGS    := -DLWS_BUILTIN_GETIFADDRS
```
https://github.com/warmcat/libwebsockets/blob/4b24369d6401efa6bb9dfdeedfb76abf3f390cea/Android.mk#L9

The define in question is `LWS_BUILTIN_GETIFADDRS`... Which in CMake is set here:
```cmake
if (NOT LWS_HAVE_GETIFADDRS)
	if (LWS_WITHOUT_BUILTIN_GETIFADDRS)
		message(FATAL_ERROR "No getifaddrs was found on the system. Turn off the LWS_WITHOUT_BUILTIN_GETIFADDRS compile option to use the supplied BSD version.")
	endif()
	set(LWS_BUILTIN_GETIFADDRS 1)
endif()
```
https://github.com/warmcat/libwebsockets/blob/4b24369d6401efa6bb9dfdeedfb76abf3f390cea/CMakeLists.txt#L477-L482

JoakimSoderberg:
Relates to this PR https://github.com/warmcat/libwebsockets/pull/993

lws-team:
Hum is it meant to be #ifdef ?

```
#ifndef LWS_HAVE_GETIFADDRS
#include "getifaddrs.h"
#endif
```

JoakimSoderberg:
You can disregard that part of my message... I was still investigating while writing that :)

I mean I hadn't looked at how this worked at all, but it just made no sense to include `getifaddrs.h` on all unix platforms.

But otherwise the logic would be 
> "If getifaddrs does not exist on the system include `getifaddrs.h`"

which is what this would do:
```c
#ifndef LWS_HAVE_GETIFADDRS
```

But I then noticed `LWS_HAVE_GETIFADDRS` is not passed on as a define to gcc directly. But rather the other define `DLWS_BUILTIN_GETIFADDRS` is set in CMake and used in `lws-private.h`.

...

Sorry for the confusion and wall-of-text

lws-team:
No I appreciate the help...

@fathyb are you actually using cmake?  I know some other Android users were tempted to build it "by hand".

If you are using cmake, it sounds from the patch description that cmake failed to find getifaddrs.h on your system.  But it's supposed to fall back to an implementation provided by lws then.

fathyb:
Well sorry guys I was confused and messed up my message, I was actually using `LWS_HAVE_GETIFADDRS=0`.

@lws-team yes and no. I use CMake by hand, I have generated a `lws_config.h` and manually set the source files to be built. This is because I have 6 target platforms and they don't all use CMake (I use CocoaPods for iOS and Mac for example). 

So I think you're right, as I don't use CMake to generated platform-specific configuration it does not add the header.

I'm sorry for opening straight-up pull requests. I'll open an issue if the problem persists. Thanks.

PR #993, opened 2017-08-17
Prevent ifaddrs redefinitions on new Android NDKs

fathyb:
Before this PR this errors may be thrown by the Android NDK :

`getifaddrs.h:58:8: error: redefinition of 'ifaddrs'`
`getifaddrs.c:163:11: error: no member named 'ifa_dstaddr' in 'struct ifaddrs'; did you mean 'ifa_addr'`

While the struct is already defined, Android seems to not fully support all functionalities needed by LWS.

This workaround will redefine the struct as a macro and should only affect Android builds.

### DIFF
--- a/lib/getifaddrs.h
+++ b/lib/getifaddrs.h
@@ -54,7 +54,12 @@ extern "C" {
 
 #undef ifa_dstaddr
 
+#if defined(ANDROID)
+#define ifaddrs __lws_ifaddrs
+struct __lws_ifaddrs {
+#else
 struct ifaddrs {
+#endif
 	struct ifaddrs *ifa_next;
 	char *ifa_name;
 	unsigned int ifa_flags;


JoakimSoderberg:
But if it says redefinition is not allowed, it seems your version of Android _DOES_ include a native `getifaddrs`? So have you tried removing `getifaddrs.c` from `Android.mk` and not setting the compile flag:
https://github.com/warmcat/libwebsockets/blob/4b24369d6401efa6bb9dfdeedfb76abf3f390cea/Android.mk#L21

https://github.com/warmcat/libwebsockets/blob/4b24369d6401efa6bb9dfdeedfb76abf3f390cea/Android.mk#L9

fathyb:
Thanks for the quick response!

Gosh I feel stupid (should've opened an issue first), now the compilation works but the linking fails saying that "undefined reference to `getifaddrs`" and "undefined reference to `freeifaddrs`".

Exact link log :
```
.../libwebsockets/lib/lws-plat-unix.c.o: In function `lws_interface_to_sa':
.../libwebsockets/lib/lws-plat-unix.c:562: undefined reference to `getifaddrs'
.../libwebsockets/lib/lws-plat-unix.c:603: undefined reference to `freeifaddrs'
.../libwebsockets/lib/lws-plat-unix.c:603: undefined reference to `freeifaddrs'
.../libwebsockets/lib/lws-plat-unix.c:603: undefined reference to `freeifaddrs'
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
ninja: build stopped: subcommand failed.
```

I don't really know Android myself and I've failed miserably when looking for getifaddrs/freeifaddrs support on Android. Some says it is supported, some says it is not. And I'm worried it may depend on the devices's stdlib (Bionic most of the cases).

PR #1000, opened 2017-08-25
Cgi chunking and fixes

pparadzik:
This set of patches fixes some of the issues in CGI handling and add support for manual chunking of CGI's which don't have content-length nor are explicitly chunked. It's done by making lws get POLLHUP event on CGI stdout (which it didn't before), and after that, sending terminating chunk to the client. POLLHUP enables lws to do reaping immediately after CGI finishes so there is no delay. This makes clients that are relying on this kind of CGI's more responsive.

### DIFF
--- a/lib/context.c
+++ b/lib/context.c
@@ -256,6 +256,13 @@ lws_callback_http_dummy(struct lws *wsi, enum lws_callback_reasons reason,
 				wsi->reason_bf &= ~1;
 			break;
 		}
+
+		if (wsi->reason_bf & 4) {
+			n = lws_write(wsi, (unsigned char *)"0\x0d\x0a\x0d\x0a", 5, LWS_WRITE_HTTP);
+			if (n < 0)
+				return -1;
+			break;
+		}
 #endif
 #if defined(LWS_WITH_HTTP_PROXY)
 		if (wsi->reason_bf & 2) {
@@ -375,6 +382,11 @@ lws_callback_http_dummy(struct lws *wsi, enum lws_callback_reasons reason,
 		break;
 
 	case LWS_CALLBACK_CGI_TERMINATED:
+		if (!wsi->cgi->explicitly_chunked && !wsi->cgi->content_length) {
+			/* send terminating chunk */
+			wsi->reason_bf |= 4;
+			lws_callback_on_writable(wsi);
+		}
 		return -1;
 
 	case LWS_CALLBACK_CGI_STDIN_DATA:  /* POST body for stdin */
--- a/lib/libwebsockets.c
+++ b/lib/libwebsockets.c
@@ -2644,6 +2644,8 @@ lws_cgi(struct lws *wsi, const char * const *exec_array, int script_uri_path_len
 			p++;
 		}
 	}
+	env_array[n++] = "PATH=/bin:/usr/bin:/usr/local/bin:/var/www/cgi-bin";
+
 	env_array[n++] = p;
 	p += lws_snprintf(p, end - p, "SCRIPT_PATH=%s", exec_array[0]) + 1;
 
@@ -2658,7 +2660,6 @@ lws_cgi(struct lws *wsi, const char * const *exec_array, int script_uri_path_len
 	}
 
 	env_array[n++] = "SERVER_SOFTWARE=libwebsockets";
-	env_array[n++] = "PATH=/bin:/usr/bin:/usr/local/bin:/var/www/cgi-bin";
 	env_array[n] = NULL;
 
 #if 0
--- a/lib/libwebsockets.c
+++ b/lib/libwebsockets.c
@@ -256,6 +256,7 @@ lws_bind_protocol(struct lws *wsi, const struct lws_protocols *p)
 	return 0;
 }
 
+#ifdef LWS_WITH_CGI
 static void
 lws_cgi_remove_and_kill(struct lws *wsi)
 {
@@ -280,6 +281,7 @@ lws_cgi_remove_and_kill(struct lws *wsi)
 	wsi->cgi->being_closed = 1;
 	lws_cgi_kill(wsi);
 }
+#endif
 
 void
 lws_close_free_wsi(struct lws *wsi, enum lws_close_status reason)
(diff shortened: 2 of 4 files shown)


lws-team:
Great job... thanks a lot... I pushed them with two small changes

 - squashed the LWS_WITH_CGI fix to the patch that introduced the code

 - added a patch to convert the `reason_bf` constants to names like `LWS_CB_REASON_AUX_BF__CGI` and change everything to use those.

lws-team:
Please take a look at https://github.com/warmcat/libwebsockets/commit/24dd1723f4d1dffd80905a2d624e07080dcad34e

This fixes busylooping and problems around clean completion of CGI file transfers for me.  It also finally wires up stdout POLLHUP detection properly.

Do your usecases remain in good (or better...) shape with it?

pparadzik:
Hi,

It seems that commit [24dd172](https://github.com/warmcat/libwebsockets/commit/24dd1723f4d1dffd80905a2d624e07080dcad34e) introduced some bugs - it's closing connection too soon in some cases, and in the others I'm getting partial response. I think one issue is introduced in this line: `wsi->reason_bf &= ~(LWS_CB_REASON_AUX_BF__CGI_HEADERS | LWS_CB_REASON_AUX_BF__CGI);`. In some cases I get `LWS_CB_REASON_AUX_BF__CGI` + `LWS_CB_REASON_AUX_BF__CGI_HEADERS` (`reason_bf = 9`) flags set. lws does one write and zeroes both flags, even though the other didn't get processed. Also, it can happen that `LWS_CB_REASON_AUX_BF__CGI_CHUNK_END` is set together with `LWS_CB_REASON_AUX_BF__CGI` flag (`reason_bf = 5`). In that case I get partial response, unless I skip `if (wsi->reason_bf & LWS_CB_REASON_AUX_BF__CGI_CHUNK_END) {...}` block. So, this fixed issues for me

```
diff --git a/lib/context.c b/lib/context.c
index faa4fd1..03f5125 100644
--- a/lib/context.c
+++ b/lib/context.c
@@ -254,7 +254,13 @@ lws_callback_http_dummy(struct lws *wsi, enum lws_callback_reasons reason,
                        }
                        if (!n)
                                lws_rx_flow_control(wsi->cgi->stdwsi[LWS_STDOUT], 1);
-                       wsi->reason_bf &= ~(LWS_CB_REASON_AUX_BF__CGI_HEADERS | LWS_CB_REASON_AUX_BF__CGI);
+
+                       if (wsi->reason_bf & LWS_CB_REASON_AUX_BF__CGI_HEADERS)
+                                       wsi->reason_bf &= ~LWS_CB_REASON_AUX_BF__CGI_HEADERS;
+                       else
+                                       wsi->reason_bf &= ~LWS_CB_REASON_AUX_BF__CGI;
+                       break;
+
                        if (!(wsi->reason_bf & LWS_CB_REASON_AUX_BF__CGI_CHUNK_END))
                                break;
                }
```

lws-team:
I'm testing these against cgit and mailman cgis on https://warmcat.com and https://libwebsockets.org... as of 4320938 (updated since 24dd172) these (eg https://warmcat.com/git/libwebsockets/commit/) work well with browser or wget, with no busylooping.

I put your patch above in, and they continue to serve OK, but there is busylooping / 100% cpu in lws then during the cgi lifetime... just removing your patch that busylooping is gone.

Can you let me know how to reproduce the kind of test case you have trouble with?  Eg with a bash script spawned as the cgi?  I'll try to understand why it benefits (indirectly...) from the busylooping and look for an alternative way.

lws-team:
Ah never mind... I tested it in the middle of the night on the actual server, I can't reproduce the cpu load problem thismorning.  So I squashed your fix into yesterday's patch, thanks a lot.

pparadzik:
I'm testing CGI with [LuCI](https://github.com/openwrt/luci), on the device that is very weak. I can't run libwebsockets with the debug flag on because everything is very slow then and these issues happen only on the device in some cases. I'm also using external polling library with lws and everytime when I get some callback related to CGI (like `LWS_CALLBACK_CGI`), I'm calling `lws_callback_http_dummy` function to handle CGI for me. I know that this is not very helpful, but it's little difficult to debug. Will try to get some more logs if I can.

I noticed that you can remove `if` block after break because it will never be executed. Like this
```
diff --git a/lib/context.c b/lib/context.c
index 2a6f620..746a8c0 100644
--- a/lib/context.c
+++ b/lib/context.c
@@ -260,9 +260,6 @@ lws_callback_http_dummy(struct lws *wsi, enum lws_callback_reasons reason,
                        else
                                wsi->reason_bf &= ~LWS_CB_REASON_AUX_BF__CGI;
                        break;
-
-                       if (!(wsi->reason_bf & LWS_CB_REASON_AUX_BF__CGI_CHUNK_END))
-                               break;
                }
 
                if (wsi->reason_bf & LWS_CB_REASON_AUX_BF__CGI_CHUNK_END) {
```

lws-team:
OK.  I can't make any problem any more with cgit + mailman, so if there's more info or even better a way to provoke it with a dummy bash script cgi or whatever and sleeps, I'll be happy to look at it.

That stanza was necessary at one point but I think getting rid of the spinning made it redundant.  I removed it after reordering the patches a bit and will push that later.

lws-team:
Just a thought... with external poll loop, you need to take care of this yourself somehow

```
/**
 * lws_service_adjust_timeout() - Check for any connection needing forced service
 * \param context:	Websocket context
 * \param timeout_ms:	The original poll timeout value.  You can just set this
 *			to 1 if you don't really have a poll timeout.
 * \param tsi: thread service index
 *
 * Under some conditions connections may need service even though there is no
 * pending network action on them, this is "forced service".  For default
 * poll() and libuv / libev, the library takes care of calling this and
 * dealing with it for you.  But for external poll() integration, you need
 * access to the apis.
 *
 * If anybody needs "forced service", returned timeout is zero.  In that case,
 * you can call lws_service_tsi() with a timeout of -1 to only service
 * guys who need forced service.
 */
LWS_VISIBLE LWS_EXTERN int
lws_service_adjust_timeout(struct lws_context *context, int timeout_ms, int tsi);
```

Someone on the ml had problems with external poll they traced to not taking care of this, thought I'd mention it just in case.

pparadzik:
Thanks, I'm already using it.

PR #1002, opened 2017-08-27
fix Spec file by adding / removing new files to the package

stv0g:


stv0g:
Btw. RPM packages are here in case someone is interested:

https://villas.fein-aachen.org/packages/x86_64/

lws-team:
Thanks, it's pushed.

Eg, Fedora also have RPMs in the distro, using their own specfile.

PR #1005, opened 2017-08-28
mbedTLS support

jvijtiuk:
Add support for mbedTLS as a crypto backend for libwebsockets, using the existing esp-idf wrapper.

The ERR_error functions have been moved from lws-plat-esp32.c to the esp-idf ssl wrapper. Some empty functions have also been added to the esp-idf wrapper to fix build errors that appeared when building with mbedTLS

The server and client tests work with SSL and without it, and valgrind reports no memory leaks. 

There was however a memory leak in lws_context_init_server_ssl with the memroy allocated by alloc_pem_to_der_file so i've fixed that although im not sure if the free should happen for ESP32. In the case of ESP32 there is a free that is commented out on line 395, but i couldn't find the reason for the comment.


lws-team:
Thanks... I made some small changes:

1) I squashed the last two patches into the "backend support" one, they just seem to be "history" editing some preprocessor strings when it's better done in place once.

2) added licensing note on commit log of wrapper one (it's Apache 2.0) and amended lws LICENSE to note stuff down that dir has its own liberal license.

3) Building natively on Fedora, I got

```
/projects/libwebsockets/lib/ssl-server.c: In function ‘lws_context_ssl_init_ecdh_curve’:
/projects/libwebsockets/lib/ssl-server.c:108:2: error: unknown type name ‘EC_KEY’; did you mean ‘ENOKEY’?
  EC_KEY *ecdh;
  ^~~~~~
  ENOKEY
...(more errors...)...
```

The problem there is we "find" a local openssl ECC support include on the build machine, which has implications for "real" openssl, but don't understand that is not relevant for mbedtls case.  I patched around it.

4) Since mbedtls is a thing in Fedora as you pointed out CMake should try to find it by default instead of insisting on being told where the libs and includes are.  You can still give it the libs and includes if you, eg, are building cross, to override what it would have found for itself.  I added changes around this.

5) The stubs to do with errors in lws-plat-esp32.c are needed when building for esp32... because they are in the plat file they don't affect any other case.  So I removed the stanza from your patch eliminating them.

6) To remove any chance for confusion I added a patch logging that lws was built for, and using, mbedtls at context creation time.

```
4:  Compiled with MbedTLS support
```

With those changes it builds on Fedora with just -DLWS_USE_MBEDTLS=1 and install `mbedtls-devel`, and it can work for the test app in --ssl at least.  And it also builds for esp32 with some other minor changes needed to match the change of the include passing define from OPENSSL_... to MBEDTLS_...

So nice job ^^  We might find more things to trip over as we go, but clearly this broke the back of it.  Please check and see if with the changes it's still working with your use-case.

jvijtiuk:
I've tested the client and server test apps with the changes and both are still working.

lws-team:
Thanks again, that's it for now then.

PR #1007, opened 2017-08-28
One more free vs. lws_free problem.

unprompted:
Allocated at the top of lws_create_vhost with lws_zalloc(), was freed here with system free().


lws-team:
Thanks... fwiw merge / pull is no good for patch series.  It's much better to rebase your tree on the latest upstream and then add your patches back on top.  If that sounds like black magic, have a look at stgit https://github.com/ctmarinas/stgit ... it's in all the distros already if you are on Linux.

As I understood it only 5cce27b is new... I pushed it on v2.3-stable and master.

PR #1013, opened 2017-09-05
Creation of secure server context without certificate

HarishSRIB:
This patch enables user of libwebsockets library to create server's secure context without providing cert file path. This change is required for systems like where user of libwebsockets library does not have access to file system where certificate is stored or does not provided certificate path for some security reasons and it can get certificate content in binary format from some other module in same system.

### DIFF
--- a/lib/libwebsockets.h
+++ b/lib/libwebsockets.h
@@ -1299,7 +1299,14 @@ enum lws_callback_reasons {
 	/**< Sent to parent to notify them a child is closing / being
 	 * destroyed.  @in is the child wsi.
 	 */
-
+	LWS_CALLBACK_OPENSSL_CONTEXT_REQUIRES_CERT	= 70,
+	/**< if configured for including OpenSSL support but no server cert
+	 * file has been specified (ssl_cert_filepath is NULL), this is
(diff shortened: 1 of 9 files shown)


lws-team:
Can you try eb84e6f?  It should do the same as your patches but is a much smaller change.  I don't think we need to change all the test apps to provide this either, since we use an already-existing callback now that provides the SSL_CTX * for the vhost already.

HarishSRIB:
This (eb84e6f) is what exactly I need :) . BTW test-server-http.c needs use cert_path and key_path as extern in case it has to demonstrate feature of LWS_SERVER_OPTION_CREATE_VHOST_SSL_CTX as certificate and private key file paths are set in main sample files like test-server.c
I am closing this pull request as eb84e6f is doing same thing this patch created for.

lws-team:
Thanks... I don't think we need to provide code in the apps for it... anyone who wants to use it has to know enough openssl apis to use them in the callback anyway (I guess they won't use files there but get the cert from a buffer, after decryption, etc).  For all other cases they can just pass in cert / key filepaths in the info as before and don't need to get involved with it.

PR #1018, opened 2017-09-08
cgi: fix uint64_t formating

pparadzik:
On some platforms 'llu' format is needed for uint64_t instead of 'lu'.
PRIu64 format specifier fixes these platform specific issues.


lws-team:
Thanks... it's pushed. It could be avoided by cast to unsigned long long and %llu... but using PRIu64 is better.

Issue #1032, opened 2017-09-24
Library ordering issue in CMakeLists


kallisti5:
In CmakeLists, the CHECK_FUNCTION_EXISTS calls are made, then much later in the CMakeLists, the platform specific libraries are appended to LIB_LIST and CMAKE_REQUIRED_LIBRARIES is set.

The CHECK_FUNCTION_EXISTS calls actually depend on CMAKE_REQUIRED_LIBRARIES, so it isn't possible to append platform libraries for the function searches unless you hack them in before the CHECK_FUNCTION_EXISTS calls.

To make all of this stuff simpler, the CHECK_FUNCTION_EXISTS calls need to occur after CMAKE_REQUIRED_LIBRARIES.

Example hack
```patch
diff --git a/CMakeLists.txt b/CMakeLists.txt
index ce2bce6..0d1e96d 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -497,6 +497,10 @@ if (LWS_WITHOUT_BUILTIN_SHA1)
        set(LWS_SHA1_USE_OPENSSL_NAME 1)
 endif()

+if (HAIKU)
+    set(CMAKE_REQUIRED_LIBRARIES network)
+endif()
+
CHECK_FUNCTION_EXISTS(bzero, LWS_HAVE_BZERO)
CHECK_FUNCTION_EXISTS(fork, LWS_HAVE_FORK)
CHECK_FUNCTION_EXISTS(getenv, LWS_HAVE_GETENV)
```

lws-team:
I moved the existing ones (which are generic toolchain / libc functions) to after the stuff that sets CMAKE_REQUIRED_LIBRARIES as you suggested.

kallisti5:
Looks good!   I was reluctant to make that change myself since it had a strong potential of breaking "all the things" if done wrong :-)

lws-team:
It did break everything :-)  CMakeCache hid it from me.... I am just editing my patch to unbreak it.

lws-team:
I yanked this patch, see discussion in #1033

kallisti5:
I can confirm the sources as they exist right now compile and run fine on Haiku.

Thanks!

PR #1033, opened 2017-09-24
Haiku: initial platform support

kallisti5:


kallisti5:
There is a minor hack related to #1032

lws-team:
Thanks, I pushed it after snipping the "hack" bit off.

If there's anything special Haiku users should do or know, please let me know so we can add it somewhere in the docs.

kallisti5:
Thanks!  It builds and the tests function normally.

kallisti5:
actually... something seems wrong. I think there are a bunch of LWS_HAVE_XXX checks now above the function tests.

lws-team:
Yeah I moved them already.  But it's dying because it generates

```
/* Define to rpl_malloc if the replacement function should be used. */
#define malloc

/* Define to rpl_realloc if the replacement function should be used. */
#define realloc
```

in build/lws_config_private.h now

kallisti5:
all reasons I was afraid to touch it :-)

* set LIB_LIST   ( no dependencies)
* platform libraries ( no dependencies)
* libs : zlib, ssl,  (HAVE_OPENSSL_ECDH_H)libev, libuv, libevent, sqlite, httpproxy
* set REQUIRED_LIBS from LIB_LIST
* function checks (set LWS_HAVE) (depends on platform libraries)
* SSL function checks

lws-team:
Yeah... people want to use lws in a lot of situations... it's quite flexible and that shows up in CMakeLists.txt.  What should we do... say no to people who turn up wanting to add support to some OS nobody else wanted for 7 years :-) ?  It's not helping telling me how much you're holding your nose.

The current situation looks like a cmake bug actually... at least I can't find any other reason those two cmakedefines are defined instead of commented.  unset(malloc) and unset(realloc) makes it work... maybe something else sets these somewhere...

lws-team:
OK I am removing the 1032 patch and adapting your patch to be as it was posted.

You never explained what was actually wrong with looking for toolchain / stdc defines before we added lws-specific libs; every other platform is OK with that.  If I am looking for `malloc`, I don't need to have openssl or whatever.  And there's plenty of code in CMakeLists.txt that legitimately wants to know the results of those tests earlier.

If your platform has some funny requirement (perhaps cmake doesn't add default paths correctly for it as it does the other platforms) then what you called a "hack" in your patch is actually the correct solution.

kallisti5:
>  It's not helping telling me how much you're holding your nose.

Err, I didn't say or imply that at all. It's a lot of complex library resolution. You should see some of the have statements in mesa's code which supports haiku/irix/aix/solaris/linux/bsd/windows/cygwin/etc :-)

The need is our getifaddrs exists in libnetwork.so and the build is broken without it included. We can call the change as-is the solution. I just thought adding network to the cmake required libraries, then resetting it later might seem "hackish" since the pattern of defining platform libraries before performing function exists checks would ease future platform portability.

PR #1070, opened 2017-10-25
ssl: fix segfault when using SSL_CTX that is not created

pparadzik:
vhost->ssl_ctx is created in 'lws_tls_server_vhost_backend_init', but it
is used before in 'lws_tls_server_client_cert_verify_config' when
calling SSL_CTX_set_session_id_context...


lws-team:

PR #1073, opened 2017-10-26
ssl: fix x509 memory leaks

pparadzik:


lws-team:
Thanks... that's a nice help.  It's pushed.

lws-team:
Actually the first stanza is wrong, `SSL_CTX_get0_certificate()` doesn't allocate, just returns a pointer.  I had been wondering why everything broke with openssl today, it feed part of the the underlying ssl_ctx ;-)

pparadzik:
Sorry, my mistake.

PR #1074, opened 2017-10-26
ssl: fix source and destination overlap in strcpy

pparadzik:
Use memmove instead...


lws-team:
Thanks... I adapted it slightly and added a comment on the union name[] parameter to explain what is
going on there.

PR #1075, opened 2017-10-27
ssl: add support for checking cert existance and verification

pparadzik:


pparadzik:
You can remove this line
```
diff --git a/lib/tls/mbedtls/ssl.c b/lib/tls/mbedtls/ssl.c
index 9ba5625..844ee4b 100644
--- a/lib/tls/mbedtls/ssl.c
+++ b/lib/tls/mbedtls/ssl.c
@@ -426,7 +426,6 @@ lws_tls_peer_cert_info(struct lws *wsi, enum lws_tls_cert_info type,
                buf->exists = !!x509;
                break;
        case LWS_TLS_CERT_INFO_VERIFIED:
-               rc = SSL_get_verify_result(wsi->ssl);
                buf->verified = SSL_get_verify_result(wsi->ssl) == X509_V_OK ? 1 : 0;
                break;
        default:
```

lws-team:
Thanks... I modified it to remove the test if the cert exists... if the cert doesn't exist, `lws_tls_peer_cert_info()` will always return -1.  It returns -1 for other things but when testing for peer cert validity, it only returns -1 for x509 == NULL.

pparadzik:
It seems that function `ssl_get_peer_mbedtls_x509_crt` doesn't return client certificate even though it is added and the same thing works with openssl. First problem is that the `SSL_VERIFY_PEER` is not set, and second is related to mbedTLS wrapper not calling `mbedtls_ssl_conf_ca_chain` because `ssl->client_CA` is not set. When `SSL_new` is called, it inherits data, including `client_CA`, from the `ssl_ctx`, but only `ssl_client_ctx` has client certificate set in function `lws_tls_client_create_vhost_context`. So, this is what made it work
```
diff --git a/lib/tls/mbedtls/client.c b/lib/tls/mbedtls/client.c
index 5e971e04..fd51c9da 100644
--- a/lib/tls/mbedtls/client.c
+++ b/lib/tls/mbedtls/client.c
@@ -172,7 +172,7 @@ lws_tls_client_create_vhost_context(struct lws_vhost *vh,
                return 1;
        }
 
-       SSL_CTX_add_client_CA(vh->ssl_client_ctx, vh->x509_client_CA);
+       SSL_CTX_add_client_CA(vh->ssl_ctx, vh->x509_client_CA);
 
        lwsl_notice("client loaded CA for verification %s\n", ca_filepath);
 
diff --git a/lib/tls/mbedtls/server.c b/lib/tls/mbedtls/server.c
index 9d43c0fb..aad7698b 100644
--- a/lib/tls/mbedtls/server.c
+++ b/lib/tls/mbedtls/server.c
@@ -25,6 +25,8 @@ int
 lws_tls_server_client_cert_verify_config(struct lws_context_creation_info *info,
                                         struct lws_vhost *vh)
 {
+       SSL_CTX_set_verify(vh->ssl_ctx, SSL_VERIFY_PEER, NULL);
+
        return 0;
 }
```
but using `ssl_ctx` instead of `ssl_client_ctx` seems wrong here

lws-team:
I added it as its own patch and pushed it on master... it should be ssl_client_ctx but if this works around it, let's try it and see if any problems come.

pparadzik:
Latest lws does not return client certificate for me because `mbedtls_ssl_get_peer_cert` returns `NULL` in `ssl_pm_handshake`. It seems that this two commits are the cause
```
commit edadb1a8cdaf6f161d31d1f64a9f51256a413629 (HEAD)
Author: Andy Green <andy@warmcat.com>
Date:   Thu Nov 2 06:22:44 2017 +0800

    LWS_SERVER_OPTION_IGNORE_MISSING_CERT

commit 72e158ac680c1db6cecdc2d3be09b6a07beebe68
Author: Andy Green <andy@warmcat.com>
Date:   Tue Oct 31 18:50:59 2017 +0800

    openssl: add csr generation

```




When building with openssl, everything works fine.

lws-team:
I added helper scripts and associated baggage in `./scripts/client-ca/`, and documentation on using it in `./READMEs/README.test-apps.md`.

I added a call on both openssl and mbedtls versions of lws_tls_server_accept() to dump the CN of the peer cert, like this

```
lwsts[4241]: lws_tls_server_accept: client cert CN 'localhost'
```

Using client + server certs + the ca generated from there, client certs work using OpenSSL or mbedtls on FFOX57 + Chrome on current master.

I'm not 100% sure if whatever it was got fixed, or the problem is something else, but at least this gives us a way to check the same thing wrt client certs now.

pparadzik:
This is what fixed the issue for me
```
diff --git a/lib/tls/mbedtls/wrapper/platform/ssl_pm.c b/lib/tls/mbedtls/wrapper/platform/ssl_pm.c
index 685b0255..4d98b960 100755
--- a/lib/tls/mbedtls/wrapper/platform/ssl_pm.c
+++ b/lib/tls/mbedtls/wrapper/platform/ssl_pm.c
@@ -798,7 +798,7 @@ void SSL_set_SSL_CTX(SSL *ssl, SSL_CTX *ctx)
        /* apply new ctx cert to ssl */
 
        mbedtls_ssl_set_hs_own_cert(&ssl_pm->ssl, x509_pm->x509_crt, pkey_pm->pkey);
-       mbedtls_ssl_set_hs_authmode(&ssl_pm->ssl, MBEDTLS_SSL_VERIFY_NONE);
+       mbedtls_ssl_set_hs_authmode(&ssl_pm->ssl, MBEDTLS_SSL_VERIFY_OPTIONAL);
        mbedtls_ssl_set_hs_ca_chain(&ssl_pm->ssl, x509_pm->x509_crt, NULL);
 
 }
```
Without it, I can't get the client certificate. Maybe `ssl->verify_mode` should be checked before, like in `ssl_pm_reload_crt` function.

lws-team:
Rather than just doing that I had started going a bit deeper into it yesterday, I pushed those changes that relate to SNI on mbedtls as well as this, please try current master.

By the way I also got ACME working on mbedtls today ;-) it's also on current master.

pparadzik:
That's great news. Regarding the latest version, I'm having issues with building
```
lib/libwebsockets.so.12: undefined reference to `lws_jws_base64_enc'
```

lws-team:
OK... if you have LWS_WITH_MBEDTLS but not LWS_WITH_JWS it will break.  I pushed what should be a fix.

None of the CI things have mbedtls available directly....

pparadzik:
Now it works, but I'm having another problem with building
```
lib/tls/mbedtls/server.c:510:1: error: ‘_rngf’ defined but not used [-Werror=unused-function]
 _rngf(void *context, unsigned char *buf, size_t len)
```
This is what made it go away
```
diff --git a/lib/tls/mbedtls/server.c b/lib/tls/mbedtls/server.c
index 6ab8ba26..f77822cf 100644
--- a/lib/tls/mbedtls/server.c
+++ b/lib/tls/mbedtls/server.c
@@ -506,6 +506,8 @@ lws_tls_acme_sni_cert_destroy(struct lws_vhost *vhost)
 {
 }
 
+#if defined(LWS_WITH_JWS)
+
 static int
 _rngf(void *context, unsigned char *buf, size_t len)
 {
@@ -515,6 +517,8 @@ _rngf(void *context, unsigned char *buf, size_t len)
        return -1;
 }
 
+#endif
+
 #if defined(LWS_WITH_JWS)
 /*
  * CSR is output formatted as b64url(DER)
```

lws-team:
Thanks... I just moved the existing `#if defined` back up to cover the helper function as well.

pparadzik:
I also have memory leak in the latest version
```
==11276== 1,793 bytes in 1 blocks are definitely lost in loss record 1 of 3
==11276==    at 0x4C2CD8F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==11276==    by 0x4C2F195: realloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==11276==    by 0x4E68038: _realloc (alloc.c:64)
==11276==    by 0x4E68080: lws_realloc (alloc.c:77)
==11276==    by 0x4E76E85: alloc_file (tls.c:162)
==11276==    by 0x4E76F5E: lws_tls_alloc_pem_to_der_file (tls.c:196)
==11276==    by 0x4E7A683: lws_tls_server_vhost_backend_init (server.c:221)
==11276==    by 0x4E798C5: lws_context_init_server_ssl (ssl-server.c:86)
==11276==    by 0x4E65F14: lws_create_vhost (context.c:838)
```

lws-team:
Thankes, that should be fixed.  Unfortunately mbedtls is so slow under valgrind, it times out communicating with the ACME server.

lws-team:
Please see if the current master helps.

Under valgrind, with 4096 bit keys as used with the ACME stuff, it can take tens of seconds to do whatever it's doing.  It basically meas using valgrind affects its normal operation for timeouts etc.

pparadzik:
That fixed it.

Issue #956, opened 2017-07-05
[esp32] Does not find libwebsockets.h when the path gets too long


johsely1234:
I can compile my esp32 project without problems. 
But if I change the path to a longer path (I just copied it there) and rebuild it, it won't compile anymore => fatal error: libwebsockets.h: No such file or directory

I copied it one folder up, and it worked again.

What could be the problem here?

lws-team:
So what was it trying to compile?

How long is 'longer path' exactly?

Did you copy ./build, or regenerate a clean empty dir there in the new destination?

Afaik nothing in the build system should be especially fragile to path length.

johsely1234:
I have a modified version of the esp32-factory app. (but only modified the sources not the build system)

/home/esp32dev/esp/orcaSVN/ABC/ABCGenius/esp32/trunc/lws_server => not working
/home/esp32dev/esp/orcaSVN/ABC/ABCGenius/esp32/lws_server => not working
/home/esp32dev/esp/orcaSVN/ABC/ABCGenius/lws_server => working

I made a make clean and rm -rf build/ every time after copying.

Probably its also on espressif side, but I don't know... the examples also have a very long path

Sorry for responding so late, a lot of other things to do...

lws-team:
I could reproduce it... it should be fixed (at least for 6 or so more directory levels deep)

johsely1234:
Yes it works now, thanks!

PR #1089, opened 2017-11-12
Esp32 on windows

benjamind:
This revises the ESP32 build to work reliably under Windows (MSYS2) shell.

Have also moved the check for `genromfs` and `xxd` from the `CMakeLists.txt` to the `esp32.mk`. This way the library can be used without the dependency on these tools, and only when its actually being used to pack the binaries is it required. This will allow users to opt out of the `genromfs` packing in their standalone implementations if they want. I will add instructions to the `lws-esp32-factory` and `lws-esp32-test-server-demos` projects detailing how to install these dependencies under Windows.

Also added some minor documentation cleanup, exposing the ESP32 and ESP8266 readmes in the doxygen'd output, adding `doc` to `.gitignore` and standardizing use of whitespace in the `README.build.md`.


lws-team:
Thanks, I added these, but I moved the patches around a bit to better summarize what they are doing.

Aso on the "fixes esp32 build", the weirdo paths `../../../..` etc are there for a reason, the depth of the dir the project itself and esp-idf are in is actually unknown.  We already had a bug where it was too far down

https://github.com/warmcat/libwebsockets/issues/956

so just trimming this to two levels won't fly.  So I removed that part of the patch and confirmed it still works here at least.  Please check the meddling around didn't break anything for you.

benjamind:
Hmm, I see.

I will investigate the path issue a little more. I think there is probably
an environment variable we can pass down to avoid this crazy path hack.
Does not seem right to be abusing the relative paths like this.

lws-team:
If you want to provide a better fix, it's welcome.  Whether "right" or not, it works, which is better than the alternative :-)

benjamind:
Agreed a working solution is better than a non-working solution ;-)

However, if you put something like this in, it will keep coming up as an issue when future maintainers see that huge collection of seemingly incorrect parent folder paths.

I spent a while looking through the build system and I think the cleanest way of implementing this is to add a `Makefile.projbuild` to the `libwebsockets` component. This file includes the single line:
```
CPPFLAGS += -I$(BUILD_DIR_BASE)/libwebsockets/include
```
This will then add the include folder to the project global `CPPFLAGS` so the headers will be properly found across the project.

Short of re-arranging the build so that the include files get copied back into the components folder after cmake completes (and the associated clean targets are added) I think this is better than the fragile path we have now.

benjamind:
FYI, without the above fix regarding the include path the current master of `libwebsockets` does not compile on windows. The issue is that the resolved relative path ends up containing a double slashed path that does not resolve properly, e.g. `'../../../../../../../..//d/my/foo/dir/'`

lws-team:
Yeah... that works here too thanks.  There's a lot of esp32-specific junk appearing at the top level of lws to keep esp build system happy.  No other platform requires that.

Anyway I added it and I will push it later thismorning along with a WIP of the -factory changes so we are testing the same thing.

PS '../../../../../../../..//d/my/foo/dir/' good old windows

benjamind:
Agreed on the esp32-specific junk. I think this could be resolved by changing the way the `libwebsockets` repository is included into `esp-idf` projects.

For other libraries I have made a shim component which lives in the `/components` folder that then reached out to a `../vendor/my-foo-library` to do compilation of the vendor library. This works pretty well, and you can then keep all the `esp-idf` specific scripts, makefiles and so on out of the core library.

PR #1091, opened 2017-11-12
Removes fragile relative path include dir

benjamind:


lws-team:
Thanks... I folded this into your earlier patch (I read your description on the other issue and added it by hand before seeing this).  Again I'll push the updated lws master shortly.

PR #1092, opened 2017-11-12
Uses PROJECT_NAME instead of DIRNAME for binary filename

benjamind:
Discovered this while integrating into my app.

My project repo structure is like this:
```
project_name
    firmware
        components
            libwebsockets
        Makefile
    other_things
```

And my Makefile defines my `PROJECT_NAME` as `project_name`. The current `esp32.mk` assumes its named after the parent folder which in this case is `firmware`.

This PR changes this to just use the `PROJECT_NAME` variable directly.


lws-team:
Seems good thanks, it's added I will push it shortly.

PR #1102, opened 2017-11-20
Add a flag to disable CMake exports

raaaar:
Add LWS_EXPORT_LWSTARGETS (defaults to ON) that can be used to disable CMake export targets as well as install export for libwebsockets.


lws-team:
Thanks, I modified it a little (to LWS_WITH_...) and pushed it on master.

PR #1106, opened 2017-11-22
Support for X-Auth-Token HTTP header

mmilata:
AFAIK the header is non-standard but there is quite a bunch of programs using it. Would you please consider adding it?

cc @tchmelar


lws-team:
Thanks, it's pushed on master.  I have also seen that header around.

PR #1109, opened 2017-11-25
gencert: Fixes #1108 - skip generation when no test server is built

### LINKED ISSUE #1108: Building with LWS_WITHOUT_TESTAPPS=ON should not generate SSH test keys
I'm building with LWS_WITHOUT_TESTAPPS=ON and the configuration step generates the keys and the install step [unconditionally wants to install them](https://github.com/warmcat/libwebsockets/blob/v2.4-stable/CMakeLists.txt#L1862).

GENCERT is conditionally protected for LWS_WITHOUT_TEST_SERVER but not for LWS_WITHOUT_TESTAPPS (which when defined also disables the test server).

I believe extending this check:
```cmake
if (LWS_WITH_SSL AND OPENSSL_EXECUTABLE AND NOT LWS_WITHOUT_TEST_SERVER)
	set(GENCERTS 1)
```

with `AND NOT LWS_WITHOUT_TESTAPPS` will be enough to solve this issue and will proceed to test following up with a PR if I'm not wrong on the solution. I also don't see a reason to generate the keys when `LWS_WITHOUT_SERVER` doesn't make much sense as that stops the test server from building.

mulander:
apart from LWS_WITHOUT_TEST_SERVER the test server will also not be
built if LWS_WITHOUT_TESTAPPS or LWS_WITHOUT_SERVER is defined,
we update the CMake configuration to skip cert generation
in those cases.

Signed-off-by: Adam Wolk <adam.wolk@koparo.com>


lws-team:
Thanks... I squashed the two commits into one patch and pushed it on master.

PR #1111, opened 2017-11-27
lib: Spelling fixes based on debian packaging

shramov:
Some minor spelling fixes that are not yet merged into 2.4 branch


lws-team:
Thanks, it's pushed on v2.4-stable and master.

PR #1122, opened 2017-12-09
CMakelists: remove build user/hostname

lynxis:
The build user and hostname breaks reproducible builds.
Reproducible builds is an effort to create bit-by-bit identical
binaries independent from certain part including the build system.


lynxis:
For further information take a look at the reproducible builds webpage https://reproducible-builds.org/

lws-team:
I understand the problem, but instead of 'remove' how about being able to control it from a cmake option?

lynxis:
I've replaced the commit. Why do you want to include the build username/hostname? Isn't the git hash already unique?

lws-team:

>  Why do you include the build username/hostname?

There is reproducibility, and there is traceability.  For example if you buy some food in the store, for a tin of pringles or whatever it should always taste the same, and the tin will always look the same; they can reproduce it.  But on the tin will be printed the exact date and time of manufacture and information such as which factory and even production line did it... if there is a problem they can trace back where it came from.  It's traceability.

For a distro where you build everything at once you might have other ways like CI to find out where and when it was build, and it's a priority to guarantee there were no changes from source inserted in there.  For other cases like embedded where the rootfs and updates are composed from separate builders, traceability is also valuable.

lynxis:
Thanks!

Issue #1129, opened 2017-12-20
Dead socket not detected on iOS


sergiohs84:
When apps go into background in iOS, they are eventually "paused" and don't receive any CPU time. We have observed that, when app is moved back into foreground and the server has already closed the socket, `libwebsockets` does not provide us with the usual callback indicating `LWS_CALLBACK_CLOSED` nor `LWS_CALLBACK_CLIENT_CONNECTION_ERROR`.

With logs enabled we can see the library seems to be stuck on a loop consuming 100% of CPU time. It tries to `SSL_read()` with error but, instead of returning `LWS_SSL_CAPABLE_ERROR`, the code returns `LWS_SSL_CAPABLE_MORE_SERVICE`, resulting in a future attempt to read and finally the loop. However, as you can see below, the socket is closed and `errno` is 57 (`ENOTCONN`).
```
[2017/12/19 17:11:07:7140] DEBUG: lws_uv_timeout_cb
[2017/12/19 17:11:07:7140] DEBUG: 0x1075b79e0: SSL_read says -1
[2017/12/19 17:11:07:7140] DEBUG: lws_ssl_get_error: 0x1076e4708 -1 -> 2
[2017/12/19 17:11:07:7489] DEBUG: 0x1075b79e0: ssl err 2 errno 57
[2017/12/19 17:11:07:7489] DEBUG: SSL_want_read returns 1
[2017/12/19 17:11:07:7489] DEBUG: lws_ssl_capable_read: WANT_READ
[2017/12/19 17:11:07:7490] DEBUG: 0x1075b79e0: LWS_SSL_CAPABLE_MORE_SERVICE
[2017/12/19 17:11:07:7490] DEBUG: 0x1076e2880: SSL_read says -1
[2017/12/19 17:11:07:7490] DEBUG: lws_ssl_get_error: 0x1075b6448 -1 -> 2
[2017/12/19 17:11:07:7490] DEBUG: 0x1076e2880: ssl err 2 errno 57
[2017/12/19 17:11:07:7490] DEBUG: SSL_want_read returns 1
[2017/12/19 17:11:07:7490] DEBUG: lws_ssl_capable_read: WANT_READ
[2017/12/19 17:11:07:7490] DEBUG: 0x1076e2880: LWS_SSL_CAPABLE_MORE_SERVICE
(...)
```
[...]

PR #1130, opened 2017-12-21
Detect dead socket on iOS when server closed connection when client is in background

sergiohs84:
It aims to solve the ticket issue #1129
(https://github.com/warmcat/libwebsockets/issues/1129).


lws-team:
Thanks, I pushed it on v2.4-stable and the openssl backend of master.

sergiohs84:
Thank you for the quick response.
In case we detect any side-effect in the coming days, I'll notify you

PR #1145, opened 2018-01-07
fix comparison between unsigned and signed integer

gmondada:
On Windows, sockets are unsigned int, and testing if socket >= 0 or socket < 0 makes no sense.
https://msdn.microsoft.com/en-us/library/windows/desktop/ms740516(v=vs.85).aspx


lws-team:

PR #1155, opened 2018-01-17
Fixed memory leak

pchero:
- Added event free function for accept.


lws-team:
Thanks... I squashed both patches and modified it a little bit... the name to `lws_libevent_destroy()` and moved the existing code to destroy the event pipe and the added call to its own function, `lws_destroy_event_pipe()`.  It's pushed on master.

PR #1185, opened 2018-02-22
Do not pass AI_CANONNAME to getaddrinfo()

mmilata:
The reason for removing the flag is that, at with uclibc, it causes reverse DNS lookup even when numeric IP address is passed to `getaddrinfo()`. This causes unnecessary one minute delay when DNS server is configured but unreachable.

In theory this should not break anything since the `ai_canonname` member does not seem to be accessed anywhere in libwebsockets codebase. According `man getaddrinfo` that should be the only difference of using the flag:
> If the AI_CANONNAME bit is set, a successful call to getaddrinfo() will return a NUL-terminated string containing the canonical name of the specified host name in the ai_canonname element of the first addrinfo structure returned.

What do you think?


lws-team:
Thanks... it sounds reasonable to me... I pushed it on master.

Issue #1200, opened 2018-03-07
Make response status available when WebSocket connection fails


joelittlejohn:
Hi. Firstly, thanks for providing this amazing library!

We're using libwebsockets to connect to an API that requires authorization, so to establish a WebSocket connection you have to provide a bearer token. If you don't provide a token (or the token is invalid/expired), the server will return a 401 response status and not upgrade the connection.

We're finding that when our authorization is invalid, we simply get `HS: ACCEPT missing` from libwebsockets rather than any information about _why_ the socket could not be established. This is a problem because if the reason for the failure is 401, then the client _must_ refresh their bearer token before retrying.

There seems to be two issues:

1. libwebsockets looks for `Sec-WebSocket-Accept` in the response first, before it checks the response status. This seems wrong, because you need to check whether the server has actually agreed to accept the connection (by checking for status code 101) before you check for the accept header.

    My suggestion here is to move this check:

https://github.com/warmcat/libwebsockets/blob/3dc71cf05998de2d47728c920e4744704552ae0c/lib/client/client.c#L766

so that it is above this check:

https://github.com/warmcat/libwebsockets/blob/3dc71cf05998de2d47728c920e4744704552ae0c/lib/client/client.c#L760

2. If the status code is not 101, it's important to provide some way for the caller find out what the status code is. If it's a 401, then the authorization is not correct, if it's  a 429, then the client should wait before retrying, if it's a 404 then the client should fix their endpoint address, etc. 

    Either some callback should be available to handle the status code, or this message should include the actual received status code:
https://github.com/warmcat/libwebsockets/blob/3dc71cf05998de2d47728c920e4744704552ae0c/lib/client/client.c#L769

What do you think?

The important case for us is 401, because the client must refresh their token when this happens. If we had an explicit failure `HS: ` message for this then that would be adequate. Obviously it's still important that the status code is validated before `Sec-WebSocket-Accept` is checked.

lws-team:
It sounds pretty reasonable to me.  Can you provide a patch?

xajdlg:
[...]

PR #1201, opened 2018-03-07
Report unauthorized error when 401 is received, instead of "HS: ACCEPT missing"

joelittlejohn:
See #1200 

The two changes included here are as discussed:

1. Check the response status before checking for the Sec-WebSocket-Accept header 

2. Return `HS: ws upgrade unauthorized` as the failure message when 401 is received.

@lws-team I opted to add a specific message here to avoid issues with having to format a new string to contain the response status, or finding an appropriate place to put it that would be freed (I couldn't find one). I considered creating a new fixed length char array to hold cce writing a formatting string to it (e.g. snprintf), but in the end just opted to keep this simple. Let me know your thoughts and I can revisit if necessary.

### DIFF
--- a/lib/client/client.c
+++ b/lib/client/client.c
@@ -757,6 +757,13 @@ lws_client_interpret_server_handshake(struct lws *wsi)
 		return 0;
 	}
 
+	if (p && strncmp(p, "401", 3)) {
+		lwsl_warn(
+		       "lws_client_handshake: got bad HTTP response '%s'\n", p);
+		cce = "HS: ws upgrade unauthorized";
+		goto bail3;
+	}
+
 	if (p && strncmp(p, "101", 3)) {
 		lwsl_warn(
 		       "lws_client_handshake: got bad HTTP response '%s'\n", p);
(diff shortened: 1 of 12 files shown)


lws-team:
Nice job, thanks, it's pushed on master + v2.4-stable.

joelittlejohn:
Thanks Andy, amazing response! :slightly_smiling_face:

joelittlejohn:
Of course you know what my next question is going to be (sorry!)... any idea when you'll cut 2.4.2?

lws-team:
It's tagged out... enjoy

joelittlejohn:
Ahh, sorry @lws-team there is a bug in my patch which is causing valid connections to fail!

This line:
```
if (p && strncmp(p, "401", 3)) {
```
in client.c should instead be:
```
if (p && strncmp(p, "401", 3) == 0) {
```

:man_facepalming:

lws-team:
Well, nothing that can't be fixed with a ninja retag.

joelittlejohn:
Andy, we're seeing that Homebrew has the old 2.4.2 and the old sha. Would it be possible to cut a new version?

PR #1206, opened 2018-03-13
fixed dead poll

VollmondT:
Removed unnecessary and dangerous test for zero read bytes.
I guess, there is no reason for `poll` to trigger a file descriptor in this case except `EOF`. Anyway, the server is hanging up for `timeout_secs` in the `poll` loop on instant client disconnect:
`nc --send-only localhost 8888 < /dev/null`

Tested with:
`ctx.options = LWS_SERVER_OPTION_VALIDATE_UTF8 | LWS_SERVER_OPTION_DO_SSL_GLOBAL_INIT | LWS_SERVER_OPTION_ALLOW_NON_SSL_ON_SSL_PORT`

### DIFF
--- a/lib/ssl.c
+++ b/lib/ssl.c
@@ -818,12 +818,6 @@ lws_server_socket_service_ssl(struct lws *wsi, lws_sockfd_type accept_fd)
 					wsi->redirect_to_https = 1;
 				goto accepted;
 			}
-			if (!n) /*
-				 * connection is gone, or nothing to read
-				 * if it's gone, we will timeout on
-				 * PENDING_TIMEOUT_SSL_ACCEPT
-				 */
-				break;
 			if (n < 0 && (LWS_ERRNO == LWS_EAGAIN ||
 				      LWS_ERRNO == LWS_EWOULDBLOCK)) {
 				/*


lws-team:
> Removed unnecessary and dangerous test for zero read bytes.

Hmmm I'm sure this reflects some real problem, but you are too quick to judge it.

It is in tls code.  Bear in mind poll() reports the situation on the socket that deals with the encrypted link, but when we read, we read decrypted data.  It's quite possible for traffic to appear on the socket that is swallowed by the tls library without producing output.

If the tls library uses recv(), 0 return is ambiguous

```
Return Value

These calls return the number of bytes received, or -1 if an error occurred. The return value will be 0 when the peer has performed an orderly shutdown.
```

https://linux.die.net/man/2/recv

At any rate different platforms act quite differently with poll().  This code has changed a lot on master.  So what is your platform, which tls lib are you using, and if I want to reproduce it, how do I do that?  Just disconnect the client and see 100% cpu for 5s?

VollmondT:
OpenSSL-1.0.2n is my TLS library, Linux platform.
>  Just disconnect the client and see 100% cpu for 5s?

Exactly!
By the way, current master is affected too.

lws-team:
On current master:

 - just build with cmake .. && make && sudo make install

 - run `libwebsockets-test-server -s`

 - visit `https://localhost:7681/` in Firefox

 - close the tab

 - watch top... no sign of anything strange

Can you give me the exact steps you are using to reproduce it, if those aren't it?

lws-team:
The OpenSSL on Fedora 27 is 1.1.0g FWIW... I can try it with your version I guess

lws-team:
I built exactly tag "OpenSSL_1_0_2n" from openssl git and forced lws to use it

`cmake .. -DLWS_WITH_MBEDTLS=0 -DLWS_OPENSSL_INCLUDE_DIRS=/usr/local/ssl/include -DLWS_OPENSSL_LIBRARIES="/usr/local/ssl/lib/libssl.so;/usr/local/ssl/lib/libcrypto.so" -DLWS_LINK_TESTAPPS_DYNAMIC=1`

It installs things in `/usr/local/ssl/...`, so I told it where to go by hand since it's not in ld.so.conf

```
$ LD_LIBRARY_PATH=/usr/local/ssl/lib libwebsockets-test-server -s
```

There's no sign of any cpu spinning looking with `top -d1` and closing the tab as described before.

VollmondT:
I've catched  this issue in my custom server, sorry for my poor explanation. I see that original server is working, but you need exactly:
`ctx.options = LWS_SERVER_OPTION_VALIDATE_UTF8 | LWS_SERVER_OPTION_DO_SSL_GLOBAL_INIT | LWS_SERVER_OPTION_ALLOW_NON_SSL_ON_SSL_PORT`
Just remove  `info.options |= LWS_SERVER_OPTION_REDIRECT_HTTP_TO_HTTPS;` in test-server.c (v2.4-stable)

Second one and very important thing: you need to test it with zero length TCP stream. `nc --send-only localhost 8888 < /dev/null` will open connection and close immediately.

VollmondT:
Update:
`libwebsockets-test-server -s -a`

lws-team:
Hum okay... so I do this and rebuild

```
diff --git a/test-apps/test-server.c b/test-apps/test-server.c
index d91e0b24..243c9689 100644
--- a/test-apps/test-server.c
+++ b/test-apps/test-server.c
@@ -441,9 +441,9 @@ int main(int argc, char **argv)
        info.ip_limit_ah = 24; /* for testing */
        info.ip_limit_wsi = 105; /* for testing */
 
-       if (use_ssl)
+//     if (use_ssl)
                /* redirect guys coming on http */
-               info.options |= LWS_SERVER_OPTION_REDIRECT_HTTP_TO_HTTPS;
+//             info.options |= LWS_SERVER_OPTION_REDIRECT_HTTP_TO_HTTPS;
 
        if (context == NULL) {
```
and hit it with `nc --send-only localhost 7681 < /dev/null` ... it still doesn't trigger anything according to top.  If I tell it the wrong port, it gives a connection refused, so I think it's trying to do its thing.

Ohhh with the extra -a, I can see the problem though.... basically that is LWS_SERVER_OPTION_ALLOW_NON_SSL_ON_SSL_PORT ... let me see if I can understand what happens then.  But this is a non-standard and non-recommended config already...

lws-team:
I pushed my take on what I think the fix should look like on master https://github.com/warmcat/libwebsockets/commit/1fd7196b393ca144932edbc6520f2636bd0a5a97 and v2.4-stable https://github.com/warmcat/libwebsockets/commit/f97b67a774b3b52a55ee5e05ef32d8058d0f0cda, please give it a try.

lws-team:
AFAICS this fixes it.  I appreciate your patch and help reproducing it.

VollmondT:
PEEKed 0
Thank you and sorry for big delay. It works!

PR #1215, opened 2018-03-24
Use mbedTLS OPTIONAL verification.

### LINKED ISSUE #1214: SSL verification skipped with mbedTLS and no CA provided
As a follow-up from the discussion in #1128 .

**The problem**

When using the mbedTLS backend and no CA is provided any certificate is accepted by the client as specified here:
https://github.com/warmcat/libwebsockets/blob/d2bc2d17043ff373c8f2c54cd3eff473078aa16f/lib/tls/mbedtls/mbedtls-client.c#L75-L79

**Expected behaviour**

All SSL connections should fail when in strict mode (`LCCSCF_USE_SSL`).

**My suggestion**

Use `SSL_set_verify(wsi->ssl, SSL_VERIFY_PEER, OpenSSL_client_verify_callback);` in any case and fail when no CA is provided.

or

Always use `MBEDTLS_SSL_VERIFY_OPTIONAL` internally, and then implement the desired behaviour in the `ssl_pm_get_verify_result` function (this should also allow to implement the less strict options there like self signed certs, hostname verification, etc).
https://github.com/warmcat/libwebsockets/blob/d2bc2d17043ff373c8f2c54cd3eff473078aa16f/lib/tls/mbedtls/wrapper/platform/ssl_pm.c#L762

If can make a PR if you are interested.
**Thank you for your amazing work on this library.**
As I'm here, I'd like to let you know that we are integrating `libwebsockets` as a third party library in Godot Engine (see https://github.com/godotengine/godot and  https://github.com/godotengine/godot/blob/master/thirdparty/README.md#libwebsockets )

Faless:
This allows lws to distinguish between untrusted CAs, hostname
mismatches, expired certificates.

**NOTE:** LCCSCF_ALLOW_SELFSIGNED actually allows for untrusted CAs, and
will also skip hostname verification. This is somewhat a limitiation of
the current lws verification process, but I didn't want to change too much and risk breaking compatibility with other libraries. It think it's kind of pointless to do further checks if you accept self-signed, as anyone can self sign a certificate (note, a self-signed certificate added in the CA chain WILL be accepted even in strict mode)

Closes #1214

### DIFF
--- a/lib/tls/mbedtls/mbedtls-client.c
+++ b/lib/tls/mbedtls/mbedtls-client.c
@@ -72,12 +72,8 @@ lws_ssl_client_bio_create(struct lws *wsi)
 	 * use server name indication (SNI), if supported,
 	 * when establishing connection
 	 */
-	if (wsi->vhost->x509_client_CA)
-		SSL_set_verify(wsi->ssl, SSL_VERIFY_PEER,
-			       OpenSSL_client_verify_callback);
-	else
-		SSL_set_verify(wsi->ssl, SSL_VERIFY_NONE,
-			       OpenSSL_client_verify_callback);
+	SSL_set_verify(wsi->ssl, SSL_VERIFY_PEER,
+		       OpenSSL_client_verify_callback);
 
 	SSL_set_fd(wsi->ssl, wsi->desc.sockfd);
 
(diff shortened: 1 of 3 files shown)


lws-team:
Thanks... it's pushed on master with some small changes and additions for docs and the test + minimal apps.  One addition is the diagnostic at the tls layer is passed back up to the CLIENT_CONNECTION_ERROR human readable arg... in the case it's down to no trusted CA on mbedTLS, that says it fairly clearly now

```
[2018/03/25 07:37:56:0725] NOTICE: http: connecting
[2018/03/25 07:37:56:9678] ERR: CLIENT_CONNECTION_ERROR: dumb: server's cert didn't look good, X509_V_ERR = 24: CA is not trusted
```

For the minimal apps I provide the CA certs for warmcat.com / libwebsockets.org they use and apply them conditional on LWS_WITH_MBEDTLS.  So hopefully this and the docs will make it clear enough.

PR #1231, opened 2018-04-10
Fixed authorization condition in 101 server response

Hasan-Jawaheri:
It seems like the condition is inverted...


lws-team:
Thanks... it's pushed on master... I broke it during the current refactor... you found it in a few hours :-)

PR #1248, opened 2018-04-20
Removing rudimental n1 variable

LabunskyA:
There is no real need in n1, since it used in only at one place once. 
Also it will prevent "variable ‘n1’ set but not used [-Werror=unused-but-set-variable]" error on build with -DLWS_MAX_SMP=N, where N > 1.


lws-team:
Thanks, it's pushed on master also adding a test to travis with LWS_MAX_SMP=32

PR #1260, opened 2018-04-30
Add MJS to lws_get_mimetype()

sigalrm:
ES6 Javascript modules may have the .mjs extension.


lws-team:
Thanks... there is a better way though, you can add a list of mimetype mappings to the mount definition.

I made a "minimal example" that shows how to do it.

https://github.com/warmcat/libwebsockets/tree/master/minimal-examples/http-server/minimal-http-server-mimetypes

PR #1265, opened 2018-05-04
Fix #1249: HTTP requests exhausting server ah pool

### LINKED ISSUE #1249: Successive HTTP requests in lws 2.4 exhausting server ah pool
We are facing an issue with lws 2.4 in unit tests that validate the HTTP client-server communication of our C++ wrapper library. Our tests create a lws client-server pair to make a succession of simple HTTP requests (no SSL or http/2) from the client to the server.

Until lws 2.3, the test cases execute extremely fast (<< 1 sec). But since precisely this commit introducing http2 support: https://github.com/warmcat/libwebsockets/commit/904a9c09205f0241dc7837225326d09009acf1b5 , execution is very slow (~1 min) and the observed behavior of the unit tests is something like the following:
> ----BEGIN TEST------
> GET /url
> GET /registry
> GET /ABC
> GET /
> GET //
> --- progression stops for a few seconds, looping around lws_service() for both client and server until the HTTP connection is established ---
> GET /abc/def/
> etc...
> ----END TEST------

This keeps happening with v2.4.1, v2.4.2 and v2.4-stable.

Enabling libwebsocket's log output gave a useful indication:
> lws_adopt_descriptor_vhost: waiting for ah

Indeed, it turns out that the *ah* pool of the lws server become exhausted after *max_http_header_pool* (default=5) previous HTTP requests have been made by the client. Execution continues after some internal timeout (?) frees the resources of previous connections.

More debugging allowed us to locate the exact change causing the connections to no longer be closed. It is the following 4 lines (2610-2613) in lib/server.c: https://github.com/warmcat/libwebsockets/commit/904a9c09205f0241dc7837225326d09009acf1b5#diff-9621bd8daa5d0a52d5af514af312fbaaR2610
Reverting back to the previous behavior of "goto fail;" here fixes the problem; Although we don't know enough about libwebsockets internals to suggest this as a proper fix.

As a workaround, we observed that forcing a return -1; from LWS_CALLBACK_HTTP_WRITEABLE in the server code also fixes the problem. However, bypassing the lws_http_transaction_completed() check as is done in test-server-http.c is unlikely to be the correct solution. We were otherwise unable to identify an appropriate change in the client code based on the existing examples and documentation. Even recreating the client context with 
lws_context_destroy() + lws_create_context() before each lws_client_connect_via_info() does not make a difference.

So the question is:
a) Could the change in behavior described above be caused by a regression in libwebsockets, or:
b) Is there anything specific to do in either the client or server code to properly close the HTTP connection(s) starting from v2.4?

rdumusc:


lws-team:
Nice job, thanks.  It's pushed on v2.4-stable.

PR #1273, opened 2018-05-13
Updates for esp32

jas39:
Not 100% sure on the lws-plat-esp32.c,, your code would work if you only have one OTA. Once you have more this seems more logical.

For protocol_esp32_lws_ota.c, i think i committed this some week ago


lib/plat/lws-plat-esp32.c
```diff
@@ -1631,8 +1631,8 @@ lws_esp_ota_get_boot_partition(void)
 	/* confirm what we are told is the boot part is sane */
 	spi_flash_read(part->address , &eih, sizeof(eih));
 	factory_part = esp_partition_find_first(ESP_PARTITION_TYPE_APP,
-			ESP_PARTITION_SUBTYPE_APP_FACTORY, NULL);
- 	ota = esp_partition_find_first(ESP_PARTITION_TYPE_APP,
+			ESP_PARTITION_SUBTYPE_APP_FACTORY, NULL);
```
jas39: Reverted to original code

Issue #1279, opened 2018-05-16
Is it possible to have all log levels in release mode?


qmatt115:
Greetings,
As I understood, if I compile LWS in release mode it only enables error, warn and notice. I am wondering if it is possible to have all log levels enable like debug mode does without changing the code? Since I have a project that the users like to control the log level by themselves. 

Thanks

lws-team:
Your understanding is basically correct... LLL_USER is also available in RELEASE... but it's like that for a reason, having the more verbose logs compiled in is expensive where the logs are in, eg, tight loops.

Similarly in DEBUG with gcc, it builds with -O0 to avoid confusing debugging efforts by compiler optimizations, and of course -g.

Your problem is you want to offer debug logs without the info from -g?  You can use `strip` if so.

qmatt115:
I see. Thanks for your timely reply. 

Since I looked into this issue, I found something odd related to the debug message. I compiled LWS with LWS_WITH_NO_LOGS and compiled my project in release mode. However, there is still a piece of debug message keep popping right after event "LWS_CALLBACK_UNLOCK_POLL" occurs. Here is my debug message:
`[2018/05/17 15:50:55:3796] DEBUG: 
[2018/05/17 15:50:55:3796] DEBUG: 0000: 47 45 54 20 2F 20 48 54 54 50 2F 31 2E 31 0D 0A    GET / HTTP/1.1..
[2018/05/17 15:50:55:3796] DEBUG: 0010: 48 6F 73 74 3A 20 6C 6F 63 61 6C 68 6F 73 74 2E    Host: localhost.
...`

If I set the log level to filter out debug, this message gone. 
Since the message is too general, I can't locate it in the code. I am not sure if it is a bug or a expected behavior?

Thanks again

lws-team:
lwsl_hexdump() is included unconditionally.  It doesn't use the same helpers like lwsl_debug() that get snipped out with NO_LOGS at build-time.

However there are almost no uses of it in lws... just turn it off in the loglevel mask.  Or if it's still bothering you, a patch to remove it is welcome.

PR #1285, opened 2018-05-22
#1279 changed the hexdump is only logged in debug mode instead of rel…

qmatt115:
#1279 changed the "lwsl_hexdump()s" are only logged in debug mode instead of release and debug mode. As your suggestion, here is the patch.


lws-team:
Thanks... I modified it to do the #if defined() inside lwsl_hexdump() itself and pushed on v2.4-stable.  Later I'll push on v3.0-stable and master.

PR #1294, opened 2018-05-27
Fixed a couple of runtime issues introduced in commit 44efcd6.

emptyVoid:
See #1291


lws-team:

PR #1298, opened 2018-06-01
fix memleak when role h1 failed to upgrade to websocket

whuji:
if LWS_ROLE_WS is defined, a wsi->ws object is allocated in rops_client_bind_h1 but never destroyed in rops_destroy_role_h1.


lws-team:

PR #1309, opened 2018-06-16
Fixes to track updates in esp-idf

negativekelvin:


lws-team:

I actually tried to update this three weeks or so ago but did not have time to track down their changes to where the fd offset applies, so this is very helpful.  I think I made other changes so I will pick those out from my and integrate them later if so.

PR #1316, opened 2018-06-24
Handle libuv >= 1.21.0 header change.

DomT4:
In 1.21.0, the latest release, `libuv` moved from `uv-version.h` to `uv/version.h`. This PR handles that change whilst not immediately killing off the old header check for people & packagers who aren't running the very latest `libuv` release.

This change can be written a whole bunch of different ways so if you'd prefer to do it differently please feel free to do so & reject this.


lws-team:
Thanks... you need to export LWS_HAVE_NEW_UV_VERSION_H in lws_config.h.in as well, otherwise libwebsockets.h isn't going to be able to see it.

With that change I pushed these on master + v3.0-stable.

DomT4:
Ah, thanks for catching & fixing that; CMake is not a massively strong point for me. Thank you for folding this in so quickly!

PR #1319, opened 2018-06-30
openssl: Fix compile with 1.1.0 - deprecated APIs

neheb:
These function have been deprecated. Keep compatibility with OpenSSL 1.0.2 as it is still supported.

### DIFF
--- a/lib/tls/openssl/ssl.c
+++ b/lib/tls/openssl/ssl.c
@@ -142,9 +142,13 @@ lws_context_init_ssl_library(const struct lws_context_creation_info *info)
 
 	lwsl_info("Doing SSL library init\n");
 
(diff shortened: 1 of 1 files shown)


lws-team:

lws-team:
At least in 1.1.0h, these compat defines already exist in `openssl/x509.h`

```
# if OPENSSL_API_COMPAT < 0x10100000L
#  define X509_get_notBefore X509_getm_notBefore
#  define X509_get_notAfter X509_getm_notAfter
#  define X509_set_notBefore X509_set1_notBefore
#  define X509_set_notAfter X509_set1_notAfter
#endif
```
I adapted the patch to only define them if additionally not already defined.

PR #1325, opened 2018-07-08
Fixes the binding error when SMP in tandem with a Unix domain socket

Zhang-Junzhi:
lws is compiled with -D LWS_UNIX_SOCK:BOOL=ON -D LWS_MAX_SMP=3 .

When using the following creation info,

static struct lws_context_creation_info info =
{
.iface = "/tmp/lwstest",
.options = LWS_SERVER_OPTION_UNIX_SOCK,
.protocols = ...,
.count_threads = 3,
};

The server gets errors:

[2018/07/07 12:49:07:2697] ERR: ERROR on binding fd 11 to "/tmp/lwstest" (-1 98)
[2018/07/07 12:49:07:2697] ERR: VH default: iface /tmp/lwstest port 12148 DOESN'T EXIST

This patch fixes the issue.

### DIFF
--- a/lib/roles/http/server/server.c
+++ b/lib/roles/http/server/server.c
@@ -131,6 +131,17 @@ done_list:
 
 	(void)n;
 #if defined(__linux__)
+#ifdef LWS_WITH_UNIX_SOCK
+	/*
+	 * A Unix domain sockets cannot be bound for several times, even if we set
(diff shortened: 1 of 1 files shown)


lws-team:
Thanks a lot... it's pushed on master + v3.0-stable.

PR #1326, opened 2018-07-09
 Fix I/O hang after received a large deflate frame

shinny-chengzhi:
When a large deflate frame been received, WSAEnumNetworkEvents will indicate the socket is ready to read. And because the frame is compressed, it may not be consumed entirely(not all bytes ready to receive have been received), since WSAEnumNetworkEvents is edge triggered, and the socket read buffer never been drained, WSAEnumNetworkEvents will never indicate the socket is ready to read again. What here need is level trigger behavior, thus add additional recv with empty buffer to reset edge status.

### DIFF
--- a/lib/plat/windows/windows-service.c
+++ b/lib/plat/windows/windows-service.c
@@ -166,8 +166,10 @@ _lws_plat_service_tsi(struct lws_context *context, int timeout_ms, int tsi)
 			if (pfd->revents & LWS_POLLHUP)
 				--eIdx;
 
-			if (pfd->revents)
+			if (pfd->revents) {
+				recv(pfd->fd, NULL, 0, 0);
 				lws_service_fd_tsi(context, pfd, tsi);
+			}
 		}
 	}
 


lws-team:
I don't 100% understand how the recv() makes a good outcome, but having looked at windows' completely broken WSA garbage, I probably wouldn't understand it even if I understood it.  So if this works better than what we have, I take your word for it.

It's pushed on v3.0-stable and master.

shinny-chengzhi:
The whole sequence of events:

T0: a large fragmented deflate frame arrived on socket(FIN flag not set).

WSAEnumNetworkEvents indicate the socket is ready to read.

lws_ssl_capable_read_no_ssl call recv(), all frame data been received, and FD_READ been re-enabled because of recv().

lws_ws+handshake_client been called, most of iterations inside while loop does nothing interesting because frame is not complete, only on the last iteration the frame will be complete and lws_ws_client_rx_sm will call deflate extension to decompress data and callback user code. Because the frame is large, deflate extension can't decompress in one go, and set rx_draining_ext to 1.

T1: an other fragmented deflate frame arrived on socket before next call to WSAEnumNetworkEvents
WSAEnumNetworkEvents indicate the socket is ready to read and call rops_handle_POLLIN_ws.

rops_handle_POLLIN_ws will call lws_ws_client_rx_sm and return because is rx_draining_ext 1, thus no recv() been called.

because FD_READ havn't been re-enabled, WSAEnumNetworkEvents will never indicate the socket is ready to read again.


A proper fix may need additional flag or change the sequence of function calls which is hard for someone as me who is not familiar with source code of libwebsocks, and may introduce new bug in the process.
Calling recv() is a quick&dirty fix since there are no other side effects except re-enabling FD_READ AFAIK.

PR #1330, opened 2018-07-10
change logic to store next iter prior to logic because logic can inva…

Babel-Fish:
@lws-team Here are the changes to allow pre storage of next iter. I can also just change `lws_start_foreach_ll` itself if preferred and update all references. I would just need to move "nxt" from end into the start macro. I am not sure which paths could result in the teardown so I just updated the one I encountered. Let me know your thoughts.


lws-team:
Thanks... I squashed the two patches into one and renamed it ...ll_safe.  It's pushed on master.

PR #1340, opened 2018-07-16
check timeout of poll()

dragos-oancea:
avoid setting timeout_ms to 0

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,7 @@ option(LWS_WITH_ACCESS_LOG "Support generating Apache-compatible access logs" OF
 option(LWS_WITH_RANGES "Support http ranges (RFC7233)" OFF)
 option(LWS_WITH_SERVER_STATUS "Support json + jscript server monitoring" OFF)
 option(LWS_WITH_ACME "Enable support for ACME automatic cert acquisition + maintenance (letsencrypt etc)" OFF)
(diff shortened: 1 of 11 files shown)


lws-team:
Can you help me understand why this is a good idea?

__lws_hrtimer_service() returns 0 if there is a scheduled event that should have happened in the past... it's not that strange if multiple events are scheduled around the same time we can't be everywhere at once, and we will be a little late for some.

poll() wait of 0 and then get to servicing the pending hrtimer event along with whatever else is happening is good, isn't it?

dragos-oancea:
Hi. Thank you for your comment. 

I've run into an issue with lws_service() and flow control, when lws_service() is just returning , not behaving like the docs say - it was supposed to return after the timeout:
"timeout_ms
    Timeout for poll; 0 means return immediately if nothing needed service otherwise block and service immediately, returning after the timeout if nothing needed service." 

However, this is a hack, just wanted your opinion. 
I run lws_service() on a timer now in my app , seems to work fine with flow control now.
Without the flow control (lws_rx_flow_contro()) lws_service() works as expected without the timer.

lws-team:
Okay... can you hack something up from a minimal example that shows the actual problem that I can reproduce?  I'm willing to imagine something's up somewhere.  But this particular patch will create different problems somewhere else.

PR #1355, opened 2018-07-30
Fix executing timers2

Babel-Fish:
Fix processing of timers. These can be processed by multiple threads with SMP. 

- Added locking around the identification of expired timers
- Changed lws_start_foreach_ll to the new lws_start_foreach_ll_safe since deletion takes place during processing
- Added new struct which contains all info needed to execute a timer
- Moved processing of the expired timers to a separate loop outside of locking

@lws-team


lws-team:
Thanks... I applied it with some changes (eg, OOM handling... the linked-list logic was also wrong), please give it a try.

PR #1369, opened 2018-08-15
Changed flash_ota and erase_ota to lws_flash_ota and lws_erase_ota

FredrikFornstad:
esp-idf has lately introduced their own version of erase_ota which to my understanding is not erasing the flash-area that matches the lws partition table. While it is possible to compile, warnings are given.

My suggestion is therefore to change the name and also do it for flash_ota for consistency. With this patch the commands to erase (or flash) the ota partition will now be:

make lws_erase_ota
make lws_flash_ota


lws-team:
Thanks, I pushed it on master and updated the esp32 branch to current master.

PR #1373, opened 2018-08-18
Fix timeout between writes on Windows

### LINKED ISSUE #1371: Loading "big picture" from test server takes about 30 seconds on Windows
Hi, love your library, it just does what I need and then gets out of the way. Thanks!

However, I noticed that on Windows loading the "big picture" from the test server takes about 30 seconds. Running the server on Linux it loads in about 0.5 seconds. This happens in both "master" and "v3.0-stable" branches.

The test server on Windows was compiled using the Visual Studio 2017 "Open Folder" feature that directly load CMakeLists.txt without generating project files first. LWS_WITH_SSL was set to OFF, everything else was left at default. Just to be sure, I tried both the x86-Release and the x64-Release builds.

This seems to be connected with LWS_CALLBACK_HTTP_WRITEABLE being called with some delay.

f69m:
These patches fix #1371 and some minor things on the way.

Most important: after servicing a write request, set timeout to zero, so WSAWaitForMultipleEvents will return immediately and not wait for timeout.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,4 @@ doc
 /build2/
 /build3/
 /cov-int/
+/.vs/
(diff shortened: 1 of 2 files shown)


lws-team:
Thanks a lot... these are pushed on master, and I rebased them for v3.0-stable and pushed them there as well.

PR #1389, opened 2018-08-29
Fix Iphlpapi.h -> iphlpapi.h

katzer:
If you cross compile for Windows you will get an error that the header cannot be found.

See here for code example from microsoft:
https://docs.microsoft.com/en-us/windows/desktop/iphlp/creating-a-basic-ip-helper-application


lws-team:
Thanks... it's pushed on master + v3.0-stable.

I guess on windows native, the problem wasn't coming due to case-insensitive filesystem hiding it.

PR #1407, opened 2018-09-14
Fix QTA erase after factory partition update

FredrikFornstad:
This PR together with a PR in lws-esp32-factory fix the problem that the OTA header cannot be erased after doing a factory partition update. 

Solves [https://github.com/warmcat/lws-esp32-factory/issues/35](https://github.com/warmcat/lws-esp32-factory/issues/35)


lws-team:
Thanks, it's pushed.  I also aligned -factory and -test-server-demos on lws with this.

PR #1416, opened 2018-09-28
fix rxflow, return err when conflicting rxflow buf

dragos-oancea:


lws-team:
Thanks, it's pushed on v2.4-stable... sorry for the delay.

PR #1427, opened 2018-10-09
-Task/sp 84

shalinichaturvedi-viewglass:
Set -Wno-deprecated-declarations in CFlags


lws-team:
Bro you need to update to current lws master.

Then add you patch(es).

Then push/

PR #1432, opened 2018-10-14
Fix cache-control when not using mounts.

ABruines:
This allows setting the cache-control header when calling lws_http_serve_file() from user code that does not use mounts.


lws-team:
Thanks, it's pushed.

This extra headers business is a holdover from the pre http/2 support era and will only work on h1.  So at some point it will have to be removed, or replaced by a callback that adds the headers in an http-version-agnostic way.  Until then, no point worrying about the case checking stuff when it's fixing something...

PR #1433, opened 2018-10-14
Fix cache-control when not using mounts. (v2)

ABruines:
My boolean logic is failing me today, sorry about that @lws-team .
This time it actually works, so I'm going to quit for today while I'm ahead ;)


lws-team:
Thanks, I squashed it into the other patch.

PR #1448, opened 2018-11-08
tls: Fix compile without deprecated OpenSSL APIs-

neheb:
These headers are included implicitly with others when deprecated APIs are enabled.


lws-team:
Thanks, it's pushed on master + v3.0-stable.

PR #1452, opened 2018-11-14
Fix iface is not preserved on redirect

Yimgo:
No idea if v2.4-stable is still maintained but if it is, we encountered this issue where interface binding would not work anymore when upgrading from 1.7.9 to 2.4.0.

This is probably due to an unfortunate copy paste of the strncpy line for method.

Note that this has been fixed in v3 by aa816e98a9081bde460c0dea36ba3ecd792bcc06.


lws-team:
Thanks... it's maintained in the sense if someone sends me a nice patch as you did, I'm happy to put it on v2.4-stable where it is now.  It's otherwise an "old version", since v3.0 has been out for quite a while.

PR #1453, opened 2018-11-15
Fix lws_hdr_total_length match the lws_hdr_copy

zw-git:
I upgrade my code to the matser, but I found a difference from v3.0-stable branch. 
```

@@ -512,13 +516,17 @@ LWS_VISIBLE int lws_hdr_copy(struct lws *wsi, char *dst, int len,
                return 0;

        do {
-               if (wsi->http.ah->frags[n].len >= len)
+               if (wsi->http.ah->frags[n].len + 1 >= len)
                        return -1;
                strncpy(dst, &wsi->http.ah->data[wsi->http.ah->frags[n].offset],
                        wsi->http.ah->frags[n].len);
                dst += wsi->http.ah->frags[n].len;
                len -= wsi->http.ah->frags[n].len;
                n = wsi->http.ah->frags[n].nfrag;
+
+               if (n)
+                       if (h != WSI_TOKEN_HTTP_COOKIE)
+                               *dst++ = ',';
        } while (n);
        *dst = '\0';

```
The lws_hdr_copy function may takes more bytes, but the lws_hdr_total_length  function same as before.
So, if I malloc the bytes base on lws_hdr_total_length, and then use lws_hdr_copy to get the header, it will failed.
And the comma is not add every frag.

### DIFF
--- a/lib/roles/http/server/parsers.c
+++ b/lib/roles/http/server/parsers.c
@@ -459,6 +459,10 @@ LWS_VISIBLE int lws_hdr_total_length(struct lws *wsi, enum lws_token_indexes h)
 	do {
 		len += wsi->http.ah->frags[n].len;
 		n = wsi->http.ah->frags[n].nfrag;
+
+		if (n && h != WSI_TOKEN_HTTP_COOKIE)
+			++len;
+
 	} while (n);
 
 	return len;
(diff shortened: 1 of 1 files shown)


lws-team:
Thanks, it makes sense... the handling for COOKIE is different since it wants to concatenate with ';', but my recent patch about it didn't adjust the length api to match.

It's pushed on master.

PR #1457, opened 2018-11-27
Fix OpenSSL ECC check

neheb:
With at least OpenSSL 1.1.0, checking the header is not enough. OpenSSL
does provide a way to check it directly though.

Fixes compilation without ECC support.


lws-team:
Thanks, it's pushed on v3.1-stable + master.

PR #1460, opened 2018-12-05
Allow IP-based SAN in automatic hostname check

tynril:
With OpenSSL, `X509_VERIFY_PARAM_set1_host` only checks matching hostnames and alternative names that are domain-based.

This change tries calling `X509_VERIFY_PARAM_set1_ip_asc` first, which attempts to parse the hostname as an IP address (v4 or v6). If this fails, it'll fall back to the current `X509_VERIFY_PARAM_set1_host` behavior.


lws-team:

Issue #1470, opened 2019-01-10
ping/pong with permessage delfate


Yimgo:
With v3.1.0, we have an issue when receiving a pong while having a binary message being transferred and deflated.

It appears that the permessage-deflate extension relies on the rsv bit of the first message (`wsi->ws->rsv_first_msg`) to decompress a message. Thus, when receiving a pong message, `wsi->ws->rsv_first_msg` is still set and the extension tries to decompress the message which leads to an inflate error and the connection being closed.

We added some debug to show the rsv and the rsv_first_msg every time we enter the lws_extension_callback_pm_deflate when reason is LWS_EXT_CB_PAYLOAD_RX: 

```c
lwsl_ext(" %s: rsv: %d, rsv_first_message: %d\n",
			 __func__, wsi->ws->rsv, wsi->ws->rsv_first_msg);
```

Here are the logs we have:

```
PMD_SERVER_NO_CONTEXT_TAKEOVER

Ext RX returned 0

post inflate ebuf len 508

__lws_set_timeout: 0x806886c00: 0 secs (reason 0)

lws_is_final_fragment: final 1, rx pk length 0, draining 0

0x806886c00: SSL_read says 4

lws_read_h1: h1 path: wsi state 0x117

lws_ws_client_rx_sm:    This RX frame Final 1

0x806886c00: SSL_read says 3583

lws_read_h1: h1 path: wsi state 0x117

lws_ws_client_rx_sm: passing 3583 to ext

 lws_extension_callback_pm_deflate: LWS_EXT_CB_PAYLOAD_RX: in 3583, existing in 0

 lws_extension_callback_pm_deflate: rsv: 64, rsv_first_message: 64

inflate ret 0, avi 2554, avo 0, wsifinal 1

-- rx grabbing held --

  lws_extension_callback_pm_deflate: RX leaving with new effbuff len 1023, ret 1, 

Ext RX returned 1

post inflate ebuf len 1023

lws_add_wsi_to_draining_ext_list: RX EXT DRAINING: Adding to list

__lws_set_timeout: 0x806886c00: 0 secs (reason 0)

lws_is_final_fragment: final 1, rx pk length 0, draining 1

rops_handle_POLLIN_ws: RX EXT DRAINING: Service

lws_remove_wsi_from_draining_ext_list: RX EXT DRAINING: Removing from list

lws_ws_client_rx_sm: doing draining flow

lws_ws_client_rx_sm: passing 0 to ext

[...]

PR #1475, opened 2019-01-11
Fix control messages are inflated

Yimgo:
RFC7692 states that control messages should not be compressed so there is no
need to inflate these messages.

There can be a bug if a control message is received while processing a
compressed message since lws relies on the RSV bit of the first message to
inflate the rx buffer or not.
Here we also check the opcode to only inflate a message if it is a data message.

Fixes: #1470


lws-team:

PR #1480, opened 2019-01-22
ws: fix coredump of lws_create_context

tanyifeng:
if we defined `LWS_MAX_SMP`, `context->pt` has max size of `LWS_MAX_SMP`, `info->context` may large than it, which caused coredump.

Signed-off-by: t00416110 <tanyifeng1@huawei.com>


lws-team:

PR #1495, opened 2019-02-15
Update client-handshake.c

fanc-chen:
in some platform， the size of sa46.sa4.sin_zero is not 8, but 6, so use 8 will cause coredump.


lws-team:
Thanks... it's pushed on v3.1-stable and master.

PR #1504, opened 2019-02-22
lib/event-libs/libuv/libuv.c: set m to 0 by default

aduskett:
Setting m to 0 by default will prevent "error: ‘m’ may be used uninitialized in this function"
while compiling with the option -DLWS_WITH_LIBUV=ON.

Tested against libuv v1.26.0.


lws-team:
Thanks... for some reason on Fedora 29 and Xenial that isn't flagged as a problem, although looking at it, it should be.

I pushed it on master and v3.1-stable.

PR #1506, opened 2019-02-26
client: check getaddrinfo return code when IPV6 is enabled

jcourtat:
The result structure from getaddrinfo was accessed before check of the return code,
and in case the function fails for any reason such as an unreachable host, it ends up
with a crash.


lws-team:
Thanks... I modified it very slighly so the variable itself is declared at the start of the basic block... some compilers care about it.

PR #1533, opened 2019-04-01
Fix bzero mis-detection on MinGW

elivdahan:
Using a C compiler ignores non-existent functions, and tries to link them anyway.

The compiler optimizes `bzero(buf, 1)` to `movb   $0x0,0xf(%esp)`, so bzero is
not called at all, and the linker succeeds.

Increase the buffer size to 100 to avoid this optimization.


lws-team:

Compiler is pretty smart...

PR #1535, opened 2019-04-01
fix windows build on VS2017 WIN10

v0l:
2 issues i noticed when trying to build with VS2017 on Win10

 - X509_NAME is defined by wincrypt.h as ((LPWSTR) 7) a different type to the openssl X509_NAME
 - Defining WINVER breaks a lot of the includes from the windows sdk, i assume this is used for mingw or something i dont know why would set this to a specific version when its automatically set to the windows sdk version you are targeting.

Sorry if something is wrong with these changes im a bit of a noob when its comes to C/C++. Please let me know if this is ok..


lws-team:
Thanks... since appveyor is also happy with it, I pushed it on master.

PR #1536, opened 2019-04-02
check wsi before accessing and return 0 (not ok) if null

diren-msft:
(same change as a58b389 in other code path)


lws-team:
Thanks, it's pushed on v2.4-stable.

PR #1537, opened 2019-04-03
http: fix broken basic auth on Firefox

pavelxdd:
Firefox sends HTTP requests with "Connection: keep-alive" header.
When LWS responds with 401 and WWW-Authenticate header, Firefox
doesn't show an authentication dialog until connection is closed.
Adding "Content-Length: 0" solves the problem.


lws-team:
Thanks, it's pushed on master and v3.1-stable.

PR #1540, opened 2019-04-05
keep protocol unchanged if upgrade does not provide a new protocol

kzhdev:
The Problem:
I have an lws service which handles both HTTP client and WebSocket client. The protocols are the following:
```
const struct lws_protocols s_protocols[] = {
  {"http", http_callback, 0, 0},
  {"ws", http_callback, 0, 0},
  {nullptr, nullptr, 0, 0}
}
```
When a websocket client connected to websocket server, some server doesn't not send the protocol ("ws") in the upgrade.  In client-ws.c if no protocol provided, it set the protocol to the first protocol in the list, which caused the WebSocket client stopped working.

The Solution:
If no protocol provided in upgrade header, keep current protocol unchanged.


lws-team:
Thanks, but this isn't backwards-compatible.  It's supposed to default to protocols[0] and there's (old) code out there that relies on it.

What you can do is pick the protocol that will be used as the default on a particular vhost instead of protocols[0].  You can do this by adding a pvo on the protocol you want to be the default for no-subprotocol connections on that vhost, named "default".  That's a one-liner if you use the JSON config, but in code it's a bit more... there's an example here, you would change "config" to "default", the value isn't important you can set it to "x" or whatever.

https://libwebsockets.org/git/libwebsockets/tree/minimal-examples/ws-server/minimal-ws-server-threads/minimal-ws-server.c#n66-78

and set the pvos in the vhost creation

https://libwebsockets.org/git/libwebsockets/tree/minimal-examples/ws-server/minimal-ws-server-threads/minimal-ws-server.c#n110

The same problem comes up when you listen with RAW role, which doesn't have a way to specify any subprotocol... it defaults to protocols[0] also, but you can separately control that by attaching a pvo "raw" to the vhost's protocol you want to be targeted by incoming raw connections.

kzhdev:
Andy,

My use case is not a ws-server. Mine is a client application which uses a single context for both HTTP client and WebSocket client. What is a proper solution for this? Or for the client application, it has to create a separate context for each protocol?

lws-team:
It seems I got the wrong idea from reading the description, sorry.

Is this what you need?

https://libwebsockets.org/git/libwebsockets/tree/include/libwebsockets/lws-client.h#n108-114

kzhdev:
The local_protocol_name does not solve the issue. The problem is that when a WebSocket client connects to a server which does not send protocol in the UPGRADE header. client-ws.c will default the protocol to the first one. In my case, it changed the protocol to 'http'. Therefore subsequent callbacks are delivered to http callback instead of ws callback.

lws-team:
Let me reopen this and I'll try the patch.

lws-team:
Does this (untested) variation work for you?

The problem with the current code is n is meaningful, and in the new case where we're proxying ws, the patch as it was doesn't initialize the proxied subprotocol name

```
diff --git a/lib/roles/ws/client-ws.c b/lib/roles/ws/client-ws.c
index cabca0fb7..13d8cd5ab 100644
--- a/lib/roles/ws/client-ws.c
+++ b/lib/roles/ws/client-ws.c
@@ -305,9 +305,17 @@ bad_conn_format:
        if (!len) {
                lwsl_info("%s: WSI_TOKEN_PROTOCOL is null\n", __func__);
                /*
-                * no protocol name to work from,
+                * no protocol name to work from, if we don't already have one
                 * default to first protocol
                 */
+
+               if (wsi->protocol) {
+                       p = wsi->protocol->name;
+                       goto identify_protocol;
+               }
+
+               /* no choice but to use the default protocol */
+
                n = 0;
                wsi->protocol = &wsi->vhost->protocols[0];
                goto check_extensions;
@@ -334,6 +342,8 @@ bad_conn_format:
                goto bail2;
        }
 
+identify_protocol:
+
 #if defined(LWS_WITH_HTTP_PROXY)
        lws_strncpy(wsi->ws->actual_protocol, p,
                    sizeof(wsi->ws->actual_protocol));
```

lws-team:
I assume all is well since I didn't hear any more.

PR #1542, opened 2019-04-05
Fix clean up crash

kzhdev:
Make sure tcr is not NULL.


lws-team:

PR #1547, opened 2019-04-08
http: refactor and fixes in lws_get_mimetype

pavelxdd:
- prioritize user-defined mimetypes over predefined server mimetypes - in case someone wants to serve images in .xml files, there was no way to redefine that.

- fix accessing memory out of string bounds - `if (!strcmp(&file[n - strlen(pvo->name)], pvo->name))` this was very dangerous without checking `n` beforehand.

- prefer case-insensitive comparison for extension matching - `jPG`, `Jpg`, `jpg`, `JPG`...

- other minor fixes and improvements - added logging and things like removing `if (n < 5)` (what if filename is `a.js`?)

### DIFF
--- a/lib/roles/http/server/server.c
+++ b/lib/roles/http/server/server.c
@@ -378,75 +378,73 @@ lws_select_vhost(struct lws_context *context, int port, const char *servername)
 	return NULL;
 }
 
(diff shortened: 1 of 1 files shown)


lws-team:

PR #1561, opened 2019-04-29
Cherry-picked 3 three fix from master to V3.1 stable

kzhdev:


lws-team:
Thanks... I appreciate getting the patch.

But the rule for the -stable branches is we can't add to the abi, only fix what's there already.  The reason is that targeting "v3.1" and its library soname (eg, libwebsockets.so.14) turns into a nightmare if "v3.1" and the same library soname libwebsockets.so.14 may or may not contain certain apis depending on the patchlevel.  packaging which only has the soname to go on is then unable to be sure the right dependent packages are there or not.

It's FOSS so you are welcome to patch it how you like for your own use, but I can't add it to the official v3.1.

Unfortunately the public api for this is part of a ton of stuff added post-3.1 on master but not yet released as v3.2.  I'm fighting a deathmatch on permessage-deflate improvements atm after that there's one more major thing I want to stabilize and there can be a v3.2 to provide this.

PR #1563, opened 2019-05-02
V3.1 stable

gapkalov:
gcc 8.3 compile fix


lws-team:
IIUI, you just want to backport  "OpenSSL: fix ECC support build-time check" to v3.1-stable?  If that's all that's needed I can do it without the Pull Request.

gapkalov:
No, I just need apply micro patch to compile libwebsockets with gcc 8.3

`From caedbb15ada05c3425ce4e5537722f06b9cd8c9c Mon Sep 17 00:00:00 2001
From: "dmitry.gapkalov" <gapkalov@gmail.com>
Date: Wed, 1 May 2019 22:00:50 -0400
Subject: [PATCH] compile fix

---
 lib/roles/http/server/lejp-conf.c | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/lib/roles/http/server/lejp-conf.c b/lib/roles/http/server/lejp-conf.c
index 0dd701d8..35024a3d 100644
--- a/lib/roles/http/server/lejp-conf.c
+++ b/lib/roles/http/server/lejp-conf.c
@@ -830,7 +830,7 @@ lwsws_get_config(void *user, const char *f, const char * const *paths,
 	unsigned char buf[128];
 	struct lejp_ctx ctx;
 	int n, m, fd;
-
+	m  = -1;
 	fd = lws_open(f, O_RDONLY);
 	if (fd < 0) {
 		lwsl_err("Cannot open %s\n", f);
-- 
2.21.0.windows.1 `

lws-team:
Great... for this kind of thing, pasting or emailing a patch or diff like that is perfectly fine.  I would simply extract the patch out of the PR anyway for those cases.

It's pushed on v3.1-stable and thanks for the patch.

PR #1577, opened 2019-05-16
Fix bzero misdetection also for GCC >=8

orgads:
The fix in 1d6128d1fe76e6148cd9955db38be39481526f65 worked for
MinGW with GCC7, but GCC8 evidentially got smarter, and it
also substitutes bzero->memset for larger arrays.

Set the emitted warning as an error to avoid misdetection.


lws-team:

PR #1592, opened 2019-05-29
Fix build with GCC 4.1

orgads:
There are 3 commits here, but this is still incomplete. These commits are quite trivial.

I pushed the other required commits separately because they worth a discussion.


lws-team:
It looks like some of the casts you're changing there can go based on changing the definition between char / unsigned char on the basis of which was used more, which is fine.  But there are a lot of "legally unnecessary casts" in this code to make microsoft's compiler happy.  So we need to see what appveyor made of it.

The other two are OK.

lws-team:
Changing the sign of .token in eff_buf is a bit of a big change... a lot of things use eff_buf.  Travis failed on it because it also tests permessage-deflate which also error out here... I added some small extra cast-removals to the patch and it builds OK.

Travis also does runtime testing for a variety of cases... I pushed it and the other patches let's see how it goes.

orgads:
Thank you.

PR #1595, opened 2019-05-29
tls: Fix yet another strict-aliasing issue on GCC 4.1

orgads:
I'm not sure about this fix.

Is `d2i_X509` supposed to modify the *pointer* ca_mem?


lws-team:
What it was set to is const, and nothing uses ca_mem_ptr after passing it into d2i_X509().  As far as I can tell by looking at it your change seems equivalent.  So I will push it on master shortly.

PR #1608, opened 2019-06-18
Modif makefile so the minimal example can compile.

nicovv44:


lws-team:
If we're going to add this, we need to add it to all the minimal examples, there's nothing really special about this one objectively.

What conditions does this help with?  It says "so it can compile" but it compiles fine here without this, and there is also -DLWS_WITH_MINIMAL_EXAMPLES=1 option in cmake that builds all the examples at the same time as the library.

PR #1613, opened 2019-06-20
Fixed LWS_WITH_ASAN with Clang

snej:
Building with LWS_WITH_ASAN was broken when using Clang, due to some flags being given that Clang doesn't recognize. (`-fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=leak`)

Fixed this by adding those flags only when the compiler isn't Clang.


lws-team:

PR #1616, opened 2019-06-25
enable libuv

ssrlive:
enable libuv, or the linux veraion will always using poll.


lws-team:
Thanks.  However the test client is pretty much deprecated... and just because the library has libuv enabled, it doesn't always mean that's what you want to use.  In fact lws allows you to enable multiple event libs at build time.

The minimal-http-server-eventlib shows how to select the event loop at runtime

https://libwebsockets.org/git/libwebsockets/tree/minimal-examples/http-server/minimal-http-server-eventlib/minimal-http-server-eventlib.c?h=master

although that's for server, if you want to provide a similar patch for a minimal client example that'd be great.

ssrlive:
OK

PR #1617, opened 2019-06-25
Clang compatibility and Windows pipe fixes

borrrden:
See discussion at bottom of #1291 

The summary is that since Windows does not have a reliable POSIX pipe and polling mechanism that `lws_cancel_service` often fails to broadcast messages when it is called.  This adds some Windows specific logic to address that in a threadsafe way.

In addition, some fixes to the CMake project are there to make sure that the ASAN mode is compatible with the clang compiler flags.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1350,7 +1350,11 @@ if (UNIX)
 	endif()
 endif()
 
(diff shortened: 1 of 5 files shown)


lws-team:
Thanks a lot... I pushed it on master and also backported it to v3.1-stable.

PR #1618, opened 2019-06-25
fix wrong return code in server.c

impegoraro:
I was trying to use `lws_vhost_create` and when the port is not available it did not return a null, did a little digging and found out that there're a couple of returns that return a positive number.

### DIFF
--- a/lib/libwebsockets.h
+++ b/lib/libwebsockets.h
@@ -2653,6 +2653,10 @@ enum lws_context_options {
 	 * example the ACME plugin was configured to fetch a cert, this lets
 	 * you bootstrap your vhost from having no cert to start with.
 	 */
+	LWS_SERVER_OPTION_FAIL_UPON_UNABLE_TO_BIND      = (1 << 27),
+	/**< (VH) When instantiating a new vhost and the specified port is
+	 * already in use, a null value shall be return to signal the error.
+	 */
 
 	/****** add new things just above ---^ ******/
 };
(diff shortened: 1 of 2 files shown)


lws-team:
These are not intrinsically wrong... as it says at the top of that function

```
/*
 * return 0: all done
 *        1: nonfatal error
 *       <0: fatal error
 *
 *       REQUIRES CONTEXT LOCK HELD
 */
```

For some cases like we try to bind a listening socket that's occupied by something else, it may be solved by killing the other thing.  Sometimes on eg, routers where lws is used, interfaces that vhosts want to bind to may disappear and reappear dynamically.

What's the actual condition that you have that gave this behaviour you're trying to fix?

impegoraro:
I was looking in lib/core/context.c:839 which specifically test for a negative number for error, maybe in here is where we should change then ? 

I was using a context to create multiples vhosts and when I ran the program twice the second instance did not die because the `lws_create_vhost` did not return a null pointer.

lws-team:
> I was using a context to create multiples vhosts and when I ran the program twice the second instance did not die because the lws_create_vhost did not return a null pointer.

Right... it should have hung around and actually started up OK if you killed the other instance.

It sounds like different people want different behaviours, if it's really the same thing they want to act differently I guess the answer is control it by a context flag

https://libwebsockets.org/git/libwebsockets/tree/include/libwebsockets/lws-context-vhost.h#n45

impegoraro:
oh ok, is there a flag to enforce that behaviour? I mean to return null if there's somebody else listening on that port?

lws-team:
No... what I mean is that would be the way to handle different expectations about what it should do... add a flag.

impegoraro:
ok, made some changes. whenever you can take a look. Thanks

lws-team:
Thanks... the changes are fine, but I can only add them to master since the option flags are part of the abi.  It's pushed on master with the option constant moved forward to (1 << 30).

PR #1623, opened 2019-07-01
Added LWS_CALLBACK_OPENSSL_SERVER_NAME callback 

zevv:
This allows for for updating SSL_CTX context after the SNI name is detected, but before
the SSL handshake is performed - handy for implementing MITM https
proxy, for example.

### DIFF
--- a/include/libwebsockets/lws-callbacks.h
+++ b/include/libwebsockets/lws-callbacks.h
@@ -153,6 +153,13 @@ enum lws_callback_reasons {
 	 * by the OS, for example if it is stored on a smartcard.
 	 * user is the server's OpenSSL SSL_CTX* */
 
+    LWS_CALLBACK_OPENSSL_SERVER_NAME = 102,
+    /**< this callback is called after a vhost accepted the connection and read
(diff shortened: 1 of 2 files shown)


lws-team:
In the case there is no callback handled and it just returns zero, this hanging if is broken isn't it?

```
	wsi = SSL_get_ex_data(ssl, openssl_websocket_private_data_index);
	if (vhost->protocols[0].callback(wsi,
				LWS_CALLBACK_OPENSSL_SERVER_NAME,
				vhost->tls.ssl_ctx, (char *)servername, 0))

	/* select the ssl ctx from the selected vhost for this conn */
	SSL_set_SSL_CTX(ssl, vhost->tls.ssl_ctx);
```

zevv:
Quoting Andy Green (2019-07-01 10:09:31)

> In the case there is no callback handled and it just returns zero,
> this hanging if is broken isn't it?

Oh dang, you're right - I messed up while taking it out of my project.
I'll fix and resubmit, sorry for the noise.

zevv:
Closing because of incompetence on my side.

PR #1632, opened 2019-07-11
openssl: add missing cleanups of error queue

hashok:
- Use ERR_get_error() instead of lws_ssl_get_error() where
  SSL context is not available.


lws-team:
Thanks, it's pushed on v3.1-stable.

PR #1646, opened 2019-07-31
solve compilation problems on vs2012

hjfbswb:
(1)There is no inttypes.h on vs2012.
(2)There is no function snprintf on vs2012.
(3)Vs2012 seems having a strict check on c variable declaration location.


lws-team:
Have you maybe missed the cmake tests for LWS_HAVE_SNPRINTF / __SNPRINTF?

Appveyor has them, but your patch kills the CI build there as it is

https://ci.appveyor.com/project/lws-team/libwebsockets/builds/26377922/job/romy2ch8tirrstxm

hjfbswb:
https://github.com/hjfbswb/libwebsockets/blob/6b79ee5d195f53243fbd62582eca51f34ca2dfaa/cmake/lws_config_private.h.in#L125

I added them(See above),  and it work well in my computer.
But cmake did not find neither snprintf nor _snprintf, how can this happen?

image

lws-team:
I don't know... but looking at how to solve it both ways, actually there is a function lws_snprintf() that is preferred to use compared to snprintf.

Internally, it actually uses vsnprintf().  So since this builds OK normally, I guess it can normally find `*snprintf()` when you build it, even if cmake can't find it for some reason.

How about modifying your patch so it just assumes the existence of _snprintf() if on windows, it looks like it might work around it.

hjfbswb:
(1)I think the better way is replacing `snprintf` with `lws_snprint`, and i did it.
(2)Difine `int64_t`, `uint64_t`,`int32_t` and `int16_t` to solve compilation errors on vs2008.

lws-team:
Thanks... I pushed it on master and backported it to v3.1-stable as well.

PR #1648, opened 2019-08-02
solve compilation problems on android

hjfbswb:
solve compilation problems on android.


lws-team:
Thanks, it's pushed on master and v3.1-stable.

PR #1661, opened 2019-08-19
Fix null checks

mlgiraud:
Some null checks were performed after dereferencing, rendering them useless.
I also added some more where they were missing entirely.


lws-team:

Because of what these are checking though, if you are passing lws apis NULL lws_context, you have some much worse problem in user code... it's no so crazy to just crash on that I think.

mlgiraud:
Imho it's the users responsibility to check the return code of the function after calling it. Just crashing can cause some serious trouble depending on the scenario the application is used in.

lws-team:
> Just crashing can cause some serious trouble depending on the scenario the application is used in.

Well, I merged the patch, so I am not really arguing against it.

However there are valid perspectives that go in other directions for this kind of basic problem... crashing makes sure it's in the developer's face during development.  This particular one isn't something that can come up except by some gross problem in the user code that should be fixed rather than handled.  Eg, apis like strcmp() do not try to handle NULL, they just crash as a sort of roughneck assert.

Anyway... it's in.

PR #1663, opened 2019-08-22
Change SmartOS detection/build to all Illumos

mikeowens:


lws-team:

PR #1668, opened 2019-08-25
sul_compare integer overflow bug

Atl4sx:
If the value of `a` is much bigger than `b` it may lead to an integer overflow bug.
I found this issue when our client app hanged in `poll` function call for 23 hours waiting for SSL certificate check to trigger.


lws-team:
Good catch, thanks.... it's pushed on v3.2-stable and master.

Atl4sx:
@lws-team Thx for your great work by the way!

PR #1696, opened 2019-09-25
Allow more logging for release builds

kristjanvalur:
In a production environment, I needed more information available.  In particular, I'm looking to see why sockets closes (timeout?).
Ideally, I would like to have some sort of "getlasterror" or similar since errors that occur during connection or hangup often don't end up being available in the callback.  But this change, at least, enables information logging in RELEASE if enabled via cmake flag.


lws-team:
I think you have a good point and we should do something, but this is a bit one-shot.  Also, we can't do new work like this on old lws version like v3.1-stable, it has to be done on master and then backported if it makes sense.

It might make sense to define a bitmask for build-time as well as the runtime bitmask... you can do bitmask tests in the preprocessor like #if (LWS_BUILD_DEBUG_MASK & LLL_INFO)  But in a handful of places we make stuff conditional on _DEBUG too they would need fixing up.  If you don't set it via cmake, it defaults to the traditional bitmask values depending id debug build or not.

kristjanvalur:
Sure.  This is a temporary solution I employed for UEWebsockets (ue4 plugin).  We are trying to diagnose why connections close on clients, and have found that the information returned on callback events is sorely lacking.  Looking at the source, I see that for example timeouts are only logged.  Close reasons appear to be official "websocket" reason codes, etc.  I didn't want to go and hack in stuff.  I'll close the pull request, but perhaps you can consider this a suggestion to improve on error reporting somehow?  Cheers!

lws-team:
> Looking at the source, I see that for example timeouts are only logged

The protocol that's bound to it will hear that it closed which is what it should do for runtime following of what happened.  Things closing or timing out is generally a normal thing that doesn't want to assert(0) the whole app.  So optionally logging it additionally is the right way.  What else should it do about that?

Lws also builds and is used in RTOS context on devices that don't have any spare room for .rodata overhead for tons of log text.  So the existing situation is a compromise where you can build in everything for debug and strip most of it out for release.

I think actually logging and management of logs is a huge subject people are probably writing doctoral theses on so I can imagine there is plenty of scope to improve it.  But I can also imagine throwing work on it that more or less never pays off for anyone.  So... open to specific suggestions.

kristjanvalur:
The use case: used in an unreal game (via UEWebSockets plugin). Out of
thousands of players, a handful have problems connecting with the ws
server, or continuously get the connection closed. We are trying to
diagnose the reasons for these remote users. Knowing that a connection was
closed due to a local timeout (eg pingpong) or from remote, or the specific
reason, would be useful. (for example the difference between econnreset and
econnaborted with tcp). Then there are cases when incorrect headers, ie
invalid response when connecting. Getting access to the specific response
would be useful in this case.
Our current approach is to store the last 20 log messages or so, and relay
them to our servers when the websocket client experiences unexpected
problems. I hope that will help us weed out bad isp providers, or other
such things to help our clients.

lws-team:
OK, I get the situation a bit better.  Let me have a think about it.

kristjanvalur:
Thanks.  No rush.  Perhaps something like
const char *lws_get_last_error(struct lws *wsi);
where the lib could optionally store strings (owned by the wsi struct)
describing the last abnormal thing that happened.
This would be backwards compatible and not require making internal apis
more complicated.

Anyway, my increased logging approach is being deployed for testing soon,
that will also hopefully shed light on the problem that we are having :)
Cheers!

kristjanvalur:
By the way, the Implementor of UEWebSockets made the api a bit awkward, in that the socket creation and conneciton was done in one single call.  That made it impossible to attach a failure callback that would catch the errors generated _during_ the Connect() call (only subsequent asynchronous events were caught).  I have since changed this.  Along with uploading client logs and getting the immediate error callback, I expect that at least our connection problems will be understood.

PR #1711, opened 2019-10-07
fix build errors and warning on illumos; fix passing cgi envs using setenv()

zhiwenzheng:


lws-team:
Thanks, they're pushed on v3.2-stable and forward-ported to master (the problems with some c++ compiler and the typedefs are already solved there).  I checked both branches build OK on Linux but the CI also builds it on OSX + windows so see how it goes.

PR #1717, opened 2019-10-16
CMakeLists.txt: fix static build with openssl

ffontaine:
openssl can depends on -latomic so use pkg-config (if available) to
retrieve these static dependencies otherwise build will fail because
HMAC_CTX_new test will return a wrong result

Fixes:
 - http://autobuild.buildroot.org/results/65d0528b208c0a470264f7e2433be89425971dd7

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


lws-team:

PR #1728, opened 2019-10-22
Fix warnings/errors for WIN64

kristjanvalur:
Various issues on windows 64 bit builds.  Socket types cannot be ints, size_t isn't int, and externs must be extern when declaring arrays, as in `extern int my_array[];`.  Newly introduced extern arrays caused errors when being compiled into non-dlls on windows.
Problems were discovered in 3.2 stable, and if accepted, possibly should be back-ported.

### DIFF
--- a/include/libwebsockets.h
+++ b/include/libwebsockets.h
@@ -100,8 +100,6 @@ typedef unsigned long long lws_intptr_t;
 #else
 #define LWS_EXTERN extern __declspec(dllimport)
 #endif
-#else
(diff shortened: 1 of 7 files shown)


lws-team:
Thanks... the last two I squashed into one, and since you changed the declaration of a plat function, also made the same change in the other instances of it in each plat.

There's actually a helper for the pointer arithmetic to int, lws_ptr_diff(), I edited that in as well.

It's on master now.

PR #1736, opened 2019-10-27
Windows fixes

kristjanvalur:
This pull requests combines three things:
1) explicit conversion from size_t to int to remove warnings on win64 platforms
2) Added error logging for some windows socket APIS
3) fallback if getprotobyname("TCP") fails, as has been observed in the wild.  The value 6 is universal, but I assume that there is a good reason for using this api never the less.

There remains one case of implicit size coercion not addressed,
```
  D:\git\libwebsockets\lib\roles\http\server\server.c(701): warning C4244: '=': conversion from '__int64' to 'char', possible loss of data
```
I leavi ti to the discression of the maintainer to find an adequate fix / check for that.

### DIFF
--- a/lib/plat/windows/windows-sockets.c
+++ b/lib/plat/windows/windows-sockets.c
@@ -70,8 +70,13 @@ int
 lws_plat_set_nonblocking(lws_sockfd_type fd)
 {
 	u_long optl = 1;
-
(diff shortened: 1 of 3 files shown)


lws-team:
Thanks... I added a fix for the extra int64 -> char one into the related patch and pushed them on master with a backport on v3.2-stable as well.

PR #1739, opened 2019-10-28
Fix incorrect check on the result of ioctlsocket.

kristjanvalur:
I accidentally reversed the return check in a previous commit.  So, a spurious error is now logged. My apologies.


lws-team:
Thanks... on master I just fixed the original patch directly.  On v3.2-stable I pushed this.

PR #1743, opened 2019-10-29
Get thread service index wsi belong to

kala13x:
Its needed when using external 'epoll' event loop (SMP) with separated epoll file descriptors for each service thread.


lws-team:

PR #1745, opened 2019-10-31
Ensure that timeout events are serviced even when polling on windows.

kristjanvalur:
Previously the timeout queue would only be polled if the supplied timeout was non-zero.  This caused functionality such as ping/pong and other timeout dependent things to stop working in polling mode.
This patch makes the behaviour similar to the platform esp32_service.c


lws-team:
Thanks... yes that check is from when checking for pending events was an expensive thing before the sul change...  I pushed this on master + v3.2-stable along with another patch renaming the internal function for "checking" the interval to the next scheduled event to make it clear it services those events (and just happens to return the interval to the next future event).  The second patch also removes the corresponding check your patch removes on windows.

PR #1748, opened 2019-11-04
make acme client support ACMEv2 http-01

zhiwenzheng:


lws-team:
Thanks... it looks like your editor corrected some line length stuff in there but didn't use the same formatting, I went though and fixed hanging whitespace and line continuation indent.  I also removed the four lines that made trouble in Travis, they were all effectively NOPs.

You removed the stuff about registering the user email, IIRC this has an impact if the guy ever wants to revoke these certs.  You're sure that's necessary?

Since the existing SNI stuff is deprecated at ACME, this is better than what we have either way, I pushed it on master.

zhiwenzheng:
I do not need for revoke cert for now, I tried registering account without email and it worked, so I didn't bother using it, I really don't know what impact it will have.

lws-team:
It means if someone gets a copy of your cert, you have no identity with Let's Encrypt where you can establish you have the right to revoke the leaked cert.  So the attacker can just "be you" until his cert runs out of time.  It's not good I think.

zhiwenzheng:
OK. I will try to add the user email.

PR #1751, opened 2019-11-04
acme client: add user email in account registration

zhiwenzheng:


zhiwenzheng:
I also removed hanging whitespace on the user email patch, don't know why it cause conflicts.

lws-team:
Hanging whitespace doesn't cause conflicts, it's complained about by my git tools so I get rid of it.

Instead of push this I extracted the bit with the change and amended the earlier patch to have it.  Otherwise this patch makes a lot of whitespace-only changes I removed from the earlier patch.  Anyway it's all OK now AFAIK and pushed on master.

PR #1755, opened 2019-11-06
let acme CSR support wildcard domain

zhiwenzheng:
I pushed ACMEv2 dns-01 support using 88811.me(a ipv6 DDNS service which is part of my project) to the acmev2-dns01-88811_me branch of my github repo, the patch contains the required change on master, the changes are compatible with http-01.


lws-team:
Thanks, I pushed it on master.

We're making a little bit of a problem with this that the new stuff is only implemented for OpenSSL.  The original, admittedly deprecated, acme stuff worked on both OpenSSL and mbedTLS equally.  AFAICS the elements[] indexes you changed don't impact the mbedtls implementation because it only knows how to handle the first five anyway, implementing SAN on mbedtls looks like a whole subproject of its own

https://tls.mbed.org/discussions/crypto-and-ssl/sample-to-create-a-x509-certificate-with-alternate-name

I pushed a little patch before this one that updates the apis in the mbedtls RSA stuff so it at least builds if LWS_WITH_ACME=1... I guess a problem for another day.

PR #1766, opened 2019-11-15
Add .woff2 support in v3.1

shenchsh:
It would be nice to have `.woff2` support in v3.1:)


lws-team:
Thanks... it's already in the default mimetype list in v3.2-stable and master, I pushed it on v3.1-stable.

FYI you can add other mimetypes at runtime, they don't have to come from the default list.

https://libwebsockets.org/git/libwebsockets/tree/minimal-examples/http-server/minimal-http-server-mimetypes/minimal-http-server-mimetypes.c#n22

shenchsh:
that's nice, thanks!

PR #1790, opened 2019-12-08
Fixing Android build errors...

frink:
This seems to be residue from some variable renaming


lws-team:
Thanks... I picked the one-liner and pushed it on master.

You won't be able to follow master using git pull most of the time.  You'll need to take the approach mentioned in the third para here

https://libwebsockets.org/git/libwebsockets/tree/READMEs/README.release-policy.md

PR #1795, opened 2019-12-10
Basic Auth Callbacks

4cad:
_Now with fixed whitespace_

I just wanted to show you the code to validate the approach before going through and changing all of the minimal example HTTP mounts to add the extra mount parameter. I did it this way to make it easy to add the more generic Authorization callback later without having to add a new mount parameter.

Also, I am going to revert my changes to test-server I just wanted to demonstrate the change in action. What testing do you think is needed to go along with the change?

I apologize for the newbie questions, I am just trying to get a feel for how you do things on this project.

### DIFF
--- a/include/libwebsockets/lws-context-vhost.h
+++ b/include/libwebsockets/lws-context-vhost.h
@@ -1040,6 +1040,16 @@ enum lws_mount_protocols {
 	LWSMPRO_CALLBACK	= 6, /**< hand by named protocol's callback */
 };
 
+/** enum lws_authentication_mode
+ * This specifies the authentication mode of the mount. For compatibility reasons, if
+ * basic_auth_login_file is non-null then basic auth file
+ */
+enum lws_authentication_mode {
+	LWSAUTHM_DEFAULT = 0, /**< default authenticate only if basic_auth_login_file is provided */
+	LWSAUTHM_BASIC_AUTH_CALLBACK = 1 /**< Basic auth with a custom verifier */
+};
+
+
 /** struct lws_http_mount
  *
  * arguments for mounting something in a vhost's url namespace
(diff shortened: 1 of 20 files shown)


include/libwebsockets/lws-context-vhost.h
```diff
@@ -1040,6 +1040,16 @@ enum lws_mount_protocols {
 	LWSMPRO_CALLBACK	= 6, /**< hand by named protocol's callback */
 };
 
+/** enum lws_authentication_mode
+ * This specifies the authentication mode of the mount. For compatibility reasons, if
+ * basic_auth_login_file is non-null then basic auth file
```
lws-team: ...comment seemed to die halfway

include/libwebsockets/lws-context-vhost.h
```diff
@@ -1040,6 +1040,16 @@ enum lws_mount_protocols {
...
+ */
+enum lws_authentication_mode {
+	LWSAUTHM_DEFAULT = 0, /**< default authenticate only if basic_auth_login_file is provided */
+	LWSAUTHM_BASIC_AUTH_CALLBACK = 1 /**< Basic auth with a custom verifier */
```
lws-team: You don't need to take care about =0 or =1, enum is specified to start at 0 and add one each time.

lws-team:
There's a member `.auth_mask` that's an unsigned int atm... instead of add a new member you can reserve say 4 upper bits of that to select the auth method for the mount (0 meaning none so default is the same).  That way nobody has to mess with their mount structs (there are dozens in lws tree itself as well).

Otherwise looks like it will be good to me.

include/libwebsockets/lws-context-vhost.h
4cad: Yeah, I was just copying the code right above for `lws_mount_protocols` to be consistent. I am fine either way

include/libwebsockets/lws-context-vhost.h
4cad: Fixed

lws-team:
Thanks, I squashed the patches into one and cleaned it up a little for line lengths and line endings, and pushed it on master.  So please check that still does what you want.

lws-team:
What I pushed works and I can close this / doesn't work / needs something else?

4cad:
Awesome, thanks for doing that - I will take a look tonight to validate.

4cad:
Yup, everything works for me as I would expect it to. Thanks!

lws-team:
Great, thanks for the patch.

PR #1798, opened 2019-12-12
lib/core-net/close.c: Fix off-by-one error for lws_inform_client_conn…

xor-gate:
This fixes an off-by-one error by letting the compiler help us by calculating the string size. We where missing the last character in our logline.

### DIFF
--- a/lib/core-net/close.c
+++ b/lib/core-net/close.c
@@ -452,10 +452,10 @@ just_kill_connection:
 	     lwsi_state(wsi) == LRS_WAITING_DNS ||
 	     lwsi_state(wsi) == LRS_WAITING_CONNECT) &&
 	     !wsi->already_did_cce && wsi->protocol) {
-		static const char reason[] = "closed before established";
+		static const char fail_reason[] = "closed before established";
 
 		lws_inform_client_conn_fail(wsi,
-			(void *)reason, sizeof(reason));
+			(void *)fail_reason, sizeof(fail_reason));
 	}
 #endif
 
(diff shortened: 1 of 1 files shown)


lws-team:
Thanks, I pushed it on master

xor-gate:
Is it possible you port it also to the v3.2-stable branch and create a new bugfix release on 3.2. As there are some other fixes on the v3.2.x release but currently using an unreleased version. Thanks!

lws-team:
I did the backport and tagged out v3.2.1... but the value of the tagging is mostly delusion, people should follow the -stable branches if they are following old versions.  The only meaningful way to do the tagging is tag every patch on there as it's added, and what's the point of that extra work compared to users just following the -stable branch?

xor-gate:
I do not agree this mostly delusion because actual fixes on a stable release are performed. For traceability a snapshot in time (version tag) is better than pointing to a commit hash in the tree or downloaded zipfile. It doesn't have to be done for every commit, but at certain points or highly important bugfixes it should be done. E.g fixing the null pointer dereference in the ping/pong timer handler which crashed our application after upgrading to 3.2.0 and was fixed in a later commit on the stable branch which is fairly severe problem.

lws-team:
As I say it's mostly delusion... commit 9f9dba9f209403c1e1ccf717b1252dd0e2bd799c is the same as v3.2-stable head currently and that's the same as v3.2.1 tag; everything was the same before I tagged it.  Nobody did any special QA flow on it.

I can see the point if something critical was coming.  But even then it's just worse than follow v3.2-stable since you are leaving all the other fixes inbetween on the table.

Guys are picking v3.2.0 tag... the best we can say is tagging out these point releases is trying to limit the damage people who have the idea they particularly have any meaning from deferring using fixes that are already pushed on v3.2-stable.

Going back to the ping / pong thing a much preferable way to change this is users help out with testing and QA for the things they care about before an x.y release, basically doing the fixes that would go in the stable branch earlier so they can go in the original release.

xor-gate:
I missed the point that you tagged it already v3.2.1 so indeed v3.2-stable HEAD and the tag are the same. I think are on the same page, thanks for the response!

PR #1800, opened 2019-12-19
Fix SSL negotiation errors handling in client

wojtekka:
If a client connects to a SSL server and the server sends handshake
alert (e.g. no matching ciphers) SSL_connect() fails, but because
SSL_ERROR_SSL return value is not handled, it's not considered a
failure. SSL_want_read() will return 1 and the client will happily wait
for more data from the server. Now if the server closes connection after
sending handshake alert, POLLIN event will be triggered,
lws_tls_client_connect() called again, but SSL_connect() will fail
without calling read(), so the client will end up consuming 100% CPU
because POLLIN will be triggered repeatedly.

Similar error handling is used in lws_tls_server_accept() and the
condition checks for SSL_ERROR_SSL. Using the same condition in
lws_tls_client_connect() fixes the problem.

Tested with OpenSSL 1.0.2k.

### DIFF
--- a/include/libwebsockets/lws-tokenize.h
+++ b/include/libwebsockets/lws-tokenize.h
@@ -141,3 +141,95 @@ lws_tokenize(struct lws_tokenize *ts);
 
 LWS_VISIBLE LWS_EXTERN int
 lws_tokenize_cstr(struct lws_tokenize *ts, char *str, int max);
+
+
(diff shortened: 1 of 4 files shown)


lws-team:
Thanks, it's pushed on master + v3.2-stable

PR #1814, opened 2020-01-06
lws_create_vhost compile failure with +SOCKS5 -client.

jkroonza:
If LWS is build with SOCKS5 support, but WITHOUT_CLIENT then a compile
failure occurs which this fixes.

Signed-off-by: Jaco Kroon <jaco@iewc.co.za>


lws-team:
Thanks, it's pushed

PR #1815, opened 2020-01-06
core-net/vhost Compile errors with SOCKS5 but WITHOUT_CLIENT.

jkroonza:
Signed-off-by: Jaco Kroon <jaco@iewc.co.za>


lws-team:
Thanks, it's pushed

Issue #1817, opened 2020-01-10
http client connect: Setting iface clobbers addrlen for connect()


chrthi:
When I do a `lws_client_connect_via_info()` with `host` set to an `ipv4address:port` and `iface` to a local IPv4 address, the following happens (line numbers refer to commit 04ec3d6f, the current master):
- In `lib/roles/http/client/client-handshake.c:472`, local variable `n` is set to `sizeof(struct sockaddr_in)`
- In line 569, `n` is overwritten by the result of lws_socket_bind(), which is a random port number
- In line 595, `n` is passed as the `addrlen` argument to `connect()`. `connect()` returns -1/EINVAL.

PR #1819, opened 2020-01-10
http client: Fix overwite of addrlen for connect

chrthi:
The addrlen argument to connect() was overwritten by the
lws_socket_bind() result, which is a port number.
Fixes https://github.com/warmcat/libwebsockets/issues/1817


lws-team:
Ouch... thanks a lot for the fix, it's pushed on master.  I checked and v3.2-stable has the same problem, I backported your fix there too.

PR #1834, opened 2020-01-22
lws_ring: use ___mtail instead of hardcoded "tail"

piona:
Changed harcoded member name "tail" into ___mtail in lws_ring_consume_and_update_oldest_tail macro.


lws-team:
Thanks, it's pushed on master and v3.2-stable.

PR #1855, opened 2020-03-02
event lib: fix LWS_WITH_GLIB typo

KarlVogel:
Build with LIBEVENT was broken due to a type / copy and paste error.


lws-team:
Ooops... thanks... it's pushed on master

PR #1867, opened 2020-03-21
lws_gencrypto_jws_alg_to_definition:

mcd500:
Fix segfault when reached end of array.

I found this on v3.2 and going to make the same PR for v.4.0 and master.

Please apply


lws-team:
Thanks Akira... I pushed it on master, don't worry about backporting I can do it here (at least, usually).

mcd500:
You are most welcome. I will make patches against master next time.

PR #1868, opened 2020-03-21
lws_gencrypto_jws_alg_to_definition:

mcd500:
Fix segfault when reached end of array.


lws-team:
Thanks it's on all three branches.

PR #1871, opened 2020-03-23
http server: Using default filename also for subdirectories.

karlrupp:
If a user sets a default filename for a http mount (`.def` in `lws_http_mount`),
eg. `default.html`, then a GET request for `/` correctly forwards to

    /default.html

However, without this commit the default filename is not taken into account for subdirectories. Thus,

    GET subdir/

will forward to

    subdir/index.html

instead of the expected

    subdir/default.html

This commit changes the behavior such that the user-provided default filename is also used for subdirectories.


lws-team:
Thanks, it's pushed on master and v4.0-stable

PR #1879, opened 2020-04-06
Fix unused variable warning.

jkroonza:
after 6bb116b8d8d59139416ceb8cb3ca522b3fa9adb3 the variable p could be
declared without being used.

https://bugs.gentoo.org/716386 refers.

This just fixes that.


lws-team:
Thanks, the test got out of whack with the code that actually uses it or not.

I pushed your patch on v4.0-stable and master.

PR #1888, opened 2020-04-13
License needed

Octavia93:
Is this file also licensed under MIT license, I don’t see a license header like the other files in this repo?


lws-team:
Oh... well, the LICENSE file makes it clear everything except stuff that is explicitly using another liberal license is MIT.  But it's definitely a oversight... I pushed you patch on v4.0-stable and master, thanks.

PR #1896, opened 2020-04-23
Fix unused-variable error in dhcpclient.c

aduskett:
dhcp_entry_names is only used when _DEBUG is defined, so wrap the array in
#if defined(_DEBUG)

Signed-off-by: Adam Duskett <Aduskett@gmail.com>


lws-team:
Thanks, it's pushed on master + v4.0-stable.

aduskett:
You're welcome!

PR #1912, opened 2020-05-08
Http2 prior knowledge

greenday9:
This adds support for H2 prior knowledge support for clear text HTTP/2 without upgrade.  Added VHOST option LWS_SERVER_OPTION_H2_PRIOR_KNOWLEDGE for enabling on server and ssl_connection option LCCSCF_H2_PRIOR_KNOWLEDGE for enabling on client.  Note that these options have no impact for TLS connections as the normal ALPN upgrade path is followed.  Tested with lws-minimal-http-client and curl --http2-prior-knowledge against lws-minimal-http-server.


lws-team:
Thanks a lot... I squashed the two patches into one and pushed it on master.

PR #1926, opened 2020-05-26
CMakeLists: add the option to disable Werror

jsmolic:
This commit adds the option to disable `Werror` compiler flag. Though it is ideal to use this, it was recently causing a lot of problems in the Gentoo ebuild package so it's not really practical to carry this option to the downstream package.\
Please review if adding this switch would be acceptable.

Thanks,
Jakov


lws-team:
> recently causing a lot of problems in the Gentoo ebuild package

I think the problem is already there, -Werror is just forcing it to be confronted.

Can you help me understand what kind of warnings were coming?

jsmolic:
There were some compile problems recently, as reported on bugzilla:
https://bugs.gentoo.org/719878
https://bugs.gentoo.org/720020
https://bugs.gentoo.org/716386

We've had some internal discussion (also mentioned in https://github.com/gentoo/gentoo/pull/15637) and concluded it would be more practical to drop `Werror`, this would provide us easy switch which can be included in the ebuild.

lws-team:
Well, this is about unused var... we have -Werror because we don't want to let any warnings creep in on any platform.  But I agree unused var warning is not worth the effort of bugzillas etc.

It's not going to help on 4.0 but on master and future releases I have dramatically more CI happening, basically there's a 32 core machine dedicated to CI now and a new tool building on several distros in systemd-nspawn.

https://libwebsockets.org/sai/

Is there a way we can combine giving you a way to avoid it with some details of how to set up a representative Gentoo container that will let me build it with -Werror and continue to fix things?

jsmolic:
> It's not going to help on 4.0 but on master and future releases I have dramatically more CI happening, basically there's a 32 core machine dedicated to CI now and a new tool building on several distros in systemd-nspawn.
>
> https://libwebsockets.org/sai/

I like the CI setup :)


> Is there a way we can combine giving you a way to avoid it with some details of how to set up a representative Gentoo container that will let me build it with -Werror and continue to fix things?

Well you could set up nspawn container with Gentoo stage3 tarball and test the build from there. I think something like this would be the general procedure: https://gist.github.com/TrentSPalmer/a929dfdb0cdb7dd946d313bde708900e
After that, you can emerge libwebsockets and all dependencies. It would have to be libwebsockets 9999 version ebuild as that one builds directly from master branch in this repository, and it would have `Werror` option enabled (though 9999 ebuild it's not in the Gentoo repository yet but we will add it soon).

lws-team:
Thanks... I pushed the patch and started on that... this being gentoo, it takes a while to "emerge" things.  I plan to just build it outside the packaging but inside the gentoo systemd-nspawn... it's possible to control what gets tested with what using an in-tree JSON file

https://libwebsockets.org/git/libwebsockets/tree/.sai.json

if you have particular mixes of lws options let me know.

lws-team:
Gentoo nspawn is building 25 build variations in the CI each push now.

jsmolic:
Thanks for accepting.


> Gentoo nspawn is building 25 build variations in the CI each push now.

That should uncover any possible issues on Gentoo, if I find some other build configuration which would be worth testing I'll let you know.

PR #1927, opened 2020-05-27
openssl: fix improper use of negative value

SeonahMoon:
SSL_get_fd() can return a negative number. (https://linux.die.net/man/3/ssl_get_fd)
'fd' should be checked if it is valid or not before being used improperly.


lws-team:
Thanks, it's pushed on master... I think maybe this is not a problem possible to meet in reality, since lws prepares the fd and only when connected hands it to OpenSSL to use in the tls tunnel, it's always valid then.  But if it's Coverity or similar complaining, no reason not to make it happy so there is less noise.

Issue #1929, opened 2020-05-28
VS cmake build error


Arenoros:
commit https://github.com/warmcat/libwebsockets/commit/6d04fb74815bac5d646f0de8fa09bd1d80ea85bd added line https://github.com/warmcat/libwebsockets/blob/a926e9d93bfcac352c38a94a99fe7022594d1562/CMakeLists.txt#L719 break CHECK_FUNCTION_EXISTS tests on VisualStudio, because CMAKE_REQUIRED_FLAGS in VS translated as
```
<AdditionalOptions>%(AdditionalOptions) C:/Users/Areno/.conan/data/zlib/1.2.11/_/_/package/16476d40ac9fb910d3ecd667186e25baab6b2f88/lib/zlib.lib</AdditionalOptions>
```
It's leading to build test errors:
[CMakeError.log](https://github.com/warmcat/libwebsockets/files/4697424/CMakeError.log)

lws-team:
Hm what cmake options are you using with that?  lws builds it on windows in CI now with various scenarios, eg,

https://libwebsockets.org/sai/index.html?task=cccfb2504064753ac74a70d75f89661240bd9315184faf15b99dbe492faa1d82

it's using the toolchain that came with VC19 without using VC19, but via commandline cmake.

You're using some cmake options that enable zlib?  Which ones?

> C:/Users/Areno/.conan/data/zlib/1.2.11/_/_/package/16476d40ac9fb910d3ecd667186e25baab6b2f88/lib/zlib.lib(1): error C2059: syntax error: '!' [C:\Users\Areno\.conan\data\libwebsockets\4.0\_\_\build\9bcce41d6c6a42de458f797a8a9543960ad24866\build_subfolder\CMakeFiles\CMakeTmp\cmTC_0bf45.vcxproj]

Arenoros:
@lws-team, I build from conan and use standalone cmake-3.16, It's bug can easy fix if use CMAKE_REQUIRED_LIBRARIES or CMAKE_REQUIRED_LINK_OPTIONS instead CMAKE_REQUIRED_FLAGS. I can create PR for fix this

lws-team:
Thanks a lot for the patch, it's pushed on master.

PR #1930, opened 2020-05-28
fix VS cmake build error #1929

Arenoros:
standalone cmake generate <AdditionalOptions>.../zlib.lib</AdditionalOptions> and breaks compile-time tests


lws-team:
Thanks a lot, it also cleaned up the pthead warnings introduced by the refactor patch.  It's pushed on master.

Issue #1941, opened 2020-06-04
unable to read client .pem or .der from memory


alexeadem:
Hi,

I was unable to read my cert from memory and I realized it was missing the `lws_tls_alloc_pem_to_der_file` code available on the server-side.

Here is what I added to make it work. 

`lib/tls/openssl/openssl-client.c`
```
        X509_STORE *x509_store;
@@ -814,14 +817,36 @@ lws_tls_client_create_vhost_context(struct lws_vhost *vh,
                }
                lwsl_notice("Loaded client cert %s\n", cert_filepath);
        } else if (cert_mem && cert_mem_len) {
-               n = SSL_CTX_use_certificate_ASN1(vh->tls.ssl_client_ctx,
-                                                cert_mem_len, cert_mem);
+
+               //lwsl_hexdump_notice(cert_mem, cert_mem_len);
+               lws_filepos_t flen;
+               uint8_t *p;
+
+               if (lws_tls_alloc_pem_to_der_file(vh->context, NULL, cert_mem,
+                               cert_mem_len, &p, &flen)) {
+                       lwsl_err("%s: couldn't read cert file\n", __func__);
+
+                       return 1;
+               }
+
+               n = SSL_CTX_use_certificate_ASN1(vh->tls.ssl_client_ctx, (int)flen, p);
+               
                if (n < 1) {
-                       lwsl_err("%s: problem interpreting client cert\n",
-                                __func__);
```

Also, I didn't see the code to read the key from memory.

```
+               n = SSL_CTX_use_PrivateKey_ASN1(EVP_PKEY_RSA, vh->tls.ssl_client_ctx, p,
+                                               (long)(long long)flen);

```

I added the code so it can read the key from memory from the client-side

```
--- a/lib/tls/openssl/openssl-client.c
+++ b/lib/tls/openssl/openssl-client.c
@@ -531,7 +531,10 @@ lws_tls_client_create_vhost_context(struct lws_vhost *vh,
                                    const char *cert_filepath,
                                    const void *cert_mem,
                                    unsigned int cert_mem_len,
-                                   const char *private_key_filepath)
+                                   const char *private_key_filepath,
+                                       const void *key_mem,
+         
```

Let me know if you want me to do you do a pull request so you can see the changes.

lws-team:
Yes please send a PR thanks

lws-team:
Thanks for the patch, it's pushed on master

PR #1942, opened 2020-06-05
issue #1941 and added support for ca, crt and key from mem

alexeadem:
I only tested with OpenSSL libraries. Mbedtls not tested.


lws-team:
Thanks, it's pushed on master

PR #1978, opened 2020-07-19
libcap should be linked instead of libpcap

RSATom:


lws-team:
Thanks a lot... unfortunately I met libpcap before I ever met libcap... it's pushed on master

Issue #1963, opened 2020-06-30
Glib glitch


lucianoiam:
Hi, and thanks for this fine piece of software.

I have implemented a WS/HTTP server (actually a plugin for the Ardour DAW) and it works flawlessly with `LWS_WITH_EXTERNAL_POLL` plus custom Glib integration code.

Since libwebsockets now comes with its own native Glib integration and external polling is deprecated, the plugin was updated to leverage `LWS_SERVER_OPTION_GLIB` on supported platforms. It works as expected but performance starts degrading over time. Specifically leaving the program running for a couple of hours leads to 100% CPU usage. This can be repeated systematically. I know this is very vague evidence sorry... but here are a couple more findings & questions that could help narrowing down the research:

- Issue repeatable on Linux and macOS

- There is a check repeated twice here, caught my attention https://github.com/warmcat/libwebsockets/blob/16cbbe2b40134cc7dd6ee16c4f356c150e3fb52d/lib/event-libs/glib/glib.c#L360

- `elops_io_glib: glib wsi source pointer is NULL` is an internal error or something the library consumer could trigger?

- Can `LWS_SERVER_OPTION_GLIB` be considered production ready?` lws-minimal-http-server-eventlib-foreign --glib` is not working on my Ubuntu box (browser fails to connect to http://localhost:7681)

- Same example app shows similar symptoms to the Ardour plugin: wsi source pointer is NULL,  Glib-CRITICAL errors on program exit. I also updated it to run for 10 hours instead of 10 seconds and boom, CPU goes 100% after perhaps one hour (see attachment)

```
./lws-minimal-http-server-eventlib-foreign  --glib
[2020/06/30 17:43:40:5057] U: LWS minimal http server eventlib + foreign loop | visit http://localhost:7681
[2020/06/30 17:43:40:5057] U:   This app creates a foreign event loop with a timer +
[2020/06/30 17:43:40:5057] U:   signalhandler, and performs a test in three phases:
[2020/06/30 17:43:40:5057] U: 
[2020/06/30 17:43:40:5057] U:   1) 5s: Runs the loop with just the timer
[2020/06/30 17:43:40:5057] U:   2) 10s: create an lws context serving on localhost:7681
[2020/06/30 17:43:40:5057] U:      using the same foreign loop.  Destroy it after 10s.
[2020/06/30 17:43:40:5057] U:   3) 5s: Run the loop again with just the timer
[2020/06/30 17:43:40:5057] U: 
[...]

PR #1979, opened 2020-07-19
Build system check for cap library and uses it without linking to it

gcsideal:
While packaging libwebsockets 4.1.0 RC1 for Debian, I found that cmake check for cap_set_flag() and uses it if available - but doesn't link to the library. This pull request corrects linking.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -799,6 +799,11 @@ if (LWS_WITH_HUBBUB)
 	list(APPEND LIB_LIST ${LIBHUBBUB_LIBRARIES} )
 endif()
 
+if (LWS_HAVE_LIBCAP)
+	find_library(LIBCAP_LIBRARIES NAMES cap)
+	list(APPEND LIB_LIST ${LIBCAP_LIBRARIES} )
+endif()
+
 
 #
 # Append the "at end" pieces to the lib list


lws-team:
Thanks a lot for the fix, it's pushed on master.  And thanks for the packaging...

RSATom:
I don't sure, but maybe with this patch there are no need to add `libcap` dependency to `libwebsockets.pc` ?
@gcsideal what do you think?

RSATom:
@gcsideal also, what do you think about enable GLIB support in libwebsockets by default in Debian/Ubuntu?

lws-team:
On the distro options, there is a `LWS_WITH_DISTRO_RECOMMENDED` that includes glib and is tested in CI on 8 platforms, eg, f32 riscv, focal and bionic x86_64

https://libwebsockets.org/git/libwebsockets/tree/CMakeLists-implied-options.txt#n76-113

It'd be great if distros could use that since it enables lws use with more complex apps from the package.

gcsideal:
> I don't sure, but maybe with this patch there are no need to add `libcap` dependency to `libwebsockets.pc` ?
> @gcsideal what do you think?

These are different things. Linking is a runtime dependency - that library should be mapped in as well to run the code. libwebsockets.pc needed for providing information via pkg-config for compile (it can give back the needed headers location as well for example) and do the actual linking of an other library or ELF binary.

gcsideal:
> @gcsideal also, what do you think about enable GLIB support in libwebsockets by default in Debian/Ubuntu?
> 
> P.S.: Sorry for offtopic...

Yup, it's off-topic here. Thus in short I have to find the balance between providing a library with several functions but with less dependencies. Question is how many users need GLIB support and how big (in number and in size) dependencies that would pull in for all the users (who probably don't need GLIB support).

gcsideal:
> On the distro options, there is a `LWS_WITH_DISTRO_RECOMMENDED` that includes glib and is tested in CI on 8 platforms, eg, f32 riscv, focal and bionic x86_64
> 
> https://libwebsockets.org/git/libwebsockets/tree/CMakeLists-implied-options.txt#n76-113
> 
> It'd be great if distros could use that since it enables lws use with more complex apps from the package.

Do you have any kind of statistics how many people / other code (would) use GLIB functionality in libwebsockets?

lws-team:
> Do you have any kind of statistics how many people / other code (would) use GLIB functionality in libwebsockets?

Not really... I only get to hear about usage when there are problems.  There's at least one other project using it for an Ardour plugin I worked with them to produce dfed5475295e552bdbf5165193eb7e40ece71a45

https://github.com/warmcat/libwebsockets/issues/1963

I can understand not wanting too many package deps but it's a bit chicken-and-egg.  I guess an answer is bind them as plugins at runtime and package them separately then, but that's not how they are atm.

gcsideal:
> > Do you have any kind of statistics how many people / other code (would) use GLIB functionality in libwebsockets?
> 
> Not really... I only get to hear about usage when there are problems. There's at least one other project using it for an Ardour plugin I worked with them to produce [dfed547](https://github.com/warmcat/libwebsockets/commit/dfed5475295e552bdbf5165193eb7e40ece71a45)
> 
> #1963
> 
> I can understand not wanting too many package deps but it's a bit chicken-and-egg. I guess an answer is bind them as plugins at runtime and package them separately then, but that's not how they are atm.

Yup, a plugin system would be nice. At the moment only one user asked for GLIB support, I guess it's not that commonly used. But prove me wrong, note me projects that use libwebsockets with GLIB bindings.

lws-team:
... nobody's trying to tell you there are many glib / lws users... I know of 2 from interacting with them.  It was only introduced a few months ago just before v4.0 release, and the people interested to use it will typically be gtk projects first, which is not a huge catchment area.  But they do exist same as there are users for all the other event libs too.  And unless they have problems, they won't come here and let me know.

Basically of the distros that carry lws, only gentoo's 'packaging' is flexible enough to reflect the build-time options in lws.  The others are missing out on a lot of stuff.

PR #1986, opened 2020-07-27
Do not serve dynamic content when not expected

8vasu:
minimal-http-server-dynamic serves dynamic content even when the mountpoint "/dyn" is not the first part of the url. That is absurd. For example, dynamic content is served upon sending urls like "/abc", "/dy", and "/dyno".

This patch fixes this by checking if the first part of the received url is our mountpoint. Now "/abc", "/dy", and "/dyno" result in 404; "/dyn", "/dyn/", and "/dyn/abc" still serve dynamic content as expected.

Signed-off-by: Soumendra Ganguly <soumendraganguly@gmail.com>


8vasu:
Additional note: static files are served as usual.

lws-team:
Thanks... I pushed it on master with some changes.

The problem is if there's only one protocol as it has been, there's nothing else for the server to do but direct it to that protocol.  I adapted the patch to add a default protocol (protocol[0]) that just calls through to the dummy handler and 404s, this should be the same behaviour your patch was working around towards.

I kept the part of your patch that shows how to get the whole original urlpath in case that's what the user wants to do with it.

The minimal examples are minimal by design, before they existed the test-server was the only thing to base off of, and that had everything plus the kitchen sink making it difficult to cut down and get started from.  So although they may seem a bit "minimal" they can always be added to once the user forks them and is off to a good start.

8vasu:
Aah... yes. My patch did seem to make the example deviate from its "minimal nature" a little bit.

I have been using libwebsockets for the past couple of months. It is wonderful; cannot thank you enough for this amazing invention.

PR #1989, opened 2020-07-28
Yunalws 1.0

hp8840-sc:
wrong try sorry.


lws-team:
I'm not sure what we're trying to do here, but v4.0-stable is whatever it is, it's not going to take patches deleting the readme.

If you have some plans to add to lws, it's better to discuss them and see if they can go in the upstream lib where I can maintain them going on, otherwise they will get left behind at whatever point you forked.  But new features would go on master, soon to be v4.1.

https://libwebsockets.org/git/libwebsockets/tree/READMEs/README.release-policy.md

PR #2001, opened 2020-08-09
Fixes to build on FreeBSD and Illumos

mikeowens:


lws-team:

I also added FreeBSD12 in Qemu to the CI that should help with compatibility issues with that in future.

PR #2028, opened 2020-08-30
Fix build with GCC < 4.4

orgads:
Prior to 4.4, -Wuninitialized could only be used with -O.


lws-team:
Thanks, it's pushed on v4.0-stable and forward-ported to master

PR #2029, opened 2020-08-30
Fix build with old glibc headers

orgads:
strnlen requires _GNU_SOURCE prior to glibc 2.10.


orgads:
Might be cleaner to just `#define _GNU_SOURCE` unconditionally. What do you say?

lws-team:
Thanks, I think it's better to solve it as you did if _GNU_SOURCE isn't popping up as required all over the place.

I pushed in on v4.0-stable and forward-ported to master.

PR #2030, opened 2020-08-30
Remove #endif without if

orgads:
Amends commit 0fa5563d18ce0b78cb682ebf6ea8a7389d725317.


lws-team:
Thanks it's pushed on v4.0-stable, it's already solved on master a while ago.

PR #2031, opened 2020-08-30
openssl: Amend recent fix for old glibc compilation

orgads:
Missed parentheses, and need to include `lws_config.h` before
testing the macro.

Also there's a chicken-and-egg situation here, because features.h is
required for testing glibc version, but including it indirectly includes
`string.h` before defining `_GNU_SOURCE`...

Fixed by defining `_GNU_SOURCE` if `strnlen` is used, regardless of glibc
version.


lws-team:
Since it's only a few minutes I squashed this into the original patch on v4.0-stable and master.

orgads:
Thank you. Sorry for the mess.

lws-team:
No worries.

PR #2046, opened 2020-09-14
unix plat: Do not break strict aliasing rules

orgads:
The type of the fields in rtentry is sockaddr, and it is
casted to sockaddr_in. Size-wise it is ok, they should both
be the same size. But casting a pointer breaks build with
optimizations with the following error:

unix-sockets.c:434: error: dereferencing pointer 'addr' does break strict-aliasing rules

Amends commit 3c95483518e8e1e203f9f75e25bd65b298445cfa.

### DIFF
--- a/lib/plat/unix/unix-sockets.c
+++ b/lib/plat/unix/unix-sockets.c
@@ -399,7 +399,6 @@ lws_plat_ifconfig_ip(const char *ifname, int fd, uint8_t *ip, uint8_t *mask_ip,
 			uint8_t *gateway_ip)
 {
 #if defined(__linux__)
-	struct sockaddr_in *addr;
 	struct sockaddr_in sin;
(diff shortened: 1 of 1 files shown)


lws-team:
Thanks, it's pushed on master.  I don't get this error on the toolchains used in CI, but the idea of -Werror is that we want to be clean for any toolchain's idea of clean using the cmake flags.

orgads:
I got it with GCC 4.4.6.

Why only master? Is 4.0 closed? What about 4.2?

orgads:
Forget it. I see that you pushed also for 4.1 :)

lws-team:
Right it is also on v4.1-stable, I pushed it last night... I forgot to close the PR at the same time or I would have listed both.  After a release I will usually only update older branches if something security-related.

PR #2049, opened 2020-09-15
smd: fix strict-aliasing violation

orgads:


lws-team:
I don't really want change the public signature of an api to char * just to work around some super-pedantic specific compiler version.

orgads:
Ok, this is a minimal change.

lws-team:
Thanks, it's pushed on v4.1-stable and master.

PR #2050, opened 2020-09-15
core-net: initialize some variables

orgads:
Amends 286cf4357a2fe2540cf38837c59f946d2b0d7263.


orgads:
You added "Subject: " in the commit message.

lws-team:
I think it's not logically needed, since hit being -1 protects it.  But pushed on master + v4.1-stable.

PR #2051, opened 2020-09-15
cmake: re-add linking with pthread

orgads:
Without it, openssl fails to link (at least on static build).


lws-team:
Can you help me understand what I should do to see this problem?  Just adding -pthread blows up other platforms.

orgads:
I'm linking with my own built openssl 1.1.x, passing `LWS_OPENSSL_INCLUDE_DIRS` and `LWS_OPENSSL_LIBRARIES`. My openssl is a static library. If you don't pass -pthread (or -lpthread), the link fails, and all the openssl-related checks in cmake fail, so it falls back to openssl 1.0 API.

orgads:
How about something like this: https://codereview.qt-project.org/c/qt-creator/qt-creator/+/313650 ?

lws-team:
The thing is lws itself doesn't have a dependency on pthread by default (depending on the platform, some other libs it does depend on may have their own dependency on pthreads and bring it in).  If you build -DLWS_WITH_THREADPOOL=1 or LWS_MAX_SMP > 1, lws will bring in pthreads.  In your scenario, you want pthreads to satisfy a dependent lib... I get it that lws is what you;re linking, but that's not really anything lws cmake can figure out.

Can you work around it by setting THREADPOOL at lws cmake?

orgads:
Replaced by https://github.com/warmcat/libwebsockets/pull/2054

PR #2052, opened 2020-09-15
cmake: use -Wuninitialized only for GCC >= 4.4

orgads:
It doesn't work with 4.3.3


lws-team:
I have added this on master and v4.1-stable, I will update the remotes later.

PR #2055, opened 2020-09-17
mingw: fix build with default options

orgads:


lws-team:
Thanks it's pushed on v4.1-stable and master

PR #2056, opened 2020-09-17
cmake: remove dead code

orgads:
Since 425da070e316de6eb1c90dca835b5bfdcf99db03 `LWS_PTHR_FLAGS` is no longer used, so just remove it.


lws-team:
Thanks it's pushed on v4.1-stable and master

PR #2061, opened 2020-09-21
cmake: change default for LWS_UNIX_SOCK on Windows

orgads:
Instead of enabling it by default and disabling after
discovering the header is not found, just disable by default.
If the user insists, the header must be available. Otherwise,
there will be fatal error as was before 3118c6659.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,8 +49,10 @@ endif()
 #   debug builds include source level debug info and extra logging
 
 set(LWS_WITH_BUNDLED_ZLIB_DEFAULT OFF)
+set(LWS_UNIX_SOCK_DEFAULT ON)
 if(WIN32)
 	set(LWS_WITH_BUNDLED_ZLIB_DEFAULT ON)
+	set(LWS_UNIX_SOCK_DEFAULT OFF)
 endif()
 
 set(LWS_ROLE_RAW 1)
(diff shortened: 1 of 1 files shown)


lws-team:
>  Otherwise, there will be fatal error as was before 3118c66.

Sorry why is it impossible to have this as the default?

orgads:
It is not impossible, but this header is not present on neither MinGW/MSVC, so it doesn't make much sense to have it as default and fail the build. 3118c66 was a workaround to fix the build, but it doesn't consider the case when the user explicitly passes `-DLWS_UNIX_SOCK=ON` and the header doesn't exist. In that case, it just ignores the user's choice while a better user experience is to fail with an appropriate message.

lws-team:
I think the larger uxp issue is that where it's possible, what happens on windows should be aligned to what happens on *nix, that means unix domain sockets ought to be built in unless the user turns them off or they can't be built.  The reason for me is Gitohashi and Sai uses UDS and at least Sai has to build and run on windows.  And any special casing I don't have to think about any more for windows is a step in the right direction.

AIUI all newish sdks from microsoft will have this so only legacy sdks and mingw will get it turned off... I think it's OK as it is.

orgads:
> AIUI all newish sdks from microsoft will have this so only legacy sdks and mingw will get it turned off... I think it's OK as it is.

I wasn't aware of that. I'll close then. Thanks.

lws-team:
Thanks for the PR

PR #2067, opened 2020-09-30
Minimal example of using smp threads with libuv foreign loops

zbroyar:


lws-team:
Thanks  it's on master with some changes.

PR #2077, opened 2020-10-08
windows: fix WSA event destruction

kcbanner:
04d947b6f731f6337dc5d2d37a0c6cdf4a8aea1d changed `pt->events` to be an array, this change fixes the destruction part to iterate the array and close all the events.


kcbanner:
Updated with null check on `wsi_from_fd` in the `_lws_plat_service_tsi`. Hit a crash here but was unable to determine exact repro conditions.

lws-team:
Thanks, they're pushed on master + v4.1-stable.

Issue #2079, opened 2020-10-09
Compiling cailed with "‘RTA_PREF’ undeclared" on old linux environment


owt5008137:
Compiler: GCC 10.2.0
System: CentOS 7(kernel 3.10)

log:

```bash
/home/owent/workspace/github/libwebsockets/lib/roles/netlink/ops-netlink.c: In function ‘rops_handle_POLLIN_netlink’:
/home/owent/workspace/github/libwebsockets/lib/roles/netlink/ops-netlink.c:161:9: error: ‘RTA_PREF’ undeclared (first use in this function); did you mean ‘RTA_PREFSRC’?
  161 |    case RTA_PREF: /* char: RFC4191 v6 router preference */
      |         ^~~~~~~~
      |         RTA_PREFSRC
/home/owent/workspace/github/libwebsockets/lib/roles/netlink/ops-netlink.c:161:9: note: each undeclared identifier is reported only once for each function it appears in
gmake[2]: *** [lib/CMakeFiles/websockets.dir/roles/netlink/ops-netlink.c.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
/home/owent/workspace/github/libwebsockets/lib/roles/netlink/ops-netlink.c: In function ‘rops_handle_POLLIN_netlink’:
/home/owent/workspace/github/libwebsockets/lib/roles/netlink/ops-netlink.c:161:9: error: ‘RTA_PREF’ undeclared (first use in this function); did you mean ‘RTA_PREFSRC’?
  161 |    case RTA_PREF: /* char: RFC4191 v6 router preference */
      |         ^~~~~~~~
      |         RTA_PREFSRC
/home/owent/workspace/github/libwebsockets/lib/roles/netlink/ops-netlink.c:161:9: note: each undeclared identifier is reported only once for each function it appears in
gmake[2]: *** [lib/CMakeFiles/websockets_shared.dir/roles/netlink/ops-netlink.c.o] Error 1
```

I found ```RTA_PREF``` was introduces with linux-4.1.x. Maybe it's required to check if it's exists.

lws-team:
Thanks for the report and patch, I pushed it on master.

PR #2080, opened 2020-10-09
Fix github issue #2079

owt5008137:
Sorry, previous commit is not enough to solve #2079 , This RP finally fix it now.


lws-team:
Thanks, I fixed LWS_HAS_ -> LWS_HAVE_ and squashed it into the other patch

PR #2082, opened 2020-10-09
Fix LWS_WITH_SUL_DEBUGGING cmake define, and warning/error for pointer comparison

kcbanner:


lws-team:
Thanks, it's pushed on main (aka master) and v4.1-stable.  I modified it slightly to also dump the callback function pointer, so you can look it up in your linker mapfile and find out which sul it is easier.

PR #2107, opened 2020-11-01
context: properly check for sys/resource.h

pinotree:
Instead of hardcoding where `sys/resource.h` is available, check for it using cmake.


lws-team:
Thanks, it's pushed on main and v4.1-stable

pinotree:
Thank you!

PR #2111, opened 2020-11-06
Update iOS.cmake for supporting iOS armv7 build

foobra:


lws-team:
Thanks, it's pushed on b4.1-stable and main.

PR #2117, opened 2020-11-16
minimal client with several active connections

zbroyar:
The client establish the given number of connections on start and keep them alive indefinitely.


lws-team:
Thanks, it's pushed on main with some modifications.

PR #2126, opened 2020-11-24
add support for sd-event as event loop

defaultbranch:
Hello @lws-team 

(This is not yet a pull request.)

We want to use libwebsockets for some application that mandates [sd-event](https://man7.org/linux/man-pages/man3/sd-event.3.html) as event loop. To that purpose, I have added a rudimentary `lib/event-libs/sdevent/sdevent.c` based your v4.1-stable branch. So far this is neither mature nor well-tested, but the software we need it for is (without libwebsockets) already in production and will remain so for the next few years.

Is there something similar (support for sd-event) in preparation yet? Would you consider adding support for sd-event in the future? What are your requirements for merging a change like this into libwebsockets?

Best regards
Christian Fuchs (cfuchs@scs.ch)

### DIFF
--- a/CMakeLists-implied-options.txt
+++ b/CMakeLists-implied-options.txt
@@ -87,6 +87,7 @@ if(LWS_WITH_DISTRO_RECOMMENDED)
 	set(LWS_WITH_LIBUV 1)				# libuv
 	set(LWS_WITH_LIBEV 1)				# libev
 	set(LWS_WITH_LIBEVENT 1)			# libevent
+	set(LWS_WITH_SDEVENT 0)				# sd-event
 	set(LWS_WITH_EVLIB_PLUGINS 1)			# event libraries created as plugins / individual packages
 	set(LWS_WITHOUT_EXTENSIONS 0)			# libz
 	set(LWS_ROLE_DBUS 1)				# dbus-related libs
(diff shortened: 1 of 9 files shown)


lws-team:
No I don't have it in progress, but before v4.1 the infrastructure for adding new event loops in lws got pretty mature, code for the event libs are very isolated, plugins are supported etc.  So maintaining more shouldn't be scary and it looks good to have this to align with systemd.

The main missing thing is ongoing testing and CI integration.  There are already some "eventlib" minimal examples, especially the "foreign" http server one

https://libwebsockets.org/git/libwebsockets/tree/minimal-examples/http-server/minimal-http-server-eventlib-foreign

that allows the user to select the event loop at runtime, lws will create a context with support for that kind of eventlib and then attach to a loop the example had already prepared with the native eventlib apis.  It tests the context operation, and then destroys it on a timer and confirms we keep going... I run it under valgrind to confirm that we completely clean up after ourselves without leaving any problems for the longer-lived "foreign" loop.  This kind of pattern is wanted if you will kind of "graft" lws into an existing loop to share it, it can be treated then as a detachable functional module cleanly if it can work reliably with the example.

I fetched your tree and it's looking more or less how it should already.

If you can add the support to minimal-http-server-eventlib-foreign so there is an unambiguous way to test it, it does not seem far away from being mergeable.  I will clean things like whitespace formatting at merge-time.

defaultbranch:
Thanks @lws-team for clarification, that sounds good. We will keep working on it!

PR #2142, opened 2020-12-11
Handle all instances of lost precision as gracefully as practical.

josharenson:
Types with increased precision are used where possible and in cases where adding increased precision wasn't practical, bounds checks were added. The behavior of the program should remain exactly the same unless side effects from implicit, lossy, type conversions were being relied upon. This is part of an effort to enable compilation with `-Werror` that will require several more patches, but this one is the largest.

### DIFF
--- a/lib/core-net/output.c
+++ b/lib/core-net/output.c
@@ -294,7 +294,7 @@ lws_ssl_capable_read_no_ssl(struct lws *wsi, unsigned char *buf, int len)
 {
 	struct lws_context *context = wsi->a.context;
 	struct lws_context_per_thread *pt = &context->pt[(int)wsi->tsi];
-	int n = 0;
+	ssize_t n = 0;
 
 	lws_stats_bump(pt, LWSSTATS_C_API_READ, 1);
 
@@ -309,7 +309,7 @@ lws_ssl_capable_read_no_ssl(struct lws *wsi, unsigned char *buf, int len)
 #endif
 		n = recv(wsi->desc.sockfd, (char *)buf, len, 0);
 
-	if (n >= 0) {
+	if (n >= 0 && n <= INT_MAX) {
 
 		if (!n && wsi->unix_skt)
 			return LWS_SSL_CAPABLE_ERROR;
@@ -325,9 +325,9 @@ lws_ssl_capable_read_no_ssl(struct lws *wsi, unsigned char *buf, int len)
 		if (wsi->a.vhost)
 			wsi->a.vhost->conn_stats.rx += n;
 #endif
-		lws_stats_bump(pt, LWSSTATS_B_READ, n);
+		lws_stats_bump(pt, LWSSTATS_B_READ, (int)n);
 
-		return n;
+		return (int)n;
 	}
 
 	if (LWS_ERRNO == LWS_EAGAIN ||
@@ -342,7 +342,7 @@ lws_ssl_capable_read_no_ssl(struct lws *wsi, unsigned char *buf, int len)
 int
 lws_ssl_capable_write_no_ssl(struct lws *wsi, unsigned char *buf, int len)
 {
-	int n = 0;
+	ssize_t n = 0;
 #if defined(LWS_PLAT_OPTEE)
 	ssize_t send(int sockfd, const void *buf, size_t len, int flags);
 #endif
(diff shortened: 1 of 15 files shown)


lws-team:
Thanks... I appreciate the patch, but can we discuss what's happening without these changes?

Eg,

```
 		n = recv(wsi->desc.sockfd, (char *)buf, len, 0);
 
-	if (n >= 0) {
+	if (n >= 0 && n <= INT_MAX) {
```

it doesn't feel like we're fighting an issue that can come up, since we are never going to be given > 32 or 64-bit INT_MAX back by recv() in the real world since len will never be more than a few KB.  It is `ssize_t`, there is a logical gap in type ranges that can be plugged, but in eg RTOS build context this is costing `.text` bloat and can never repay it.

Lws builds in CI on 23 platforms using compiler default warnings, and is tested daily in Coverity, none of those warn on these, although I see the logical problem.  How did you become aware of these, some other static analyzer?

josharenson:
My first attempt at this, I tried expanding all necessary variables to `size_t` but long story short, found it was not practical. The thought here was just to ensure that when casting from size_t to int (implicitly or explicitly) that the cast was safe. I realize this condition should never occur which is why I felt it was ok to treat `n > INT_MAX` as an error. 

I am compiling with
`-Wall -Wsign-compare -Wstrict-aliasing -Wuninitialized -Wundef -Werror`
Using these flags:
iOS - the build fails without this patch
Mac - build fails without this patch and requires another patch or two on top
Ubuntu x86/GCC and clang - build fails without this patch and requires another patch or two on top

If there is any interest at all in continuing this discussion, I can also try and arm/gcc build. I totally understand the library size constraints as I'm using this on an embedded device myself (not quite RTOS level though).

lws-team:
If you build by lws cmake (which is recommended and done in CI) you anyway get

```
-Wall -Wsign-compare -Wstrict-aliasing  -Wundef -Werror
```

https://libwebsockets.org/git/libwebsockets/tree/CMakeLists.txt#n725-732

`-Wuninitialized` is the only other one you mentioned and doesn't sound like it is responsible for complaining about these.  And all the platforms you mentioned are built in CI just fine with those flags

iOS: https://libwebsockets.org/sai/index.html?task=878180e29a9508fa5c7808b1955582df2a99e176f1a9fd7a698fa8b2f749f752
OSX (llvm): https://libwebsockets.org/sai/index.html?task=878180e29a9508fa5c7808b1955582df8065de500a6d9007af38be276c70dea0
Ubuntu (gcc): https://libwebsockets.org/sai/index.html?task=878180e29a9508fa5c7808b1955582dfc29f95915ecda01909fad9f85393ba4a

In total there are 330 builds done in CI over 23 platforms, with those compiler flags and no problems.

Is your platform and its toolchain unusual somehow?  Lws builds on 32-bit int / 64-bit long, 32/32 and 64/64, and even platforms where uint32_t is long, without any warnings or errors at all... all builds are done with -Wall / -Werror.

Or perhaps there is another flag enabling additional warnings hiding in your build arrangements somewhere?

I can imagine studying how to clean these out but first we need to figure out how to reproduce the problem.

lws-team:
... if you paste the actual errors, it should mention in there the flag responsible.

lws-team:
It seems for whatever reason, for you -Wall includes -Wconversion, which provokes this.

I dunno what your larger build process is, but a good way to integrate lws is build it using cmake and it produces a libwebsockets.a (or .so if that's what you want).  You can workaround this then IIUI by adding -Wno-conversion temporarily in the lws CMakeLists.txt where I linked to earlier, so basically a get-out for this just for lws for now.

Then import the lib into your larger build at link-time, which can continue to use -Wconversion.

Tomorrow I will try building this myself with it and look at each problem in turn to see if we can / should solve them (since they may affect api arg types).

josharenson:
Ah, that makes sense. I am building this via cmake as part of a larger project and the current work around is to do exactly what you've described (might be just using cmake to set -Wno-conversion for this specific target). I'll try getting some more verbose logs to confirm your suspicions. I'll keep following this thread to see what your findings are and please let me know if I can get you any more logs. At the end of the day, my goal is to build for Linux (x86 and arm),  ioS, Android, Mac, and Ubuntu so happy to provide local logs from any of those. I'm also aware that this library is well tested and deployed all over the place, so if I have to disable a single warning I'm not terribly worried.

lws-team:
OK... I will go through it tomorrow, if I don't fix all of them directly then I'll fix the ones that make sense and add -Wno-conversion to make sure we override any parent build flag set for it.

lws-team:
I have pushed a patch on main atm that enables the -Wconversion on all the unix platforms, and fixes 98%+ of the warnings.

Because this is a huge patch (18K lines) and we are coming up to release 4.2, I plan to complete it, remove it briefly, issue 4.2 and then return the patch on to main to continue development post 4.2.

Please give it a try.

lws-team:
This solves all the warnings found in all 331 CI builds now same as before the patch.

lws-team:
... if any problems with it remaining, let me know.

PR #2185, opened 2021-01-27
pipe: avoid busy loop in windows pipe event

luterrytw:
Hi Sir:
I found there is a busy loop in windows library. It cause by the pipe write without pipe read.
I have a workaround by add recv() in rops_handle_POLLIN_pipe() function.
Is it a bug or I should change by function call to avoid this problem?

Thanks.


lws-team:
Thanks, it's pushed on main.

PR #2204, opened 2021-02-18
pollfd: Interrupt WSAPoll while event changes

luterrytw:
Dear Sir:
Since the windows busy loop already fix in 3bf3b540ad8f386095d8fe95ac9b1fe78afcf8e1, it cause another issue.
Because WSAPoll() is blocked so the event change will be ignore.
I don't know if you are still developing on it, but I commit one line to avoid this problem, and it is work for me.

Thanks.

### DIFF
--- a/lib/plat/windows/windows-sockets.c
+++ b/lib/plat/windows/windows-sockets.c
@@ -238,6 +238,7 @@ lws_plat_change_pollfd(struct lws_context *context, struct lws *wsi,
 		       struct lws_pollfd *pfd)
 {
 	//struct lws_context_per_thread *pt = &context->pt[(int)wsi->tsi];
+	lws_plat_pipe_signal(wsi->a.context, wsi->tsi);
 
 	return 0;
 }


lws-team:
Thanks... but what is actually happening in your code at the time you care about that?

It's not safe to call lws apis from another thread, except `lws_cancel_service()`.

luterrytw:
Hi Andy:
It's not safe to call lws apis from another thread, except lws_cancel_service(). ==> If so, I think this change is not good.
I have two thread, one will while(1) run the lws_service(), another will listen user input and call lws_callback_on_writable() to trigger data sending to websocket server. And I found the lws_callback_on_writable() is not work on Windows but linux is ok.

Do you have any suggestion to resolve this condition?

Thanks.

lws-team:
`lws_cancel_service()` is the way that's safe for other threads to call... it puts a byte in the platform-specific pipe or UDP connection that is dedicated for this purpose... that causes an event directly waking the service loop from the poll wait.  This works safely on all mixtures of platforms and event libraries, with any number of threads, the same.

Lws realizes that is why it woke and sends a protocol message `LWS_CALLBACK_EVENT_WAIT_CANCELLED` in the lws service thread context.

So the usual pattern is the other thread prepares something (on, eg, a linked-list) in memory shared by the threads that's protected by a mutex, describing what happened or what it wants.  Then it calls `lws_cancel_service()`, the lws thread wakes and in your `LWS_CALLBACK_EVENT_WAIT_CANCELLED` handler, it locks the mutex for the shared region and figures out what to do from that, eg, ask for the `lws_callback_on_writable()` from there, in the lws thread already.  Because it's handling it from the lws thread, it can do any api call safely, like create new client connections or whatever.

luterrytw:
Hi Andy:
Thanks for your help, now I know the design pattern more detail.
It, lws_cancel_service(), works for me now without this change.
Please help me close this request, sorry to bother you about this.

Thanks.

lws-team:
No worries I appreciate the patch, especially for windows stuff.

PR #2220, opened 2021-03-01
lws_struct: fix implicit schema recognition

sp-andwei:
The current attempt to implicitly match the schema is only working if the first child of the schema struct matches the first encountered key in the JSON object. The code never actually iterates through all entries but uses only the first (m times).


lws-team:
Thanks for the patch, nice find... it looks like the intention was to have child++ at the end of the loop, but I forgot it.

Your way also fixes it just fine, it's pushed on main and v4.1-stable.

PR #2228, opened 2021-03-05
windows-resolv: fix dns resolution

xCuri0:
Before this patch all I got was ```cant get dns servers``` on my laptop running Windows 10 1909,


lws-team:
Thanks... but I don't think this can be right... it's basically saying the ERROR_BUFFER_OVERFLOW is success.

If I look at https://docs.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-getnetworkparams there is an example that allocates the FIXED_INFO, if you align your change with that, does it still work for you?

xCuri0:
@lws-team the example uses ERROR_BUFFER_OVERFLOW too.

xCuri0:
I think I should change it to also work if NO_ERROR is returned

lws-team:
Thanks, I adapted your patch to have the one-time resize logic in the ms example, tested it on a win 10 VM with SYS_ASYNC_DNS, and pushed it on main.

https://libwebsockets.org/git/libwebsockets/commit?id=8dbe8507a3dc7e78b90b875e83729f6add7d7e88

PR #2233, opened 2021-03-12
windows: load certificates from certificate store

xCuri0:
code adapted from curl. only loads the certificates once as it takes ~25ms on my system.

works without issues but i guess it will need improvements before can be merged

### DIFF
--- a/lib/plat/windows/private-lib-plat-windows.h
+++ b/lib/plat/windows/private-lib-plat-windows.h
@@ -68,6 +68,10 @@
 #include <afunix.h>
 #endif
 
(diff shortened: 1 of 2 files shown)


lws-team:
Thanks... yes some of the lws integration bits need adapting, but that doesn't change the value of getting the patch with the guts working.

Curl license is also MIT so there's no problem basing on some of their code, so long as we attribute it as you did here.

I modified the integration part and adapted the style, then tested on win 10, my openssl has its own trust bundle configured, this reported 25 (presumably, additional) certs from the platform store.

It's pushed on main.

Issue #2244, opened 2021-03-22
MT-Safe functions getgrgid_r, getgrnam_r, getpwuid_r, getpwnam_r


pavelxdd:
Hello. Consider changing functions `getgrgid` to `getgrgid_r`, `getgrnam` to `getgrnam_r`, `getpwuid` to `getpwuid_r` and `getpwnam` to `getpwnam_r` for multithread safety.

The current functions access the static memory internally which can cause races in multi-threaded applications.

lws-team:
This is on main now

https://libwebsockets.org/git/libwebsockets/commit?id=1503bc72846c13c34f3afa8cedb594d9cc66da8f

PR #2247, opened 2021-03-24
unix: fix usage of getpwnam_r and friends

pavelxdd:
These functions can return 0 code but still store NULL in result, if no matching group or username found.
Also the buffer of 64 size could be too small to store all string fields in result.

#2244


lws-team:
> These functions can return 0 code but still store NULL in result, if no matching group or username found.

Ugh... thanks for the help and patch, it's pushed.

PR #2249, opened 2021-03-24
Add client support for reusing sessions (OpenSSL)

catalinr-m:
This is an attempt to allow clients to reuse SSL sessions. Potentially reusable sessions will be kept in a LWS cache.
This can be enabled by setting the new `client_ssl_reuse_sessions` flag in `lws_context_creation_info` to `1`.
The precompiler condition for using the new functionality seemed to be
```
#if defined(LWS_WITH_NETWORK) && defined(LWS_WITH_TLS) && defined(LWS_WITH_CLIENT)
```

A session is identified by a string composed of the "\<IP\>:\<port\>". For an existing sessions identified in this way, after attempting to reuse it, `SSL_session_reused()` did return `1` for it.

### DIFF
--- a/include/libwebsockets/lws-context-vhost.h
+++ b/include/libwebsockets/lws-context-vhost.h
@@ -514,6 +514,9 @@ struct lws_context_creation_info {
 	/**< VHOST: Client SSL context init: length of client_ssl_key_mem in
 	 * bytes */
 
+	int client_ssl_reuse_sessions;
+	/**< VHOST: Client SSL context init: 0 = default, do not use session
+	 * cache; 1 = cache and attempt to reuse sessions */
 #endif
 
 #if !defined(LWS_WITH_MBEDTLS)
--- a/lib/core/context.c
+++ b/lib/core/context.c
@@ -230,6 +230,44 @@ lws_context_creation_completion_cb(lws_sorted_usec_list_t *sul)
 }
 #endif /* WITH_SYS_STATE */
 
+#if defined(LWS_WITH_TLS) && defined(LWS_WITH_CLIENT)
+int
+lws_context_cache_session(struct lws_context* ctx, SSL_SESSION* sess, const char* name)
+{
+	/* update an existing item */
+	for (struct lws_session_cache* s = ctx->client_ssl_session_cache; s; s = s->next) {
+		if (!strcmp(name, s->name)) {
+		SSL_SESSION_free(s->session);
+		s->session = sess;
+		lwsl_info("Session cache: replaced session %s\n", name);
+		return 1;
+		}
+	}
+
+	/* add a new item */
+	struct lws_session_cache* cache_item = malloc(sizeof(struct lws_session_cache));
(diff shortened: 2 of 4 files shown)


lws-team:
Thanks for the patch.  There are a few problems but I have pushed a modified version that solves most of them

 - new work has to be done on main branch, we can't add new abi to stable branch.  Once it's straight you can probably backport it if you need to, v4.1 and main are not hugely different and this is mainly adding code.

 - use a vhost flag in the context / vhost creation struct to enable session reuse

 - use a cmake option to enable it `LWS_WITH_TLS_SESSIONS`

  - bring all the related code into its own file `./lib/tls/openssl/openssl-sessions.c` and build that based on the cmake flag

 - the patch doesn't do session expiry at all.  I added it, it's just set to a fixed 300s atm

 - there is no test I can run to confirm it works, both for this work and in CI so we can know it didn't break.  There needs to be some way to use a minimal example to get a go / no-go test so we get alerted to problems.  It can be a new flag on an existing one or a new example.

If you could take a look on main and check / fix the changes, and address the last point, it would be great.  Because of the last point, I don't know if I broke it atm.

catalinr-m:
Thanks for the feedback and updates! (For the reference, it's 933ff036a81f5724fc64f8848892d8f0f0410153.)

I'll get `main` branch and test with it, and get back with any required changes if needed.
I'd gladly add a test, but I don't really know how it should work. Connecting to a URL a few times and checking that the same session was reused might sound good enough, except
- what URL would that be, so that it's stable enough to not break tests from unrelated issues?
- a server can decide to refuse to resume a session, and create a new one, so simply _not_ having a session reused would not mean that it was "not working".

Any suggestions for the test would be great. TIA!

lws-team:
It just needs to test it under some constrained circumstance, eg, to a specific server controlled by the tester, like libwebsockets.org.  It should not be fixed in the code but done so you can pass it in via commandline parameter.

AFAIK the server part of session resumption "is the default"

SSL_SESS_CACHE_SERVER

    Server sessions are added to the session cache. When a client proposes a session to be reused, the server looks for the corresponding session in (first) the internal session cache (unless SSL_SESS_CACHE_NO_INTERNAL_LOOKUP is set), then (second) in the external cache if available. If the session is found, the server will try to reuse the session. **This is the default.**

https://www.openssl.org/docs/man1.1.0/man3/SSL_CTX_set_session_cache_mode.html

if so, eg, connecting to libwebsockets.org:443 (or whatever you want), disconnecting then reconnecting with LWS_WITH_TLS_SESSIONS enabled should be enough to have it reuse the session.  If it isn't clear already, there'll need to be a way to check if the session was reused successfully, that can end up being used as the exit value from the test process.

catalinr-m:
Thanks, I'll try that way.
Yes, testing that a session was reused can be done using `SSL_session_reused(SSL *ssl)`. That's the one I used for testing my initial changes too.

lws-team:
Closing this and continuing discussion on the v3 patch PR

PR #2251, opened 2021-03-25
cmake: check if _GNU_SOURCE is already defined

pavelxdd:
`_GNU_SOURCE` can already be defined in `CMAKE_C_FLAGS`, so instead of redefining it (which will cause an error), check it beforehand.


lws-team:
Thanks, it's pushed on main

PR #2255, opened 2021-03-29
TLS session cache update

catalinr-m:
I'll explain my changes:
- Removed:
  - Management of "used" state of cached sessions.
  - Session timeout mechanism managed by LWS; this was inaccurate, as custom session timeout must be set for each session, to be visible by the server etc.
- Added support for session timeout, using  SSL_SESSION existing features. No other special handling is needed for it, as described [here](https://security.stackexchange.com/a/182636):
  > After the timeout the session will no longer be accepted for reuse. Note that this does not mean that a current session will be shutdown, only that a new handshake with this session will not result in a session reuse.

(Still pending: test code)


lws-team:
Thanks, I squashed this into the earlier patch on main (and fixed the author...).  It'll be great if we can confirm it's working and staying working with the test.

catalinr-m:
Thanks!
I've fetched `main` and the lib builds fine, but tests fail with
```
[ 98%] Building C object test-apps/CMakeFiles/test-client.dir/test-client.c.o
[100%] Linking C executable ../bin/libwebsockets-test-client
/usr/bin/ld: lib/libssl.a(ssl_key_share.o): in function `bssl::(anonymous namespace)::X25519KeyShare::~X25519KeyShare()':
ssl_key_share.cc:(.text._ZN4bssl12_GLOBAL__N_114X25519KeyShareD0Ev+0x5): undefined reference to `operator delete(void*)'
lib/libssl.a(ssl_key_share.o): in function `bssl::(anonymous namespace)::CECPQ2KeyShare::~CECPQ2KeyShare()':
ssl_key_share.cc:(.text._ZN4bssl12_GLOBAL__N_114CECPQ2KeyShareD0Ev+0x5): undefined reference to `operator delete(void*)'
/usr/bin/ld: lib/libssl.a(ssl_key_share.o): in function `bssl::(anonymous namespace)::ECKeyShare::~ECKeyShare()':
ssl_key_share.cc:(.text._ZN4bssl12_GLOBAL__N_110ECKeyShareD0Ev+0x34): undefined reference to `operator delete(void*)'
/usr/bin/ld: lib/libssl.a(ssl_key_share.o):(.data.rel.ro._ZTVN4bssl11SSLKeyShareE[_ZTVN4bssl11SSLKeyShareE]+0x20): undefined reference to `__cxa_pure_virtual'
/usr/bin/ld: lib/libssl.a(ssl_key_share.o):(.data.rel.ro._ZTVN4bssl11SSLKeyShareE[_ZTVN4bssl11SSLKeyShareE]+0x28): undefined reference to `__cxa_pure_virtual'
/usr/bin/ld: lib/libssl.a(ssl_key_share.o):(.data.rel.ro._ZTVN4bssl11SSLKeyShareE[_ZTVN4bssl11SSLKeyShareE]+0x38): undefined reference to `__cxa_pure_virtual'
collect2: error: ld returned 1 exit status
make[2]: *** [test-apps/CMakeFiles/test-client.dir/build.make:89: bin/libwebsockets-test-client] Error 1
make[1]: *** [CMakeFiles/Makefile2:1392: test-apps/CMakeFiles/test-client.dir/all] Error 2
make: *** [Makefile:163: all] Error 2
```
This is with ` -DLWS_WITHOUT_SERVER=ON`. When including the server, there are even more errors.

Am I doing something wrong?

lws-team:
Not sure where your C++ tests come from, but it's not lws afaik.

catalinr-m:
The tests can only come from LSW repo. Maybe the problem is with `libssl` that is being used...

lws-team:
It looks like you built boringssl specifically with some dangling c++ std library guts hanging out of it.

Working with boringssl is difficult due to it trying to exactly look like, and conflicting with openssl on the same box.

catalinr-m:
Yes, it is `borringssl` (the one used by `webrtc`). With or without `-DLWS_WITH_BORINGSSL=ON` makes no difference, which is even more confusing..

lws-team:
Boringssl manifests itself with exactly <openssl/ssl.h>, and libssl.so / libcrypto.so, the WITH_BORINGSSL is mainly for UI indication.  That is why it's difficult to work with when the build machine has its own copy of openssl going around.

Here's some notes about how I built it and bound lws to it

https://libwebsockets.org/git/libwebsockets/tree/READMEs/README.lsq.md?h=lsq

Check with ldd what your boringssl libs are binding to or not being able to bind to.  On my local build of it described above, its bindings look like

```
ldd /projects/boringssl/build/ssl/libssl.so 
	linux-vdso.so.1 (0x00007ffc353dd000)
	libcrypto.so => /projects/boringssl/build/crypto/libcrypto.so (0x00007fdec1819000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fdec17d4000)
	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007fdec15b5000)
	libm.so.6 => /lib64/libm.so.6 (0x00007fdec1471000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007fdec1456000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fdec1287000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fdec1ac6000)
```

it seems your build is unsatisfied in the C++ department and expecting the lws link to satisfy it... it won't because lws is just C.  So you must make sure it binds to what it needs at boringssl build time.

catalinr-m:
Thanks!
The build here already uses `-DLWS_OPENSSL_INCLUDE_DIRS` and `-DLWS_OPENSSL_LIBRARIES`, which helped building the lib. I guess no one ever built the test app or examples...
Anyway, no luck with boringssl, no matter which cmake options and linker flags I used, but I eventually managed to build LWS and test app using system openssl. Well, with one small tweak: unless I defined `-DLWS_WITHOUT_SERVER=ON`, build failed with the following error:
```
libwebsockets/lib/tls/openssl/openssl-server.c: In function ‘lws_tls_server_certs_load’:
libwebsockets/lib/tls/openssl/openssl-server.c:454:38: error: dereferencing pointer to incomplete type ‘lws_tls_ctx’ {aka ‘struct ssl_ctx_st’}
  454 |  x = sk_X509_value(vhost->tls.ssl_ctx->extra_certs, 0);
      |                                      ^~
```

lws-team:
> I guess no one ever built the test app or examples...

What are you talking about?  The CI builds and runs them in hundreds of diffetent scenarios.

https://libwebsockets.org/sai/index.html

...including against boringssl, where it runs dozens of minimal examples via ctest, eg

https://libwebsockets.org/sai/index.html?task=8aef0c6e6cca865751638140ec6028ca95c234f540ab72967daefac635169046

If I met a situation where it worked before and then my build broke with mysterious errors around a selfbuilt dependent lib, I would think about deleting my CMakeCache.txt and redoing that from scratch before making up other explanations.

catalinr-m:
> What are you talking about?

Sorry, it got out the wrong way. I meant no one _here_ ever built them

PR #2256, opened 2021-03-31
Add option to test TLS session reuse

catalinr-m:
Running
```
libwebsockets-test-client https://warmcat.com --port=443 --ssl --multi-test --tls-session-reuse
```
prints
```
[2021/03/31 02:30:48:7136] N: __lws_tls_session_create: new 46.105.127.147:443 ttl 7200s
[2021/03/31 02:30:48:7137] N: __lws_tls_session_create: update 46.105.127.147:443 ttl 7200s
```
```
[2021/03/31 02:30:48:7244] N: __lws_tls_session_create: update 46.105.127.147:443 ttl 7200s
[2021/03/31 02:30:48:7245] N: __lws_tls_session_create: update 46.105.127.147:443 ttl 7200s
```
```
[2021/03/31 02:30:48:7278] N: __lws_tls_session_create: update 46.105.127.147:443 ttl 7200s
[2021/03/31 02:30:48:7279] N: __lws_tls_session_create: update 46.105.127.147:443 ttl 7200s
```


lws-team:
FYI you can update your PR with new patches rather than a new PR for each one.

Thanks for the patch with the test, I modified it here to have a user-visible api as I suggested to find out if the connection was reused.

I also changed the cmake option to default-on, and the logic for the vhost option to "give it to disable session reuse", so it will become on by default for openssl users.

I think the logs you show above don't show session reuse working.  They just show that we keep updating the SESSION in the cache entry for the endpoint.

There's a corner case that there may be more than one vhost / SSL_CTX going to the same endpoint with different client certs.  So I added the vhost name to the string identifying the endpoint.

I spent some time trying to understand the code now it's nearly there and putting valgrind on it, it seems we leak some openssl pieces atm, whether that's my changes or always like that don't know yet.  I added comments so we can pick the code up easier later.

There's also a limitation on using this that SESSIONs don't exist and so can't be used until one negotiation to the endpoint completes.  That means broadside tests like in default lws-minimal-http-client-multi cannot take advantage of it, since they create the SSL bio for the others before the first can complete.  With the -s (staggered) option that delays the other connects it can use it.

I noticed the situation in the PR you just pushed about endless session extensions and added a comment there.

I suggest you rebase on _temp branch and we can discuss these things.

lws-team:
(this is _temp on libwebsockets.org)

https://libwebsockets.org/git/libwebsockets/log?h=_temp

lws-team:
... I also added a settable limit for the max client cache size, defaults to 10, and MRU-sort the linked-list containing the cache, trimming the LRU entry if we run out of space.

catalinr-m:
> I think the logs you show above don't show session reuse working. They just show that we keep updating the SESSION

Oh, right. I'll wait for your changes then.

> there may be more than one vhost / SSL_CTX going to the same endpoint

Ok, those should not mix, indeed. The aim of the initial patch was to make a cache per context, so there would not have been any clashes. But any type of cache you find more appropriate should be perfectly fine.

> I suggest you rebase on _temp branch

Sorry, I'm  a bit lost. Where do I fetch `_temp` branch from? Or should I just rewind `main` to the last commit showed there and add my changes on top of that? And which changes should I put in a new branch rebased on `_temp`?

lws-team:
The cache needs to be per SSL_CTX at openssl... the SSL_CTX in lws lives in struct lws_vhost.  It doesn't make any difference in the case you only use the default vhost, it's created along with the context for you by default using the same info struct.  But it will allow more complicated scenarios to work right.

```
git fetch https://libwebsockets.org/repo/libwebsockets +_temp:_temp
git reset --hard _temp
```

Just send a patch containing your changes on top of that, I will squash them back into the single topic patch for this, currently

https://libwebsockets.org/git/libwebsockets/commit?id=e232d42d2648ed88d2a2d26bf323c7500200949a

PR #2258, opened 2021-03-31
TLS session resume support

catalinr-m:
The initial code was wrong. By extending the lifetime before reusing a session (thus not giving the chance to the server to see that it expired and refuse it), it would most likely never timeout.


catalinr-m:
Redone, based on `_temp` branch.

catalinr-m:
When a client requests a resume on an expired session ("original time" + ttl < "now"), it should be the server's duty to notice that it expired, do a full handshake, and return a new session. Then, in the new-session-callback, the old one will be replaced by the new one in cache.

lws-team:
OK.  Then I'll continue to look at it later.

catalinr-m:
I'd like to add one more thing: support for persistent sessions.

For dumping a session, the API could be something like this:
```
int
lws_tls_dump_session(struct lws *wsi, unsigned char **pp);
```
which would use `int i2d_SSL_SESSION(SSL_SESSION *in, unsigned char **pp)`, and mimic its functionality.


For loading, maybe 2 new members could be added to `lws_client_connect_info`:
```
unsigned char *sessAsn1;
long sessAsn1Len;
```
Then `lws_client_connect_via_info()` could use `SSL_SESSION *d2i_SSL_SESSION(SSL_SESSION **a, const unsigned char **pp, long length)` to create a new session, call `SSL_SESSION_set_time()` on it, add it to cache, and set it for the new connection.

Do you think something like this could work?

lws-team:
I guess but on different platforms, storing and retrieving the blobs might be very different... eventually it will presumably exist also on mbedtls so then there, eg, might be some key value store into nvram and not even a filesystem.

How about pass a callback function that receives the blob pointer and length and does the platform-specific storage.  Also, I am not sure what you want it for, but might it not more be dumping and retrieving the session cache than dumping individual wsi SESSIONSs?

catalinr-m:
Right, I didn't think about different SSL back-ends.

Maybe there could be a new lws structure, containing a SSL back-end id (i.e. 1 for OpenSSL, 2 for mbedtls etc.), and other elements specific to the back-end (`unsigned char *sessAsn1` and `long sessAsn1Len` in case of OpenSSL). LWS could dump and load that kind of info.
The need is for trying known connections later, after an app shutdown and restart. But it might be needed to only resume selected sessions, and not the whole cache.
LWS would not need to manage session dumps in any way (deal with files or otherwise), just provide one into a memory buffer upon request, and be able to load it later.

lws-team:
I think it's OK that the serializations of the sessions are opaque and private to a particular tls backend, they don't have to try to interoperate.  This d2i.. api should just emit an opaque blob of a particular length and pass that to a callback to store it.  On a server it might use open(), write() or whatever in the callback, on another platform use something else.  But the lws part just passes blobs in and out and wires them up to the tls library apis.

Isn't it a bit gnarly to know which wsi to pick, and easier to operate on the cache as a whole each way?

catalinr-m:
`i2d_SSL_SESSION()` and `d2i_SSL_SESSION()` are already part of OpenSSL, in case it looked like they were part of the API proposal. And they do work with an opaque, openssl-specific, blobs. The question is what's the best way to wrap them in LWS.

I agree that it's more cumbersome to deal with individual sessions, but this was requested to me and, if it doesn't get to hairy, I'd like to try this approach first.

lws-team:
I mean that you call one of the ..._vh() (to save or restore a whole session cache) or ..._wsi() (to do the indvidual ones you want) apis.  And that causes the cb to be called either once or multiple times accordingly, with tag passed in along with the serialized session blob in or out.

The cb can use posix open or read or write and filename based off the tag, or it can do a key-value blob set or get using the tag, it's up to user code to decide depending on the platform.

catalinr-m:
Ok, I think I understand now.
Personally I wouldn't go for functions that change what they do based on the value of one of their parameters. It also risks having some other parameters remain unused, depending on what the function does...

catalinr-m:
How about the following:
```
struct session_dump
{
	int impl_type; // 1: OpenSSL; 2: mbedtls; etc
	void *blob;
	size_t blob_len;
	char *tag; // probably only ip:port
};


typedef void (*lws_tls_sess_dump_cb_t)(struct session_dump*); // callback for dumping a session
typedef struct session_dump* (*lws_tls_sess_load_cb_t)();     // callback for loading a session

int lws_tls_dump_sessions(struct lws_context *ctx, lws_tls_sess_dump_cb_t cb); // dump all sessions, until passing NULL to  cb
int lws_tls_load_sessions(struct lws_context *ctx, lws_tls_sess_load_cb_t cb); // load all sessions, until receiving NULL from cb

int lws_tls_dump_session(struct lws *wsi, lws_tls_sess_dump_cb_t cb); // dump one session
```

The names can be changed. These are only to make them more self-explanatory.

catalinr-m:
Actually, the last one can be
```
struct session_dump* lws_tls_dump_session(struct lws *wsi); // dump one session
```

lws-team:
OK... although lws has been uneven about it in the past, and we're stuck with those, it's better if the api names are like lws_tls_sessions_xxx().

We should pass the context into the callback as a parameter, because on some platforms the actual storage apis will go through lws_system_ops, that is reachable by the context.

If you're planning on storing these in a generic filesystem for your usecase, it might make sense to export canned example filesystem read and write callback implementations that the user can just wire up as the callbacks given to the apis if that what he wants.  Maybe add a void *opaque on the _dump() and _load() apis that is passed through the session_dump struct to the callback, you can define it for your callbacks to contain the const char * base dir things are stored and recovered from.

Just to clarify (api is the same) tag might also be the vhost name or whatever when dumping all the sessions, so we don't bring back authenticated sessions into the wrong vhost.


I have stitched the user api for checking if the session was reused into the minimal-http-client-multi example and check it matches what's expected depending on the options.  So I'll be trying that in CI shortly.  I also toned down the logs to info.

catalinr-m:
> If you're planning on storing these in a generic filesystem for your usecase, it might make sense to export canned example filesystem read and write callback implementations

The intention was to _not_ deal with file system in any way. If the user wants to store a blob in a file, or insert it in a db, or send it to the cloud, they can choose what to do with the _in-memory_ blob later.

> tag might also be the vhost name or whatever when dumping all the sessions, so we don't bring back authenticated sessions into the wrong vhost.

I understand that concern. I don't know how a vhost is identified so I probably cannot accurately evaluate these implication. But can vhost naming be controlled by the user? What if an app dumps the cache for a vhost with a certain name, but later, when loading, the existing vhost has a different name simply because that's how it was generated, or "default" was already taken when it was created?

> I have stitched the user api

I just checked and I couldn't see new changes in https://libwebsockets.org/git/libwebsockets/log?h=_temp
Are they in a different place or should I wait for the changes to show up there?

Thanks again!

lws-team:
They're in main branch atm.

Yes it should be built to not have any dependency on the filesystem, that's good and that's how it is described atm.  However, if in fact you are going to store them in the filesystem yourself, you can export (but not hook up to anything yourself) generic filesystem save / restore callback implementations.

If the user wants generic filesystem storage for them, he can provide those exported callback names in his api call instead of having to rewrite his own (and control the base dir using the opaque ptr interpreted by the callback).  If he doesn't want that, he doesn't reference them at all and provides his own custom callback to do the save and restore action.

catalinr-m:
I couldn't find them in `main` either, maybe they're not pushed yet? Anyway, I'll wait.

lws-team:
Currently this

https://libwebsockets.org/git/libwebsockets/commit?id=f897f4f998b05face8e0ac85813c092fb50fa334

Too unstable to put in CI yet  ---> libwebsockets.org _temp branch
Testing in CI ---> libwebsockets.org  main branch
Passed CI --> libwebsockets.org main branch + github main branch

catalinr-m:
Can you please tell me what is the next step? Should I still wait for something or start implementing a way for dumping and loading sessions?
Dumping a single session should be straight forward, but for working with the cache, I'd need to get the `lws_vhost` starting from `lws_context`, and I can't find a way to get that.

Also, it's getting more and more unclear to me what is the API that you'd accept for this - function signatures, arguments etc.
Can you please help me clear this out?

lws-team:
>  I'd need to get the lws_vhost starting from lws_context, and I can't find a way to get that.

I did some more reading on this today, it's not recommended to have very long-lived session IDs or store them on "relatively insecure environments", like eg unencrypted filesystem.

```
   F.1.4
...
   An upper limit of 24 hours is suggested for
   session ID lifetimes, since an attacker who obtains a master_secret
   may be able to impersonate the compromised party until the
   corresponding session ID is retired.  Applications that may be run in
   relatively insecure environments should not write session IDs to
   stable storage.
```

https://tools.ietf.org/html/rfc5246

For that reason, what I was assuming about dumping the whole cache being the natural way is probably not a good idea.  We should drop that and just allow cherrypicking wsi sessions if that's what the user really wants to do.  So that simplifies it a lot.

> Can you please help me clear this out?

Here's what I think is aligned with what you had and the comments, it doesn't have to be exactly this as you actually do it you might need to change it.  But this should be the starting reference point we can hopefully agree on.

```
struct lws_tls_session_dump
{
	int impl_type; // 1: OpenSSL; 2: mbedtls; etc
	void *blob;
	size_t blob_len;
	const char *tag;  // "vhost-IP:port" tag suitable for use as filename
        void *opaque;  // has meaning to callback, eg, a const char * base_dir for filesystem stg
};

/* common callback signature for both the load and save callbacks */
typedef int (*lws_tls_sess_cb_t)(struct lws_context *cx, struct session_dump *info);

/* prepares a struct session_dump on the stack with 1) tag from wsi, 2) blob / blob_len pointing
 * to the wsi session info, 3) .opaque from "opaque" parameter, then calls cb_save with it to store
 * it "somewhere" */
int
lws_tls_session_dump_save(struct lws *wsi, lws_tls_sess_cb_t cb_save, void *opaque);

/* prepares a struct session_dump on the stack with 1) tag from wsi, 2) .opaque from
 * "opaque" parameter, then calls load_cb with it.  That allocates on heap for .blob and
 * sets that and .blob_len to the loaded data.  If the callback returns 0 then we instantiate
 * the SSL_SESSION and bind the wsi to it, then lws_free() the struct's .blob.
 */
int
lws_tls_session_dump_load(struct lws *wsi, lws_tls_sess_cb_t cb_load, void *opaque);
```

Because of the test in the RFC quoted above, I think it's not going to be recommended to
store these in normal files for typical client.  So no need to provide the generic filesystem-based
callbacks I mentioned earlier.

lws-team:
Some other points... you removed the time-to-live scheduling code that I added before in the current version... if there are a lot of transient client connections, isn't it just going to fill up with dead sessions?  Am I missing the point something else is going to come and weed out outdated cache entries in our code, or we need to add the ttl stuff back in?

Also I have implemented the mbedtls version of this, which looks like it is working.  But there are no equivalent ttl related apis exported and nothing obvious in the public mbedtls session struct telling the ttl.  So it seems even more stark that we have to do our own scheduled weeding of the client cache...

catalinr-m:
> isn't it just going to fill up with dead sessions?

Yes, it will.
One idea was that as long as it's a cache, it might just behave like the most common ones -- web browser, swap partition -- they all contain data very useful in the short term, but which eventually become garbage. Since there is a limit for all that, the garbage will be automatically thrown overboard when newer and more useful data is loaded. There should be no need for extra management...

 > there are no equivalent ttl related apis exported and nothing obvious in the public mbedtls session struct telling the ttl

...oh, then again, maybe it should be? I had no idea about that, and if it is indeed necessary, then I also have nothing to say against it. It might be an unexpected optimization for OpenSSL too.

catalinr-m:
Getting back to dumping and loading: if it's settled for 1 by 1, isn't it simpler without callbacks? Just one function for dumping, and one for loading?
Another thing that puzzles me: a wsi is created when calling `lws_client_connect_via_info()`. How could that wsi reuse a session that would need a wsi to be loaded into the cache?

lws-team:
Well, I added it back in for mbedtls.

lws works down on constricted devices like ESP32, which typically has about 200KB heap for ALL user code.  This logic about squatting heap uselessly might fly on a server, but it's just completely wrong on these small devices.

On openssl, rather than an "optimization" I think it's actually broken without the ttl scheduling, we hold a ref on it forever as it stands, and even when the ttl is up, openssl cannot destroy it when we sit there forever holding the ref.

lws-team:
> isn't it simpler without callbacks?

Your usecase does not need the callbacks. But others will.  So you need to think also from the perspective of other library users not just what is in front of you today.  If it seems too difficult to agree on how to do it just leave it and I will do it.

> Another thing that puzzles me: a wsi is created when calling lws_client_connect_via_info(). How could that wsi reuse a session that would need a wsi to be loaded into the cache?

No idea what you are talking about.

catalinr-m:
> On openssl, rather than an "optimization" I think it's actually broken without the ttl scheduling, we hold a ref on it forever as it stands, and even when the ttl is up, openssl cannot destroy it when we sit there forever holding the ref.

Let's say that for a cached session, Session1, there should be 2 references: one held by openssl, the other by the cache. When Session1 has expired, at some point openssl will decrement its ref count. The cache will still have one ref. When attempting to reuse Session1, the server will refuse it, and create Session2. Session2 will replace Session1 in the cache, thus the ref count of Session1 is decremented for the second time.
If Session1 is never reused, it will not be released until 10 other new sessions have been created and they push it beyond the cache cap.
If I've missed something and openssl case is still broken, or if it will cause a waste of memory with a significant impact somewhere else, then sure, let's add it back in there and fix things.

> So you need to think also from the perspective of other library users not just what is in front of you today. If it seems too difficult to agree on how to do it just leave it and I will do it.

I'm really trying to see other perspectives, it's definitely not about that. I'm only trying to understand how things are supposed to work, outside my usecase, otherwise it's even less likely I'd do things right. Ideally we'd agree on everything, but ultimately we don't have to. If you say they have to be like _this_, so be it.

> No idea what you are talking about.
```
struct lws_context *wscontext;
struct lws_context_creation_info info;
memset( &info, 0, sizeof(info) );
/* init info members */
wscontext = lws_create_context(&info);

struct lws *wsi;
struct lws_client_connect_info i;
memset(&i, 0, sizeof(i));
i.context = wscontext;
/* init other members */
wsi = lws_client_connect_via_info(&i);
```
Where should a session be loaded in the example above, so that `wsi` will reuse it?

lws-team:
> I'm really trying to see other perspectives, it's definitely not about that.

... don't strain yourself.

I'll provide the implementation and then we can discuss once it exists.

> Where should a session be loaded in the example above, so that wsi will reuse it?

Wsi have a lot of callbacks in the protocol, you can catch it there if you want to get the tag from the connection details in the wsi just in time, that's the idea of the wsi in the load api.

If you want to preload the cache, rather than the wsi you'd synthesize the tag and have a load api based on that.

That's another problem as it is, the tags should use the hostname not the peer IP.

lws-team:
Here is what I ended up with, you can get it from libwebsockets.org main branch

https://libwebsockets.org/git/libwebsockets/commit?id=60607832e0f6fcf3c12f4366a0f2b0f28d3ac3ab

It's untested and the load / store parts likely need debugging, but it's all there.  AIUI you just need to hook up load and store callbacks (once it works...)

I also unified the tag creation code in its own file so there's one definitive way to do that, and modified it to use the client hostname instead of the IP.

lws-team:
I debugged and tested the save and load code for openssl now.

I added test code on minimal-http-client-multi, so it will load and save sessions in /tmp, and confirmed that works (all 8 connections are reusing the saved session after the first run).   I found when doing this that you were right about the awkwardness of using the wsi to talk about the session... the target of the load and save is really the vhost cache, it's not really tied up with any specific wsi.  So I ended up making both load and save use discrete strings to describe the session they want to load and save, without reference to any wsi, and that worked well.

I also added a README explaining the ins and outs

https://libwebsockets.org/git/libwebsockets/tree/READMEs/README.tls-sessions.md?h=main

One quirk you may need to delete your CMakeCache.txt or hand-edit it to set `LWS_HAVE_SSL_SESSION_up_ref:INTERNAL=1` just one time.  It's because SSL_SESSION_up_ref is not a function but a "symbol" in openssl, looking for it with cmake as a function won't find it.  I fixed that but if you used last night's build, it will set it in CMakeCache.txt that it isn't there and keep believing that.

Mbedtls does not seem to have helpers for load and save

https://forums.mbed.com/t/serialize-mbedtls-ssl-session-for-use-in-rfc-5077-session-resumption/3958/3

it seems you're supposed to walk the struct by hand, I think it can wait until someone wants it, so the load and save is only provided for openssl atm.

I think a bit more to do in CI to make the file operations code work on windows, then I will look again at the cache lifecycle problem on openssl.

catalinr-m:
> https://libwebsockets.org/git/libwebsockets/commit?id=60607832e0f6fcf3c12f4366a0f2b0f28d3ac3ab

Can you please tell me how I can get to that code? I've tried `main` and `_temp` and the changes are not there. Is there a different branch, or should I get it as a patch and apply it on top of one of them?

lws-team:
Github is just a mirror for us.  Things get pushed to libwebsockets.org infrastructure first.  The CI happens there, if a tree passes the CI then the github repo can get updated with it.  So for this kind of development work, stuff is often not passing CI yet and only exists at libwebsockets.org repo.

It's all public, you can teleport a branch from there to your local git like this

```
$ git fetch https://libwebsockets.org/repo/libwebsockets +main:lws-main && git reset --hard lws-main
```

will bring over libwebsockets.org "main" branch to your local git as "lws-main", then the second part force your checked out tree to that.  This fetch / reset flow is also how you can rebase on the latest tree in one step.

catalinr-m:
Hi, first, thank you very much for working on this! Now there is support for saving and loading cached sessions.

For the record, the current API is
```
struct lws_tls_session_dump
{
	char			tag[LWS_SESSION_TAG_LEN];
	void			*blob;
        void			*opaque;
	size_t			blob_len;
};

typedef int (*lws_tls_sess_cb_t)(struct lws_context *cx,
				 struct lws_tls_session_dump *info);

int
lws_tls_session_dump_save(struct lws_vhost *vh, const char *host, uint16_t port,
			  lws_tls_sess_cb_t cb_save, void *opq);

int
lws_tls_session_dump_load(struct lws_vhost *vh, const char *host, uint16_t port,
			  lws_tls_sess_cb_t cb_load, void *opq);
```

Using `save` and `load` in the sample is
```
lws_tls_session_dump_save(lws_get_vhost_by_name(context, "default"),
					i.host, (uint16_t)i.port,
					sess_save_cb, "/tmp")
```

```
lws_tls_session_dump_load(lws_get_vhost_by_name(context, "default"),
				  i.host, (uint16_t)i.port, sess_load_cb, "/tmp")
```

---

### A few notes and personal thoughts, biased by the use case I'm having here:

1) In the current form, `tag` is pretty useless; if it needs to be broken into components in the user code, then maybe it's worth storing them separately in the first place.

2) `opaque` is rather some "user data", which could be left out of the struct.

3) > you were right about the awkwardness of using the wsi to talk about the session... the target of the load and save is really the vhost cache

Actually, I was only referring to _load_.
   _save_ on the other hand, IMHO would make sense for a wsi, at least for OpenSSL, because every wsi would wrap a `SSL*` which would have a `SSL_SESSION*` which may no longer be cached (i.e. still active, but has been thrown out of cache by newer sessions). Why not allow saving the session of an active connection, whether it was cached or not?

4) The idea of having `lws_tls_session_dump`, was also to use it in the save/load API, as opposed to the separate arguments.

---

### How about the following changes:
```
struct lws_tls_session_dump
{
	char			vhost[LWS_VHOST_LEN];
	char			host[LWS_HOST_LEN];
	uint16_t		port;
	void			*blob;
	size_t			blob_len;
};

typedef int (*lws_tls_sess_cb_t)(struct lws_context *cx,
				 struct lws_tls_session_dump *info, void *opq);

/* save the session, only if it was cached */
int
lws_tls_session_dump_save(struct lws_context *cx, struct lws_tls_session_dump* save,
			  lws_tls_sess_cb_t cb_save, void *opq);

/* save the session of the given connection */
int
lws_tls_session_dump_save2(struct lws *wsi, struct lws_tls_session_dump* save,
			  lws_tls_sess_cb_t cb_save, void *opq);

/* load the session into cache */
lws_tls_session_dump_load(lstruct lws_context *cx,
				  struct lws_tls_session_dump* load, sess_load_cb, void *opq);
```
and used in user code like

#### save
```
lws_tls_session_dump save_target;

/*
init members:
save_target->vhost;  // suggested flexibility: when empty, use "default" internally
save_target->host;
save_target->port;
*/

lws_tls_session_dump_save(initial_cx, &save_target, user_save_cb, "");
  /*
    this will
     - fill in
       save_target->blob;
       save_target->blob_len;
     - call user_save_cb, if it was not NULL;
     - allow getting the blob with and without a callback and opq.
  */
```

#### save 2
```
lws_tls_session_dump save_target;

lws_tls_session_dump_save2(wsi, &save_target, user_save_cb, "");
  /*
    this will
     - fill in all members of save_target
       suggested flexibility: leave save_target->vhost empty for "default"
     - call user_save_cb, if it was not NULL;
     - allow getting the blob with and without a callback and opq.
  */
```

#### load
```
lws_tls_session_dump load_target;

/*
init members:
load_target->vhost;  // suggested flexibility: when empty, use "default" internally
load_target->host;
load_target->port;
load_target->blob;       // optional !
load_target->blob_len;   // optional !
*/

lws_tls_session_dump_load(initial_cx, &load_target, user_load_cb, "");
  /*
    this will
     - call user_load_cb, if it was not NULL;
     - load the blob if it was not NULL; thus allow loading the blob with and without a callback and opq.
  */
```

lws-team:
I am grateful for your initial patch that got the ball rolling, I look at these kind of patches as ways of saving my time.  Of course discussing them and integrating them takes my time, but usually I am still ahead.  But this is starting to go into the red timewise.

There are still real issues to solve like the ttl management on openssl case.  But bikeshedding details on the apis isn't a real issue if they are already able to do what is needed.  Especially if it's just going to irk you no matter what if it doesn't look like what you would have written.  I'm going to have to maintain it along with the rest of lws, it needs to look and act like lws code.

> In the current form, tag is pretty useless; if it needs to be broken into components in the user code, then maybe it's worth storing them separately in the first place.

"Pretty useless"?... no, tag is doing a critical job so the session's applicability and scope can be stored and searched for.  The store-reload can't work without it.  Storing the pieces of the tuple of vhost-host-port in pieces doesn't help anything, the tuple pieces have to be serialized together in the blob metadata, eg, filepath anyway.  The code specifying them separately is a convenience, how they are stored is a separate issue.  AFAICS changing this doesn't buy anyone anything.

> opaque is rather some "user data", which could be left out of the struct.

In the rest of lws, structs contain members called "opaque" or variations on it to indicate that this data is just passed through to other user callback code without interpretation or modification by lws.  So this is already well-named and follows the established strategy.  Is that even worth our time to discuss if it should be in the struct or passed as a parameter on the load / save call?  It's the same functionality.

> save on the other hand, IMHO would make sense for a wsi, at least for OpenSSL, because every wsi would 
...
> The idea of having lws_tls_session_dump, was also to use it in the save/load API, as opposed to the separate arguments.

Yes in the rest of lws, wsi is the object that has and conceals all the details, so this isn't unaligned and it's why that was my first plan.  The problem I had was that I read in TLS 1.3, the issue by the server of the session ID is disjoint from the negotiation, and not only is not guaranteed to have occurred by the time of the SSL_connect() completed, but there is no upper bound on when it is issued at the connection.  That's why openssl has this callback scheme to indicate a new SSL_SESSION arrived rather than added an api to query the session at some fixed point.  So trying to query the wsi to get the session info... there is no good time / lws callback to tell people to use.  mbedtls approach is leave all the pieces on the floor and let the user deal with it, it does not have the same callback scheme to notify user code the session arrived, so, eg, no compatible way to add a callback indicating it happened.

For that reason I concluded actually we are loading and saving to and from the lws session cache, not to or from wsi.  And that led to the api approach that is there today.

Can I gently suggest if you have some time, the thorn that is sticking out of this still is the cache lifecycle management for openssl?  You can refer to the mbedtls version to see how the lws apis are applied there.

catalinr-m:
> There are still real issues to solve like the ttl management on openssl case.

I've force-pushed this branch with a new commit that reinstates internal management of expired sessions.

> in TLS 1.3, the issue by the server of the session ID is disjoint from the negotiation, and not only is not guaranteed to have occurred by the time of the SSL_connect() completed, but there is no upper bound on when it is issued at the connection

My idea on this was that a _user_ can mark a connection as "ready for backup" after, say, it started tranfering a file. But never mind, it's dropped.
I guess the same goes for making saving and loading work with and without NULL callbacks, and everything else.

> this is starting to go into the red timewise

Ok, I understand it's time for me to give up. Done.

lws-team:
Thanks, I added the missing `lws_sul_cancel()` and it's pushed on main.

Thanks again for your patches.

PR #2259, opened 2021-04-02
Fixes setting the length of the last will and testament's message

GaryChristiansen42:
It was previously using the length of the topic for the message length. It
would break if the topic and message were not the same length.


lws-team:
Ouch... thanks a lot, it's pushed on main and v4.2-rc1-stable

GaryChristiansen42:
Thanks! Glad that my first pull request went well.

PR #2281, opened 2021-04-20
cmake: also set LWS_BUILD_HASH to unknown if no git is found

sp-andwei:
If there isn't any reason for a build to fail on a system with no git, I'd suggest setting the LWS_BUILD_HASH to unknown in a case where no git found.


lws-team:

PR #2283, opened 2021-04-27
TLS session cache fixes

catalinr-m:


lws-team:
Thanks, they're pushed on both main + v4,.2-stable... I dropped the 4th patch and just changed the affected lines to tabs in the earlier patch.

lws-team:
There are two issues found in Sai, some platforms have openssl where the SSL_session_reused() api returns a long.  Casting the return to int got around the type warning that's promoted to an error.

The second problem is that with http2, the wsi is a multiplexed stream separate from the nwsi that has the tcp connection and the tls tunnel, so it crashed since the child wsi has no wsi->tls.ssl.   The old code gets the nwsi from the wsi (returning the wsi itself if it's not h2 or is the nwsi already), I changed the last patch to follow what was already there before and it is passing CI now.

catalinr-m:
Thanks for the fixes! The implementation here has http2 disabled..

catalinr-m:
> with http2, the wsi is a multiplexed stream separate from the nwsi that has the tcp connection and the tls tunnel, so it crashed since the child wsi has no wsi->tls.ssl.

Then maybe the same thing would be needed in `lws_tls_reuse_session()`, where it does
```
SSL_set_session(wsi->tls.ssl, ts->session);
```
?

lws-team:
No, he's only called from `lws_ssl_client_bio_create()`, the related wsi is definitely doing the nwsi job.  muxed wsi that are children of the nwsi don't know what tls is even, they will never try to create a tls lib bio.

PR #2285, opened 2021-04-28
Add borringssl synthetic cb for caching TLS sessions

catalinr-m:
This is not exactly what you proposed, because I tried adding the new `sul`, and in its callback I coud not find a way of getting the `SSL`.

I've added this relatively simpler change, because `SSL_SESSION_is_resumable()` always returned true in the synthetic callback, so each new session could be added to the cache directly from that cb. It does not sound completely bullet proof, but so far it worked every time.
It can also be improved later.


lws-team:
This is pushed on main as discussed on the ml

PR #2286, opened 2021-04-29
Allow OPENSSL_INCLUDE_DIR to be empty

thefallentree:
on some system the include dir is empty, and the detection will miss a lot of functions and fail to compile.


lws-team:

Issue #2287, opened 2021-04-29
Wrong logic of processing attr in RTM_NEWLINK message (Linux with LWS_WITH_NETLINK ON)


millikk:
Hello everybody! I found wrong logic in the function `rops_handle_POLLIN_netlink()` ([lib/roles/netlink/ops-netlink.c:162 on the master branch](https://github.com/warmcat/libwebsockets/blob/main/lib/roles/netlink/ops-netlink.c#L162)). According to the [manual](https://man7.org/linux/man-pages/man7/rtnetlink.7.html) the `RTM_NEWLINK` message can't have `RTA_SRC` attribute, so here, in fact, we are dealing with the `IFLA_BROADCAST` attribute (the values of `RTA_SRC` and `IFLA_BROADCAST` are the same). Thus the data of the attribute `RTA_DATA(attribute)` is a 6-bytes broadcast hardware address - `FF:FF:FF:FF:FF:FF`, but in the code we see a wrong cast to the `lws_sockaddr46 *` pointer. Moreover, the next condition performs checking if the address family is valid by comparing it to `0xffff` value - so here this condition will always be true and then all the routes for the current interface will be removed.

The above problem has appeared in my project. I wrote a client application which makes connections repeatedly in case of errors. I started it (without server) and after a couple of seconds I turned off the network interface (`ifconfig eth0 down`) and got error messages like `lws_sort_dns: 192.168.70.1 has no route out`. Then I turned on the network interface and started the server but the error messages didn't stop...Then I looked at the source code, start debugging and after some time I figured out the problem.

So, my suggestion is to remove the whole `case RTA_SRC` block because it's useless: we can't get the type of the address family (AF_INET, AF_INET6, etc..) here - it's not available for `RTM_NEWLINK` messages. I did this fix in my project and the problem went away: the client successfully connected to the server. And if you don't mind I'd like to create a pull request.

Thanks

lws-team:
Thanks for the detailed report... it sounds wrong as you describe.  Yes please send a PR, I think the broadcast mac  addr isn't interesting atm, so pulling out the code for that case is probably right.

millikk:
I created a PR, but seems like I can't add reviewers there

lws-team:
Thanks for the patch, it's pushed on main + v4.2-stable.

millikk:
Thank you too, it was fascinating :)

PR #2288, opened 2021-04-29
Netlink: fix of wrong processing of the attr from the RTM_NEWLINK message

millikk:
**Brief**
There is wrong logic of processing of the `RTA_SRC` attr from the `RTM_NEWLINK` message. In fact it is the `IFLA_BROADCAST` attr which provides 6-bytes broadcast hardware address. So here we have a wrong pointer cast and the whole logic under the `case RTA_SRC` block is useless because there's no possibility to get network address family type (AF_INET, AF_INET6, etc..) from the `RTM_NEWLINK` message.

For more details see the corresponding [issue#2287](https://github.com/warmcat/libwebsockets/issues/2287)


lws-team:
Thanks, it's pushed on main and v4.2-stable

PR #2304, opened 2021-05-20
fix the overflow and the control of private key when using mbedtls.

ycyang1229:
When using mbedtls with cert_filepath, it may causes memory corruption. Not sure the "buf[amount++] = '\0';" is necessary so I add one byte in alloc_file(). 
And add the implementation of private_key_filepath. 
Please give me some feedback since I am not familiar with this codebase. Just tried the private-key, cert, and cacert, and found that it did not work yesterday.
thanks.


lws-team:
Thanks a lot for the patch, I pushed it on main and v4.2-stable.

ycyang1229:
> Thanks a lot for the patch, I pushed it on main and v4.2-stable.

Thanks for the immediate response. :)

PR #2305, opened 2021-05-20
mbedtls: fix client certificate validation

orlupo:
use MBEDTLS_SSL_VERIFY_REQUIRED authentication mode if client certificate validation is required.

### DIFF
--- a/lib/tls/mbedtls/wrapper/platform/ssl_pm.c
+++ b/lib/tls/mbedtls/wrapper/platform/ssl_pm.c
@@ -234,7 +234,7 @@ static int ssl_pm_reload_crt(SSL *ssl)
     struct x509_pm *crt_pm = (struct x509_pm *)ssl->cert->x509->x509_pm;
 
     if (ssl->verify_mode == SSL_VERIFY_PEER)
-        mode = MBEDTLS_SSL_VERIFY_OPTIONAL;
+        mode = MBEDTLS_SSL_VERIFY_REQUIRED;
     else if (ssl->verify_mode == SSL_VERIFY_FAIL_IF_NO_PEER_CERT)
         mode = MBEDTLS_SSL_VERIFY_OPTIONAL;
     else if (ssl->verify_mode == SSL_VERIFY_CLIENT_ONCE)


lws-team:
I reproduced the situation with serving + mbedtls + client certificate validation, and pushed a modified version of your patch on main, it solved it here, please check it works OK for you

https://libwebsockets.org/git/libwebsockets/commit?id=18ed5dd3a0473ab0bc4a124b74dd56b539a3d303

orlupo:
> I reproduced the situation with serving + mbedtls + client certificate validation, and pushed a modified version of your patch on main, it solved it here, please check it works OK for you
> 
> https://libwebsockets.org/git/libwebsockets/commit?id=18ed5dd3a0473ab0bc4a124b74dd56b539a3d303

I tried testing the patch. I got segmentation fault when accessing the server

lws-team:
can you use vlagrind or gdb to look at where it blows?  It passes the CI, and running the test server by hand here.

orlupo:
> can you use vlagrind or gdb to look at where it blows? It passes the CI, and running the test server by hand here.

Thread 2 received signal SIGSEGV, Segmentation fault.
[Switching to Thread 2634.2635]
0x000b6cf8 in src_aligned ()
(gdb) bt
#0  0x000b6cf8 in src_aligned ()
#1  0x002ddbc8 in lws_tls_server_conn_alpn ()
#2  0x002e690c in lws_server_socket_service_ssl ()
#3  0x002f2690 in rops_handle_POLLIN_h1 ()
#4  0x002e7cd4 in lws_service_fd_tsi ()
#5  0x002e7e38 in _lws_plat_service_forced_tsi ()
#6  0x002e8128 in _lws_plat_service_tsi ()
#7  0x002e7da0 in lws_service ()

lws-team:
It's a bit difficult without symbols but it looks like this is unrelated build issue?

`lws_tls_server_conn_alpn()` is like this

```
int
lws_tls_server_conn_alpn(struct lws *wsi)
{
#if defined(LWS_WITH_MBEDTLS) || (defined(OPENSSL_VERSION_NUMBER) && \
                                  OPENSSL_VERSION_NUMBER >= 0x10002000L)
        const unsigned char *name = NULL;
        char cstr[10];
        unsigned len;

        if (!wsi->tls.ssl)
                return 0;

        SSL_get0_alpn_selected(wsi->tls.ssl, &name, &len);
        if (!len) {
                lwsl_info("no ALPN upgrade\n");
                return 0;
        }

        if (len > sizeof(cstr) - 1)
                len = sizeof(cstr) - 1;

        memcpy(cstr, name, len);
        cstr[len] = '\0';

        lwsl_info("negotiated '%s' using ALPN\n", cstr);
        wsi->tls.use_ssl |= LCCSCF_USE_SSL;

        return lws_role_call_alpn_negotiated(wsi, (const char *)cstr);
#endif // OPENSSL_VERSION_NUMBER >= 0x10002000L

        return 0;
}
```

https://libwebsockets.org/git/libwebsockets/tree/lib/tls/tls.c?h=main#n110-142

... I guess `src_aligned` is related to the memcpy implementation.  The source is on the stack

```
void SSL_get0_alpn_selected(const SSL *ssl, const unsigned char **data,
                            unsigned int *len)
{
#if defined(LWS_HAVE_mbedtls_ssl_get_alpn_protocol)
	const char *alp = mbedtls_ssl_get_alpn_protocol(&((struct ssl_pm *)(ssl->ssl_pm))->ssl);

	*data = (const unsigned char *)alp;
	if (alp)
		*len = (unsigned int)strlen(alp);
	else
		*len = 0;
#endif
}
```

it seems OK for alignment to me, and handles alpn being NULL OK too.

orlupo:
@lws-team I think the reason is "len" isn't initialized to 0 in "lws_tls_server_conn_alpn"

lws-team:
If the bug goes away with that, then I will stand corrected :-) but in SSL_get0_alpn_selected() it should be set to zero.

Is it possible that `LWS_HAVE_mbedtls_ssl_get_alpn_protocol` is not set in CMakeCache.txt somehow?

Especially when building cross, you have to take care that cmake is seeing the cross-world headers and libs, not what's on your build machine.

orlupo:
I set LWS_HAVE_mbedtls_ssl_get_alpn_protocol and it fixed the segfault.
Also the client certificate is verified as expected

lws-team:
Great, thanks a lot for the patch and the extra verification.

orlupo:
@lws-team When to do you expect to have tagged version?

PR #2313, opened 2021-05-26
netlink: fix UB on RTM_NEWADDR/RTM_DELADDR

boussole:
this patch makes ubsan (-fsanitize-undefined-trap-on-error) happy.

let's look at the following scenario:
1. netlink sends RTM_NEWADDR/RTM_DELADDR message
2. rops_handle_POLLIN_netlink() handles message attributes as RTA_*
3. rops_handle_POLLIN_netlink() finds RTA_PRIORITY attribute and
handles it as u32 value, but it's IFA_CACHEINFO attribute in reality
and payload holds another type
4. ubsan throws SIGTRAP

so, don't handle RTM_NEWADDR/RTM_DELADDR attributes as RTA_*


lws-team:
Thanks, it's pushed on main and v4.2-stable.  Netlink was a lot more difficult than expected because the canonical parser library has an incompatible license, it had to be done from scratch.

Issue #2315, opened 2021-05-29
check_function_exists doesn‘t work on macOS


caobug:
Sorry, I'm here to interrupt again.

When I cross-compile macOS via https://github.com/leetal/ios-cmake, the "check_function_exists" report can find some symbols, but these symbols do not exist. maybe should use "CheckSymbolExists" instead.

https://github.com/libsdl-org/SDL/issues/3467
https://github.com/libgit2/libgit2/pull/5898

https://cmake.org/cmake/help/v3.8/module/CheckFunctionExists.html

```
-- Looking for SSL_CTX_load_verify_file
-- Looking for SSL_CTX_load_verify_file - found
-- Looking for SSL_CTX_load_verify_dir
-- Looking for SSL_CTX_load_verify_dir - found
```
```
-- Looking for clock_gettime
-- Looking for clock_gettime - found
```

For example, I changed `CHECK_FUNCTION_EXISTS(clock_gettime LWS_HAVE_CLOCK_GETTIME)` to `CHECK_FUNCTION_EXISTS(abcdefg LWS_HAVE_CLOCK_GETTIME)`, it still prompts that it can be found
```
-- Looking for abcdefg
-- Looking for abcdefg - found
```

lolgear:
@caobug
Could you check that the solution that I added in [PR](https://github.com/libgit2/libgit2/pull/5898) works?

```cmake
enable_warnings(error)
check_symbol_exists(clock_gettime "sys/header.h" HAVE_CLOCK_TIME_FUNCTION)
IF (HAVE_CLOCK_TIME_FUNCTION)
  # do whatever you want
endif()
disable_warnings(error)
message(FATAL_ERROR)
```

lws-team:
@lolgear enable_warnings() doesn't seem to be a cmake thing, I guess it's a macro from elsewhere in libgit2.

The canonical way to do what you're doing is `set(CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS} " -Werror")`

I don't think it will help CHECK_FUNCTION_EXISTS, because a) it seems to not generate any warning to promote, and b) it generates a non-meaningful link-time warning.

```
Determining if the function clock_gettime exists passed with the following output:
Change Dir: /Users/sai/jobs/0-0.0/libwebsockets/build/CMakeFiles/CMakeTmp

[...]

PR #2316, opened 2021-05-30
Fixed can't find clock_gettime in macOS

caobug:
which was introduced in macOS 10.12.

https://github.com/warmcat/libwebsockets/issues/2315


lws-team:
Thanks, this is pushed on v4.2-stable and main now.

PR #2320, opened 2021-06-05
core-net: Remove unused variable

kraj:
variable 'n' is being set but it is not used anywhere, latest clang is
able to detect this and flags it

Fixes

lib/core-net/route.c:41:6: error: variable 'n' set but not used [-Werror,-Wunused-but-set-variable]
|         int n = 0;
|             ^


lws-team:
Thanks, this is pushed on v4.2-stable and main now.

PR #2323, opened 2021-06-14
CHECK_FUNCTION_EXISTS not working on macos

kekxv:
`CHECK_FUNCTION_EXISTS` not working


lws-team:

PR #2355, opened 2021-07-21
gcc 4.4 base64-decode.c complains char to int conversion

hassanctech:
gcc 4.4 complains with the following:
lib/misc/base64-decode.c: In function ‘_lws_b64_encode_string’:
lib/misc/base64-decode.c:76: error: conversion to ‘char’ from ‘int’ may alter its value
/lib/misc/base64-decode.c:77: error: conversion to ‘char’ from ‘int’ may alter its value

After the explicit cast the warning is gone.


lws-team:
Thanks for the patch, it's pushed on main and v4.2-stable.

PR #2357, opened 2021-07-22
correctly restrict file serving on http2 to max frame size

thefallentree:
previous fix was buggy and result in sending incomplete files.


lws-team:
Thanks, it's pushed on v4.2-stable with some minor changes and squashed into the first patch on main.

thefallentree:
i found that you have applied the patch wrong:  the `poss = poss - ..` line needs to be outside the #ifdef LWS_ROLE_H2  block

lws-team:
Isn't it that subtracting n should be there, but subtracting LWS_H2_FRAME_HEADER_LENGTH should be inside the block?

thefallentree:
> Isn't it that subtracting n should be there, but subtracting LWS_H2_FRAME_HEADER_LENGTH should be inside the block?

I havn't looked closely the code down below, but the existing code always reserve LWS_H2_FRAME_HEADER_LENGTH no matter what, so the correct amount to read ("poss")  is always min(pt_bufer_size, MAX_FRAME_SIZE) - n - LWS_H2_FRAME_HEADER_LENGTH

Issue #2391, opened 2021-08-23
How to iterate custom headers?


astone2014:
How do I iterate over all custom headers? Or even iterate all headers (custom or not).

I see it's fairly simple to extract custom headers
```c
lws_hdr_custom_length(wsi, "dnt:", 4);
```

But how do I iterate custom headers **if I don't know their name ahead of time**?

I'm trying to use LWS dynamically so that users can create webservers that serve and respond to any sort of custom protocol.

It seems like `lws_hdr_custom_length` is itterating to find the name in question. Is it possilbe to callback for each or to create some list?
```c
	ll = wsi->http.ah->unk_ll_head;
	while (ll) {
		if (ll >= wsi->http.ah->data_length)
			return -1;
		if (nlen == lws_ser_ru16be(
			(uint8_t *)&wsi->http.ah->data[ll + UHO_NLEN]) &&
		    !strncmp(name, &wsi->http.ah->data[ll + UHO_NAME], (unsigned int)nlen)) {
			n = lws_ser_ru16be(
				(uint8_t *)&wsi->http.ah->data[ll + UHO_VLEN]);
			if (n + 1 > len)
				return -1;
			strncpy(dst, &wsi->http.ah->data[ll + UHO_NAME + (unsigned int)nlen], (unsigned int)n);
			dst[n] = '\0';

			return n;
		}
		ll = lws_ser_ru32be((uint8_t *)&wsi->http.ah->data[ll + UHO_LL]);
	}
```

I tried to do this myself but it seems like everything I need access to is private in user context.

astone2014:
```0x0000026228743404  00 07 00 01 00 00 00 24 76 61 6c 75 65 31 3a 31  .......$value1:1
0x0000026228743414  00 07 00 01 00 00 00 34 76 61 6c 75 65 32 3a 32  .......4value2:2
0x0000026228743424  00 07 00 01 00 00 00 00 76 61 6c 75 65 33 3a 33  ........value3:3
0x0000026228743434  50 6f 73 74 6d 61 6e 52 75 6e 74 69 6d 65 2f 37  PostmanRuntime/7
0x0000026228743444  2e 32 38 2e 33 00 2a 2f 2a 00 6c 6f 63 61 6c 68  .28.3.*/*.localh
0x0000026228743454  6f 73 74 3a 37 36 38 31 00 67 7a 69 70 2c 20 64  ost:7681.gzip, d
[...]

PR #2393, opened 2021-08-25
Add new function to iterate custom header names

astone2014:
New function `lws_hdr_custom_name_foreach` takes callback function that is called for each custom header name. Use `lws_hdr_custom_copy` to get values of headers.

Closes How to iterate custom headers? #2391


lws-team:
Thanks for the patch Austen, I squashed the topic patches into one and it was already in good shape.  I noticed the return for success was accidentally also -1 and changed that to zero, and modified the callback type to end with _t: it's pushed on main

https://libwebsockets.org/git/libwebsockets/commit?id=059f5e1a61693b326948b31e894361c85b915eaa

Issue #2396, opened 2021-08-30
Doesn't work on some Windows


caobug:
Hello,  LWS-4.2.1 does not work on **some** Windows, however 4.0.16 is ok.

I captured in the log: Unable to connect

That `lws_client_connect_info.address` only configures an IP address, not a domain name.

More LWS uses are as follows:

https://github.com/warmcat/libwebsockets/issues/2176#issuecomment-907566069

lws-team:
Can you paste the log?  What's the ipv6 situation?

lws-team:
... and ASYNC_DNS situation...

caobug:
Hello, I only log the error level. The IP address is IPV4, and ASYNC_DNS is not configured.

I need to republish the software to my friend’s computer for specific error messages.

lws-team:
Yes need better logs.  It's also interesting if LWS_IPV6 enabled at cmake.

caobug:
1.1.1.1 is a fake IP
```
[2021-08-30 22:57:40] D __lws_vhost_unbind_wsi: vh default: count_bound_wsi 1
[2021-08-30 22:57:40] D __lws_free_wsi: [wsicli|34|WS/h1/1.1.1.1], tsi fds count 2
[2021-08-30 22:57:40] D  -- [wsicli|34|WS/h1/1.1.1.1] (1) 33.983ms
[2021-08-30 22:57:40] D _lws_change_pollfd: [wsicli|35|WS/h1/1.1.1.1]: fd 4448 events 784 -> 768
[2021-08-30 22:57:40] D elops_io_uv: [wsicli|35|WS/h1/1.1.1.1]: 10
[2021-08-30 22:57:40] D lws_client_connect_check: connect check take as FAILED: errno 10014
[2021-08-30 22:57:40] D lws_client_connect_3_connect: getsockopt check: conn fail: errno 10014
[2021-08-30 22:57:40] D elops_io_uv: [wsicli|35|WS/h1/1.1.1.1]: -2147483637
[2021-08-30 22:57:40] E Unable to connect
```

caobug:
After I delete the connect related code, it can work normally. I don’t know what your consideration is to not use SO_ERROR on Windows?

```
static lcccr_t
lws_client_connect_check(struct lws *wsi)
{
	int en = 0;
	int e;
	socklen_t sl = sizeof(e);

	(void)en;

	/*
	 * This resets SO_ERROR after reading it.  If there's an error
[...]

PR #2400, opened 2021-09-01
fix: Windows sock status check.

caobug:
Details: https://github.com/warmcat/libwebsockets/issues/2396


lws-team:
Thanks for the patch but this didn't work out for me as discussed on the issue.  It seems something else is behind the original problem.

PR #2421, opened 2021-09-22
Fix _lws_plat_file_seek_cur to update pos in fd

martlaak-gmail:
* needed to make fops_zip to work in windows


lws-team:
Thanks, it's pushed on main and v4.2-stable

PR #2427, opened 2021-09-28
mbedtls: fix closing socket early

caobug:


lws-team:
Sorry I forgot to update here, this is actually on main already

https://libwebsockets.org/git/libwebsockets/commit?id=7eb5656b2ed416973489beb30a89f5e757451353

PR #2432, opened 2021-10-01
examples: fix sizeof lws_tls_cert_info_results in test_jwt_RS256

pavelxdd:


lws-team:
Sorry I did not get the email notification about this and did not see it until thismorning.  Thanks for the patch, it's pushed on v4.3-stable and main.

PR #2434, opened 2021-10-06
event loop: fix monitor cannot be close.

caobug:
after we initialize the event driver, we need to call the related API to close, for example: uv_poll_init-uv_close(poll).

when we call lws_client_connect_via_info to connect to a wrong IP address, we will receive LWS_CALLBACK_CLIENT_CONNECTION_ERROR and LWS_CALLBACK_WSI_DESTROY notifications, uv_poll_init has been called, but uv_close is not called.

https://github.com/warmcat/libwebsockets/blob/40a87eaf2a1da918061e4c9bc4e3c707f152e491/lib/core-net/client/connect3.c#L375

https://github.com/warmcat/libwebsockets/blob/40a87eaf2a1da918061e4c9bc4e3c707f152e491/lib/core-net/client/connect3.c#L242

### DIFF
--- a/lib/core-net/client/connect3.c
+++ b/lib/core-net/client/connect3.c
@@ -153,7 +153,7 @@ lws_client_connect_3_connect(struct lws *wsi, const char *ads,
 #if defined(LWS_WITH_SYS_FAULT_INJECTION)
 	int cfail;
 #endif
-	int m, af = 0;
+	int m, af = 0, sock_accepted = 0;
 
 	/*
 	 * If we come here with result set, we need to convert getaddrinfo
@@ -371,7 +371,7 @@ ads_known:
 		lwsl_wsi_debug(wsi, "WAITING_CONNECT");
 		lwsi_set_state(wsi, LRS_WAITING_CONNECT);
 
-		if (wsi->a.context->event_loop_ops->sock_accept)
+		if (wsi->a.context->event_loop_ops->sock_accept) {
 			if (wsi->a.context->event_loop_ops->sock_accept(wsi)) {
 				lws_snprintf(dcce, sizeof(dcce),
 					     "conn fail: sock accept");
@@ -379,6 +379,8 @@ ads_known:
 				lwsl_wsi_warn(wsi, "%s", dcce);
 				goto try_next_dns_result_closesock;
 			}
+			sock_accepted = 1;
+		}
 
 		lws_pt_lock(pt, __func__);
 		if (__insert_wsi_socket_into_fds(wsi->a.context, wsi)) {
@@ -690,8 +692,10 @@ try_next_dns_result_closesock:
 	/*
 	 * We are killing the socket but leaving
 	 */
-	compatible_close(wsi->desc.sockfd);
-	wsi->desc.sockfd = LWS_SOCK_INVALID;
+	if (!sock_accepted) {
+		compatible_close(wsi->desc.sockfd);
+		wsi->desc.sockfd = LWS_SOCK_INVALID;
+	}
 
 try_next_dns_result:
 	lws_sul_cancel(&wsi->sul_connect_timeout);


caobug:
Sorry, there is something wrong with this code.

caobug:
Yes I noticed.

We must call uv_close to close this uv_poll_init_socket handle, otherwise, when we call uv_loop_close, we will receive a resource busy or locked error prompt.

LWS calls uv_poll_init_socket here:
https://github.com/warmcat/libwebsockets/blob/40a87eaf2a1da918061e4c9bc4e3c707f152e491/lib/core-net/client/connect3.c#L375

When connection fails, come here
https://github.com/warmcat/libwebsockets/blob/40a87eaf2a1da918061e4c9bc4e3c707f152e491/lib/core-net/client/connect3.c#L242

When execution reaches here, it return 0. didn't continue to execute uv_close(...)
https://github.com/warmcat/libwebsockets/blob/40a87eaf2a1da918061e4c9bc4e3c707f152e491/lib/event-libs/libuv/libuv.c#L428

caobug:
b
c

caobug:
In order to reproduce the problem, I wrote an example. If you have time, I suggest you try to run it.

1. don't start that server program (127.0.0.1:7681)
2. start this example on command line window
3. wait a few seconds
4. Ctrl + C

```
#include <uv.h>
#include <libwebsockets.h>

static struct lws_context *context;
static uv_timer_t timer;
static uv_signal_t sigint_watcher;

static struct lws_protocols protocols[] = {
    {
        "http",
        lws_callback_http_dummy,
        0, 0, 0, NULL, 0
    },
    { NULL, NULL, 0, 0 }
};

static void timer_cb(uv_timer_t *handle) {
    struct lws_client_connect_info info;
    memset(&info, 0, sizeof(struct lws_client_connect_info));
    
    info.context = context;
    info.address = "127.0.0.1";
    info.port = 7681;
    info.protocol = protocols[0].name;
    
    lws_client_connect_via_info(&info);
}

static void signal_cb(uv_signal_t *handle, int signum) {
    lwsl_warn("exiting...\n");
    
    uv_close((uv_handle_t *)&timer, NULL);
    uv_close((uv_handle_t *)&sigint_watcher, NULL);
    
    lws_context_destroy(context);
    
    uv_stop(uv_default_loop());
}

static void walk_cb(uv_handle_t *handle, void *arg) {
    lwsl_warn("active %s %p...\n", uv_handle_type_name(handle->type), handle);
}

int main(int argc, const char *argv[]) {
    struct lws_context_creation_info info;
    memset(&info, 0, sizeof(struct lws_context_creation_info));
    
    info.foreign_loops = (void *[]){uv_default_loop()};
    info.port = CONTEXT_PORT_NO_LISTEN;
    info.protocols = protocols;
    info.options = LWS_SERVER_OPTION_LIBUV;
    info.pcontext = &context;
    if (!context) {
        exit(-1);
    }
    
    uv_signal_init(uv_default_loop(), &sigint_watcher);
    uv_signal_start(&sigint_watcher, signal_cb, SIGINT);
    
    uv_timer_init(uv_default_loop(), &timer);
    uv_timer_start(&timer, timer_cb, 0, 500);
    
    uv_run(uv_default_loop(), UV_RUN_DEFAULT);
    
    // closing..
    uv_walk(uv_default_loop(), walk_cb, NULL);
    uv_run(uv_default_loop(), UV_RUN_ONCE);
    int r = uv_loop_close(uv_default_loop());
    if (r) {
        lwsl_err("failed close loop: %s\n", uv_strerror(r));
    }
    return 0;
}
```
Output:
```
^C[2021/10/07 18:46:54:0929] W: exiting...
[2021/10/07 18:46:54:0932] N:  -- [wsi|0|pipe] (0) 6.416s
[2021/10/07 18:46:54:0933] W: active poll 0x60f000000130...
[2021/10/07 18:46:54:0933] W: active poll 0x60f000000220...
[2021/10/07 18:46:54:0933] W: active poll 0x60f000000310...
[2021/10/07 18:46:54:0933] W: active poll 0x60f000000400...
[2021/10/07 18:46:54:0933] W: active poll 0x60f0000004f0...
[2021/10/07 18:46:54:0933] W: active poll 0x60f0000005e0...
[2021/10/07 18:46:54:0933] W: active poll 0x60f0000006d0...
[2021/10/07 18:46:54:0933] W: active poll 0x60f0000007c0...
[2021/10/07 18:46:54:0933] W: active poll 0x60f0000008b0...
[2021/10/07 18:46:54:0933] W: active poll 0x60f0000009a0...
[2021/10/07 18:46:54:0933] W: active poll 0x60f000000a90...
[2021/10/07 18:46:54:0933] W: active poll 0x60f000000b80...
[2021/10/07 18:46:54:0933] W: active poll 0x60f000000c70...
[2021/10/07 18:46:54:0934] E: failed close loop: resource busy or locked
```

lws-team:
I added your patch on main a couple of hours ago

https://libwebsockets.org/git/libwebsockets/commit?id=95a8d2903c0969c1c453ca0e72c00b01dd768bf6

PR #2441, opened 2021-10-12
Error in lws_tls_server_conn_alpn.

orefkov:
When using mbedtls and LWS_HAVE_mbedtls_ssl_get_alpn_protocol not defined - lws_tls_server_conn_alpn using unitialized len.


lws-team:

PR #2452, opened 2021-10-15
Report actual socket error to user callback rather than EINPROGRESS.

FGasper:


lws-team:
Sorry about the long delay, it did not apply by the time I got to it originally.  I made it apply and it's on main now.

PR #2454, opened 2021-10-15
Contextualize logging in async-dns.c

FGasper:
Of note: A single use of lwsl_debug() remains because the function in
question is documented as a public API, but the passed-in parameter
doesn’t offer a path to a logging context.


lws-team:
Just a fyi empty format strings are warned (-> error in our builds) on some platforms.

PR #2457, opened 2021-10-18
Sshd, mbedtls 3 and Putty

orefkov:
Fixed errors detected when connecting via "putti" when using fresh mbedtls:
- In calculate sc25519 uncomment carry, otherwise the authentication signature check would failed.
- In fresh mbedtls rsa public decrypt is deprecated and not work on public keys.
  Replaced by lws_genrsa_hash_sig_verify.
- In ssh messages with channel numbers, the sender's channel number is sent instead of the recipient's channel number.
  See for example https://datatracker.ietf.org/doc/html/rfc4254#section-5.1
  Putty uses channel 256 for session, and existing code break it with "Received SSH2_MSG_CHANNEL_OPEN_CONFIRMATION for nonexistent channel 0" error.

### DIFF
--- a/plugins/ssh-base/crypto/sc25519.c
+++ b/plugins/ssh-base/crypto/sc25519.c
@@ -68,8 +68,8 @@ static void barrett_reduce(sc25519 *r, const uint32_t x[64])
       if(i+j >= 31) q2[i+j] += mu[i]*x[j+31];
   carry = q2[31] >> 8;
   q2[32] += carry;
-  //carry = q2[32] >> 8;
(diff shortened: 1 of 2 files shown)


lws-team:
Thanks a lot... I split the patch into three patches for each of the things it covered, and pushed on main + v4.3-stable.

The hack commenting the end part of sc25519 one of these un-does, is because Coverity complained about it.  From memory, there is actually a problem there in the original implementation that came from.  So I guess this will need more tracking down later.

It's good to see the sshd work getting used and contributed to.

PR #2458, opened 2021-10-19
Sshd window change

orefkov:
Added processing of window-change request, according to RFC 4254 6.7
https://datatracker.ietf.org/doc/html/rfc4254#section-6.7
For processing called callback pty_req with empty terminal and modes.


lws-team:
Thanks, this is also pushed on main and v4.3-stable.

Issue #2471, opened 2021-10-31
Simply send


hannesa2:
Hi, I want to simply send some data very fast from Raspberry and other clients should be able to listen.

I dig into `minimal-examples/client/hello_world` but to be honest it is to complicated for me. I don't want to read websites, etc.... I want only send some basic data (string or numbers) from sensors without any pointless delay.

Do you have a hint or much better an example to do so ?
Thank you

lws-team:
> I don't want to read websites, etc.... I want only send some basic data (string or numbers) from sensors without any pointless delay.

What is the "pointless delay", learning how to do what you want?

> much better an example

er... how about you write a much better issue that describes what you actually want to do.  Eg, the rpi is being a server?  Http?  Ws?  What happens to the data at the server?  What consumes it ultimately, a browser?

Tbh it sounds like you are perhaps not aligned with using a C library but maybe would be more comfortable using node, or python or somesuch, if you don't care about how it works and its resource usage.  Even then you still have to actually learn how to use whatever it is...

hannesa2:
> > I don't want to read websites, etc.... I want only send some basic data (string or numbers) from sensors without any pointless delay.
> 
> What is the "pointless delay", learning how to do what you want?

Ok, I do my best.
If it really helps you: I want to control over GSM a aircraft/glider. For this want to read gyro data on my mobile phone, when it's out of visual area.
My current solution with python has a delay, and for sure, the network/GSM has a big part of it ! That's why I'll test the whole round trip with C, if I can measure there some improvements. 
The goal is to measure if there is some improvement !

The server in C is now working on Raspberry.
Read-data in C is now working on Raspberry.
Now I want to to combine both and need a send-data on client. But I want to avoid a monolithic C server which read gyro data and provide it too.
Doing so, I want to add send-data to read-data program.

> > much better an example

Sorry to write this, but existing hello world client does too much for my point of view. For me it's not transparent how to send char/float **only**.

> er... how about you write a much better issue that describes what you actually want to do. 

I'll to my best

> Eg, the rpi is being a server? Http? Ws? What happens to the data at the server? What consumes it ultimately, a browser?

As I described above. rpi is the server. Http or Ws ? Anything which will work (best). I've no preference

> Tbh it sounds like you are perhaps not aligned with using a C library 

Yes of sure, I'm a bloody C beginner. But that's why I ask ! If I would know, there would be no reason to ask. That simple !

[...]

PR #2473, opened 2021-11-01
Readme when no TLS should be used

hannesa2:
Follow up of https://github.com/warmcat/libwebsockets/issues/2471#issuecomment-956050969


lws-team:
Thanks, it's pushed on main and v4.3-stable.

PR #2478, opened 2021-11-04
Fix inproper return

PeiweiHu:
lws_jwk_destroy(jwk) before return -1


lws-team:
Yes although it looks a bit suspicious with the return coming after the bail path, it's able to exit like that because the error path handles not being able to identify and create the right jwk.

PeiweiHu:
> Yes although it looks a bit suspicious with the return coming after the bail path, it's able to exit like that because the error path handles not being able to identify and create the right jwk.

You are correct. So I closed it immediately. :-)

lws-team:
No worries I appreciate someone is looking and sending patches

Issue #2488, opened 2021-11-11
Purpose of cache-control header for any HTTP client request


ilsme:
Hi Andy,

May I know the reason why we always need these headers for any HTTP requests including PUT, POST or DELETE?

https://github.com/warmcat/libwebsockets/blob/f595aafda53c68cb9a55d215349e5c5d46777db5/lib/roles/http/client/client-http.c#L1198-L1200

Best regards.

lws-team:
... does it make trouble for your use-case?

ilsme:
Sort of, these headers are added as object's metadata in S3 objects storage when I'm uploading them using lws based client which is undesirable for my usecase, it may be a problem of S3 configuration and probably there is a way to filter them out on S3 side but I'm wondering why were they added unconditionally for all HTTP methods by default and is it possible to remove these default headers in my code or should I modify the library?

lws-team:
Those were added more than 6 years ago, I assume because under some conditions, intermediaries otherwise served stale content.

If you want to send a patch that adds a client connection-time flag to defeat adding them, like in here

https://libwebsockets.org/git/libwebsockets/tree/include/libwebsockets/lws-client.h?h=main#n50

that sounds reasonable to me.

lws-team:
Thanks a lot for the patch.

PR #2491, opened 2021-11-11
Flag defeating default HTTP client's cache-control headers added

ilsme:


ilsme:
#2488

lws-team:
Thanks, it's pushed on main (changes ABI so it can't go on v4.3-stable).

PR #2506, opened 2021-11-29
http proxy: support PUT, PATCH and DELETE methods

viordash:


lws-team:
Thanks, it's pushed on main, it'll be on v4.3-stable next time it's pushed.

PR #2513, opened 2021-12-07
added no-cache handling

horchi:


horchi:
added no-cache handling

lws-team:
Thanks, it's pushed on main.

horchi:
Thanks!

I have not counted up a version in the patch because I do not know the procedure. Hope that was not wrong. Can I now still use one of the LWS_LIBRARY_VERSION* constants to query whether the feature is there?

lws-team:
As a convenience I updated the patch to define LWS_FEATURE_MOUNT_NO_CACHE, you can test if that #if defined().

horchi:
Perfekt, many thanks!!

PR #2520, opened 2021-12-13
http proxy: support for URL processing with max_http_header_data length 

viordash:
Hi!
This PR is for removing the 256 character limit for header data.


lws-team:
Thanks... I don't use the dynamic stack sizing from C99 personally, but a while back we started using the much saner C99 struct initialization, so I guess we are on board with C99-isms anyway.   It's pushed on main and it'll go on v4.3-stable later.

viordash:
Ok, I understand about memory rules. And in probable future PRs I will use lws_malloc/lws_free.

lws-team:
Yes, not complaining... although it's selected at runtime, the size is still ultimately controlled by the program in the sense it is whatever the context creation info said, so there's no problem.  I just haven't been using it because I wanted to be portable, but looking at this I realized the C99 ship sailed long ago, I don't have to worry about it where it's safe.

PR #2521, opened 2021-12-13
small documentation improvement

chill1n:


lws-team:
Thanks it's pushed on main and will go in v4.3-stable later.

PR #2534, opened 2022-01-09
Avoid writing null terminator into zero byte output buffer

tlsa:
Hi,

I found that if you pass a zero length output buffer to the base64 decode functions, libwebsocket will always write a null terminating byte "into" the buffer.

For context, I was writing a client for the Chrome DevTools API, and using the [captureScreenshot](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-captureScreenshot) method, which returns a JSON containing a `data:"..."` base64 encoded PNG/JPEG/WEBP. If the requested format is not supported (the server I'm using seems not to support WEBP), then the server simply returns `data:""`. I then had a wrapper like so:

```c
static bool b64_decode(const char *b64, size_t b64_len,
		uint8_t **data_out, size_t *len_out)
{
	uint8_t *data;
	int ret;

	data = malloc(b64_len);
	if (data == NULL) {
		fprintf(stderr, "%s: Allocation failed\n", __func__);
		return false;
	}

	ret = lws_b64_decode_string_len(b64, (int)b64_len,
			(char *)data, (int)b64_len);
	if (ret < 0) {
		fprintf(stderr, "%s: Base64 decode failed. ret: %i\n",
				__func__, ret);
		free(data);
		return false;
	}

	*len_out = (size_t)ret;
	*data_out = data;
	return true;
}
```

Effectively passing a 0 length input and a zero length output buffer:

* Expected result: libwebsocket would simply return 0, having read nothing and written nothing.
* Actual result: libwebsocket writes an invalid byte into the 0 length `data` allocation.

I've added a check for zero length to my code, but I thought it's worth upstreaming in libwebsocket for everyone.

I also added documentation for the base64 function's returned integers.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -7,7 +7,8 @@ them.
 Original liberal license retained:
 
   - lib/misc/sha-1.c                     - 3-clause BSD license retained, link to original [BSD3]
-  - win32port/zlib                       - ZLIB license (see zlib.h) [ZLIB]
+  - win32port/zlib
+    lib/drivers/display/upng.*		 - ZLIB license (see zlib.h) [ZLIB]
   - lib/tls/mbedtls/wrapper              - Apache 2.0 (only built if linked against mbedtls) [APACHE2]
     lib/tls/mbedtls/mbedtls-extensions.c
   - lib/misc/base64-decode.c             - already MIT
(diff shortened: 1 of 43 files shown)


lws-team:
Thanks, I pushed it on main... it seems it should probably wait for v4.4 rather than update v4.3 with it, since it's kind of an api change.

PR #2535, opened 2022-01-11
Fix compilation issue with OpenSSL subproject

fgerlits:
We (are going to) use libwebsockets in our project: https://github.com/apache/nifi-minifi-cpp.

We download and compile both OpenSSL(*) and libwebsockets during the build process.  This doesn't quite work out of the box, because CMake sets `OPENSSL_FOUND` to `TRUE` and OpenSSL sets `OPENSSL_INCLUDE_DIR` (and `OPENSSL_LIBRARIES`), but it does not set `OPENSSL_INCLUDE_DIRS`, which is read by libwebsockets.

This patch fixes the issue, so it will allow us, and others, to use libwebsockets in this way.

(*) we currently use the LibreSSL flavor of OpenSSL for licensing reasons, but according to [FindOpenSSL.cmake](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenSSL.cmake), setting `OPENSSL_INCLUDE_DIR` but not `OPENSSL_INCLUDE_DIRS` is the correct behavior and other flavors probably do the same.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,6 +159,7 @@ option(LWS_WITH_SYS_STATE "lws_system state support" ON)
 option(LWS_WITH_SYS_SMD "Lws System Message Distribution" ON)
 option(LWS_WITH_SYS_FAULT_INJECTION "Enable fault injection support" OFF)
 option(LWS_WITH_SYS_METRICS "Lws Metrics API" OFF)
+option(LWS_WITH_UPNG "Enable stateful PNG stream decoder" ON)
 
 #
 # Secure Streams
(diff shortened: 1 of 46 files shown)


lws-team:
Thanks for the patch, I applied it on main, but it breaks macos and windows builds in CI, eg

https://libwebsockets.org/sai/index.html?task=6738739bf2ceabdfe4c7afc2639ce9f801cde2ae6355c8c88579912c73101077

```
In file included from /Users/sai/jobs/0-1.0/libwebsockets/lib/plat/unix/unix-misc.c:28:
In file included from /Users/sai/jobs/0-1.0/libwebsockets/lib/core/./private-lib-core.h:146:
/Users/sai/jobs/0-1.0/libwebsockets/include/libwebsockets.h:306:10: fatal error: 'openssl/ssl.h' file not found
In file included from /Users/sai/jobs/0-1.0/libwebsockets/lib/plat/unix/unix-caps.c:28:
In file included from /Users/sai/jobs/0-1.0/libwebsockets/lib/core/./private-lib-core.h:146:
/Users/sai/jobs/0-1.0/libwebsockets/include/libwebsockets.h:306:10: fatal error: 'openssl/ssl.h' file not found
#include <openssl/ssl.h>
```

https://libwebsockets.org/sai/index.html?task=6738739bf2ceabdfe4c7afc2639ce9f8b2af91b29324e4446b77988c130b3dec

lws-team:
Yes that solved it, thanks.  It's on main and will be on v4.3-stable next time that's updated.

PR #2548, opened 2022-01-27
Don't require ':' to be a part of header name in lws_add_http_header_by_name AGAIN

yogpstop:
https://github.com/warmcat/libwebsockets/commit/0c5aa2f4843c17815e87f2daedb54b8d0b4c0de9
This commit makes to recover header.c, but it's missing #273 patch.

https://github.com/warmcat/libwebsockets/blob/main/minimal-examples-lowlevel/http-client/minimal-http-client-custom-headers/minimal-http-client-custom-headers.c#L46
Unfortunately, above example doesn't contain colon to call `lws_add_http_header_by_name`.
And it doesn't work correctly.

So I fixed it.


lws-team:

PR #2569, opened 2022-02-16
add Openssl client callback also for wolfssl

tiolan:
Please consider this WIP for now.

When using WolfSSL and connecting a client against a TLS server, the callback `LWS_CALLBACK_OPENSSL_PERFORM_SERVER_CERT_VERIFICATION` is not invoked - also not on `protocols[0]`.

Going through the code, I realized it is implemented to only do this for OpenSSL.

This change allows to re-use the existing OpenSSL callback `OpenSSL_client_verify_callback` also for WolfSSL.

What I tested as of now is loading the root certificate via `client_ssl_ca_filepath`, set `LCCSCF_SKIP_SERVER_CERT_HOSTNAME_CHECK` (because it is not implemented for WolfSSL) and then do the hostname check inside the verify callback. Also an OCSP check etc. could be done in this callback.

I am not yet sure about the already existing logic insde `OpenSSL_client_verify_callback` and whether this fits to WolfSSL.
An option would be to introduce `WolfSSL_client_verify_callback`, that only does
```
	n = lws_get_context_protocol(wsi->a.context, 0).callback(wsi,
			LWS_CALLBACK_OPENSSL_PERFORM_SERVER_CERT_VERIFICATION,
			x509_ctx, ssl, (unsigned int)preverify_ok);
```
and leaves the rest to the user.

Happy about any thoughts.

### DIFF
--- a/lib/tls/openssl/openssl-client.c
+++ b/lib/tls/openssl/openssl-client.c
@@ -44,7 +44,6 @@ int lws_openssl_describe_cipher(struct lws *wsi);
 extern int openssl_websocket_private_data_index,
     openssl_SSL_CTX_private_data_index;
 
-#if !defined(USE_WOLFSSL)
 
 #if 0
 #if defined(LWS_WITH_TLS_JIT_TRUST)
@@ -228,7 +227,6 @@ OpenSSL_client_verify_callback(int preverify_ok, X509_STORE_CTX *x509_ctx)
 	 */
 	return !n;
 }
-#endif
 
 int
 lws_ssl_client_bio_create(struct lws *wsi)
@@ -368,6 +366,8 @@ lws_ssl_client_bio_create(struct lws *wsi)
 #else
 	if (wsi->tls.use_ssl & LCCSCF_ALLOW_SELFSIGNED)
 		wolfSSL_set_verify(wsi->tls.ssl, SSL_VERIFY_NONE, NULL);
+	else if (wsi->tls.use_ssl & LCCSCF_USE_SSL)
+		wolfSSL_set_verify(wsi->tls.ssl, SSL_VERIFY_PEER, OpenSSL_client_verify_callback);
 #endif
 #endif /* USE_WOLFSSL */
 


lws-team:
Many users don't take much care about security, so the main thing is it should continue to default to requiring a trusted CA and checking the necessary pieces itself, ie, it shouldn't require the callback and should still default to secure if it is not provided.

The callbacks to do this kind of thing were added a long time ago when there was only 1 vhost and 1 protocol typically.  The callback becomes difficult to access if, eg, there is no user lws_protocol because it just does http and wants to use the default, internal "dummy callback".  So for new things, it is better to put a function pointer you can set in the context info if you want to get a look-in.

If the goal is OCSP, rather than push it to user code to deal with, it'd be more ideal to add the client check part of that in lws itself, eg with a configurable lws_cache_ttl in heap remembering the last few while they are valid.  Then even if you personally only have energy to implement it around wolfssl, the generic pieces (all the OCSP fetch and perhaps validation pieces are generic) are established and someone else can bring them to openssl / mbedtls.  Then, that functionality will go in the CI here and stand a much better chance of not getting broken.

lws-team:
> if all the OpenSSL methods do the same in WolfSSL 

Yes that's why I said it, there's a danger that replaces rather than augments the internal verification flow.

> I guess you rather proposed to implement a generic solution right? Some users might need to do the OCSP call through a proxy, some don't.

Whatever else they're planning to do, they need to fetch a URL and check the signature.  Lws already supports `http_proxy` which will be set if the client must use one... lws should already support anything they're going to need to do the network query async-wise and check the crypto.

There's going to need to be some magic somewhere to hold up and resume the tls connection until OCSP results are available, I can help with that if it's generic.

> I saw lws using the "global" function calls dealing with SSL*

What apis are we talking about?  Generally touching SSL is specific to one connection.

tiolan:
> Lws already supports http_proxy which will be set if the client must use one... 

In case one uses an lws client for doing the OCSP manually, yes. If using the client built into e.g. WolfSSL, the situation would be different.

> There's going to need to be some magic somewhere to hold up and resume the tls connection until OCSP results are available

If you use the OCSP API of the TLS library probably not, as it would do it during the TLS negotiation.
E.g. https://www.wolfssl.com/wolfssl-online-certificate-status-protocol-ocsp-support/
So, I would personally prefer leaving those additional things like OCSP to the user, as they now how they want to do it. But it is a matter of taste.

> What apis are we talking about? Generally touching SSL is specific to one connection.

I was referring to e.g. `wolfSSL_CTX_set_verify(WOLFSSL_CTX*, ...)` vs. `wolfSSL_set_verify(WOLFSSL*, ...)` 

Let me go through the code and think about what I can implement, being new to lws.

lws-team:
> If using the client built into e.g. WolfSSL, the situation would be different.

Lws supports, eg, mbedtls,

>>> I saw lws using the "global" function calls dealing with SSL*

>> What apis are we talking about? Generally touching SSL is specific to one connection.

> wolfSSL_set_verify(WOLFSSL*, ...)

AFAIK that will just set the verify callback for a single connection tunnel, represented by the [WOLF]SSL, not globally.

PR #2579, opened 2022-03-01
Remove redundant redeclaration

tjwalton:
This is already defined at line 554.


lws-team:
Thanks it's pushed on main.

PR #2580, opened 2022-03-02
add callback reason for generic TLS purposes for clients

tiolan:
This follows up on #2569.
When the client TLS pointer has been configured, it is handed over to user code via a new callback reason.
Optionally the user code can modify the TLS pointer, e.g. do activate additional checks etc.
The TLS handshake will be aborted, of the user returns != 0.

`lws_callback_dummy` returns 0 per default, so this should not influence existing applications.

In case you would want to merge this, is there a test suite I can run, making sure nothing got broken?


lws-team:
Thanks... as I mentioned on the previous PR

> The callbacks to do this kind of thing were added a long time ago when there was only 1 vhost and 1 protocol typically. The callback becomes difficult to access if, eg, there is no user lws_protocol because it just does http and wants to use the default, internal "dummy callback". So for new things, it is better to put a function pointer you can set in the context info if you want to get a look-in.

So I don't really want to do what this is doing.  Keeping the interesting part out of tree makes it meh.

tiolan:
Okay I misunderstood this.
What I cannot do is, implement all the TLS logic directly in lws (OCSP, hostname check, etc.). Especially for my use-case this would not be flexible enough. I would not be able to integrate this into our environment and the implementation inside lws would be very specific.

Coming back to your proposal, I would add a member `int (*tls_configure_callback)(void* user_space, void* tls)` (or similar) to `lws_client_connect_info`.
Then at the same position where I now call `LWS_CALLBACK_CLIENT_CONFIGURE_TLS`, I would call this callback.

```
if ((*tls_configure_callback)(wsi->user_space, wsi->tls.ssl)) {
	*pcce = "user aborted in tls_configure_callback";
	return CCTLS_RETURN_ERROR;
}
```

Did I get you correct?

Thanks

lws-team:
> Okay I misunderstood this.
> What I cannot do is, implement all the TLS logic directly in lws

Hmmm I bet almost everything can go inside lws just fine, quite a lot better than what you have atm.

I am thinking about a 2-layer (heap, and file-backed) `lws_cache_ttl,` fed from async lws-handled OCSP lookups.  Instead of blocking the event loop while it checks OCSP, add an extra "wait for OCSP" state on the wsi after validated connection so there is no blocking.

> Especially for my use-case this would not be flexible enough.

What else do you actually need?

tiolan:
I for instance would not want to let lws do the OCSP check with some buffered data, etc.
Specifically I would want to let wolfssl do it, without any implementation around it.
So no need for an additional "wait for OCSP" step.
I would also want to support OCSP stapling for some connections, while for other parallel connections I would want the normal OCSP.

There needs to be support for HTTP Proxies, when wolfssl does the OCSP request (not via lws client).

Further I have to configure a sign callback, because signing cannot happen inside the TLS library if for example the private key is not accessible.

And the hostname check also must be configurable based on the server I connect to.

So summarized, I need access to the TLS pointer.

I think a generic callback such as I proposed (via the client connect info) would allow users to highly customize their TLS setup and adapt to their environment.

lws-team:
Maintainers have different perspectives than individual users... your thinking ends at wolfssl since that's what is in front of you, but OCSP is also useful for other tls library users, eg, mbedtls users.  I spent many man-months making tls library-based features available via lws the same no matter if mbedtls or openssl-based and both kinds are widely used, eg, lws + mbedtls + freertos.

As a library, telling each user that wants it to reimplement his own proprietary tls-library specific OCSP stack through a callback wormhole is not supporting the feature in any useful way.

Mbedtls is less monolithic than the openssl approach, it is not going to provide the blocking, high level 'I'll get that for you' interface that wolfssl might have.

> So no need for an additional "wait for OCSP" step.

Isn't your implementation going to block in this callback while it waits an open-ended amount of time for wolfssl to contact the OCSP server?  In-band sequencing it can do nonblocking because lws manages the peer connection fd on the lws event loop.  But if wolfssl wants to manage its own fds in another event loop, lws will be blocked until it gets a result.

Adding another later step to a normal tls connection that passes its usual validation, to hold it asynchronously while the OCSP lookup is done means there is no blocking.

> check with some buffered data, etc.

OCSP stapling implies the validation has a TTL and is cached (at the server in that case).  You could artificially restrict the TTL configurably, but mainly, people don't want every connection over a few seconds to check for the same cert's revocation.  So it seems a client cache that respects TTL is desirable.

> There needs to be support for HTTP Proxies,

You mentioned this before, but what does that mean to you... lws supports http proxies.  You mean getting the OCSP result from a different URL than told in the cert?

> I would also want to support OCSP stapling for some connections

OCSP stapling validation probably is something the tls library should be left to do if it supports it... it gets the OCSP response in-band to an already established connection, it saw the other certs in-band, it is in a position to validate it or not all by itself in the normal run of things.  Since it's already doing the connection action async, it can do it without blocking too.

Mbedtls doesn't support OCSP stapling validation at all, neither fetching the result nor even the extension to deliver it AFAICS.

tiolan:
> Maintainers have different perspectives than individual users... your thinking ends at wolfssl since that's what is in front of you, but OCSP is also useful for other tls library users, eg, mbedtls users

Yes that is totally true. I do agree to all your points and can fully understand them.

> As a library, telling each user that wants it to reimplement his own proprietary tls-library specific OCSP stack through a callback wormhole is not supporting the feature in any useful way.

That was not my intend. OCSP was just one example of what people could do.

> Isn't your implementation going to block in this callback while it waits an open-ended amount of time for wolfssl to contact the OCSP server?

This is absolutely correct. For my usecase that is intended, for other users it might not be.
This is also why I did not want to include my specific implementation of wolfssl into lws.

> You mentioned this before, but what does that mean to you... lws supports http proxies. You mean getting the OCSP result from a different URL than told in the cert?

No I mean that I (in my specific use case) need to configure the HTTPS client inside wolfssl to use a proxy.

So, to wrap it up.
All I wanted was to provide the TLS pointer to the user, so that advanced/expert users can customize to their needs.
OCSP, signature callbacks, verify callbacks, PKCS11 and all the other functionality would just be examples of what advanced developers could, but not have to do.
I could wrap this callback in a precompiler flag (define), to only do it if users actively activate it - making sure nobody accidentially makes the standard implementation of TLS insecure.
Of course, implementing additional functionality directly in lws would be good and I do not see it standing against the approach to hand over the TLS pointer to the user.

Is there some implementation for handing over the TLS pointer, that you could take in and see advantage of?

tiolan:
Closing this, because there is `provided_client_ssl_ctx` already available.

PR #2587, opened 2022-03-09
Fix FindOpenSSLbins error when no OpenSSL executable found

DerBer:
This fixes CMake error "mode MATCH needs at least 5 arguments total to
command" while configuring when no OpenSSL executable was found.
Tested under Windows using VS2019.


lws-team:
Thanks, it's pushed on main.

lws-team:
... I see it came from 4.3, I will backport things to that, including this, later this week.

DerBer:
Thank you!

PR #2613, opened 2022-04-13
CMakeLists.txt: fix build without C++

ffontaine:
Fix the following build failure without C++ raised since version 4.2.0 and https://github.com/warmcat/libwebsockets/commit/962e9ee345bc28749577c1fca2542a273404627d:

```
CMake Error at /nvmedata/autobuild/instance-11/output-1/per-package/libwebsockets/host/share/cmake-3.18/Modules/CMakeTestCXXCompiler.cmake:59 (message):
  The C++ compiler

    "/usr/bin/clang++"

  is not able to compile a simple test program.
```

Fixes:
 - http://autobuild.buildroot.org/results/550e7f7d54adf74f8cc078be5b91b3567d622ede

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


lws-team:
Thanks that's much better, it's pushed on main.

PR #2622, opened 2022-04-21
Remove esp_attr.h file

hmalpani:
We can remove this file as it is present in esp-idf: [esp_attr.h](https://github.com/espressif/esp-idf/blob/master/components/esp_common/include/esp_attr.h)


lws-team:
Thanks, this is pushed on main, I had to add a couple of extra preprocessor conditionals since it tried to bring in esp_attr.h when building for freertos with or without esp-idf.

PR #2632, opened 2022-05-03
validity: fix bad pointer access

SevastianovAndrii:
We cannot access wsi after `__lws_close_free_wsi`.


lws-team:
Youch... thanks, it's pushed on main

https://libwebsockets.org/git/libwebsockets/commit?id=68d0b42626ec0a8084939ac479724f0c039f73ce

PR #2705, opened 2022-08-05
backport from main: WSAEWOULDBLOCK as MBEDTLS_ERR_SSL_WANT_READ|WRITE

bogdad:
Hey, 
thank you for libwebsockets!

please consider merging this, found while debugging issue:
4.30-stable branch cant ssl connect on windows (with mbedtls, did not test openssl).
main works fine.

error is:
`mbedtls connect -1 5 10035`

caused by `lws_plat_mbedtls_net_recv` getting WSAWOULDBLOCK (10035) while checking for EWOULDBLOCK(140).

main works because in this commit https://github.com/warmcat/libwebsockets/commit/53d195022ff6519621da779f6a958d7b4d84f0da `lws_plat_mbedtls_net_recv` started considering WSAWOULDBLOCK as MBEDTLS_ERR_SSL_WANT_READ.
(the commit is about cmake examples change, so maybe the WSAWOULDBLOCK change got there by accident)

I am open to just cherry-picking https://github.com/warmcat/libwebsockets/commit/53d195022ff6519621da779f6a958d7b4d84f0da instead, please say if you want me to open this cherry-pick pr!

maybe an even better change would be to change both main and v4.30-stable to do this instead

```
@@ -594,7 +594,7 @@ lws_plat_mbedtls_net_recv(void *ctx, unsigned char *buf, size_t len)
                return ret;

        en = LWS_ERRNO;
-       if (en == EAGAIN || en == EWOULDBLOCK)
+       if (en == LWS_EAGAIN || en == LWS_EWOULDBLOCK)
                return MBEDTLS_ERR_SSL_WANT_READ;

        ret = WSAGetLastError();
```
this seems how `LWS_ERRNO` is used elsewhere, please say if you prefer me to do this instead!

Thank you again,
Vladimir.

### DIFF
--- a/lib/plat/windows/windows-sockets.c
+++ b/lib/plat/windows/windows-sockets.c
@@ -569,7 +569,7 @@ lws_plat_mbedtls_net_send(void *ctx, const uint8_t *buf, size_t len)
 		return ret;
 
 	en = LWS_ERRNO;
-	if (en == EAGAIN || en == EWOULDBLOCK)
+	if (en == EAGAIN || en == EWOULDBLOCK || en == WSAEWOULDBLOCK)
 		return MBEDTLS_ERR_SSL_WANT_WRITE;
 
 	ret = WSAGetLastError();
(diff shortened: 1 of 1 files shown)


lws-team:
Thanks, it's pushed on v4.3-stable.

PR #2709, opened 2022-08-17
remove LWS_CALLBACK_OPENSSL_CONTEXT_REQUIRES_PRIVATE_KEY callback

### LINKED ISSUE #2495: server: no OPENSSL_CONTEXT_REQUIRES_PRIVATE_KEY when ssl_private_key_filepath is NULL
According to the documentation of `ssl_private_key_filepath` in `lws_context_creation_info`

>   /**<  VHOST: filepath to private key if wanting SSL mode;
>     * if this is set to NULL but ssl_cert_filepath is set, the
>     * OPENSSL_CONTEXT_REQUIRES_PRIVATE_KEY callback is called
>     * to allow setting of the private key directly via openSSL
>     * library calls.   (For backwards compatibility, this can also be used
>     * to pass the client cert private key filepath when setting up a
>     * vhost client SSL context, but it is preferred to use
>     * .client_ssl_private_key_filepath for that.)

This worked fine in older versions of libwebsockets (prior to 3.0), but doesn't seem to be the case any more.
The server certificate specified in `ssl_cert_filepath` is not included in the `SSL_CTX` and I don't seem to get
the OPENSSL_CONTEXT_REQUIRES_PRIVATE_KEY callback.

It looks like the ssl context processing stops [here](https://github.com/warmcat/libwebsockets/blob/40a87eaf2a1da918061e4c9bc4e3c707f152e491/lib/tls/openssl/openssl-server.c#L183) when 
only `ssl_cert_filepath` is specified (I am not using the `mem_cert`,  `mem_private_key` settings).

I found though, that I can workaround this issue for now by specifying both the certificate and private key in the LWS_CALLBACK_OPENSSL_LOAD_EXTRA_SERVER_VERIFY_CERTS callback.

dhobsong:
When a certificate for a TLS connection is provided, but a private
key is not, the SSL_CTX initialization exits early, before the
CONTEXT_REQUIRES_PRIVATE_KEY callback can be issued.
Remove the now obsolete callback and update the vhost
field description to state that the LOAD_EXTRA_SERVER_VERIFY_CERTS
callback should be used instead.

Fixes #2495

### DIFF
--- a/include/libwebsockets/lws-callbacks.h
+++ b/include/libwebsockets/lws-callbacks.h
@@ -160,15 +160,6 @@ enum lws_callback_reasons {
 	 * the default callback action of returning 0 allows the client
 	 * certificates. */
 
-	LWS_CALLBACK_OPENSSL_CONTEXT_REQUIRES_PRIVATE_KEY	= 37,
-	/**< if configured for including OpenSSL support but no private key
-	 * file has been specified (ssl_private_key_filepath is NULL), this is
-	 * called to allow the user to set the private key directly via
-	 * libopenssl and perform further operations if required; this might be
-	 * useful in situations where the private key is not directly accessible
-	 * by the OS, for example if it is stored on a smartcard.
-	 * user is the server's OpenSSL SSL_CTX* */
-
 	LWS_CALLBACK_SSL_INFO					= 67,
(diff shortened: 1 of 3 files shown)


lws-team:

Issue #2668, opened 2022-06-23
Unable to connect on Windows platform


MacGe:
Hi Andy,
    I'd like to add some header when get LWS_CALLBACK_CLIENT_APPEND_HANDSHAKE_HEADER callback, but if i didn't sleep some time like 5ms， i will get LWS_CALLBACK_CLOSED_CLIENT_HTTP immediately, just wondering why only Windows platform have such behavior or is my use is incorrect.

 
![image](https://user-images.githubusercontent.com/25633087/175250379-9386c909-7c32-4939-bcdb-0b30fee1cf03.png)

lws-team:
Do you have any other threads in your code?

You should look at the logs to see why it feels it wants to close.

lws-team:
Have a read of this

https://libwebsockets.org/git/libwebsockets/tree/READMEs/README.event-loops-intro.md

MacGe:
Thank you Andy , i've read it before. i don't think its thread related but more platform related, because this code works fine on Linux and Androd, even in 60 channel concurrent scenarios. but when comes to windows, without the sleep , it failed to connect

lws-team:
> and new connection from new thread attach to it.

How are you handling that?  You can't call lws apis from the other thread.

> You should look at the logs

Your logs don't tell us anything.  You need to look at the verbose lws logs.

lws-team:
10035 is WOULDBLOCK it's not an error but it doesn't mean we're connected either, it just means it is ongoing.

What version of lws is this?  Still the same with latest main?

MacGe:
> 10035 is WOULDBLOCK it's not an error but it doesn't mean we're connected either, it just means it is ongoing.
>  yes 10035means noblocking socket, means next send should also return 10035 other than 10057
> What version of lws is this? Still the same with latest main?

4.3 stable.

MacGe:
I'll try it with  latest main and keep you updated

Mitch-Martinez:
I have been seeing similar behavior when using mbedtls + windows, I hope this helps at least identify the issue a little better

I had to modify `lws_plat_mbedtls_net_recv` and `lws_plat_mbedtls_net_send` to include WSAEWOULDBLOCK to trigger a want_read or want_write.
```
int
lws_plat_mbedtls_net_recv(void *ctx, unsigned char *buf, size_t len)
{
...
-	if (en == EAGAIN || en == EWOULDBLOCK)
[...]

Issue #2700, opened 2022-07-20
When client created with invalid server IP ... goes in constant looping for 17-20 seconds


Paramveer24b:
Description:-
In windows 
I created a client, and gave invalid IP to connect as a server.
There is a function in libwebsocket 

     lws_plat_service_tsi()
    {
      // In below loop if value of n is positive then this loops continues .
         for (i = 0; i < pt->fds_count; ++i) {
	  	pfd = &pt->fds[i];
	  	if (!(pfd->events & LWS_POLLOUT))
	  		continue;

		wsi = wsi_from_fd(context, pfd->fd);
		if (!wsi || wsi->listener)
			continue;
		if (wsi->sock_send_blocking)
			continue;
		pfd->revents = LWS_POLLOUT;
		n = lws_service_fd(context, pfd);
		if (n < 0 )
			return -1;

		/*
		 * Force WSAWaitForMultipleEvents() to check events
		 * and then return immediately.
		 */
		timeout_us = 0;

		/* if something closed, retry this slot */
		if (n)
			i--;
		printf("looping\n");
	}
}

In my case n > 0 thats why loop gets continued ...but this situation remains for 17-20 seconds every time.
In result for loop continues in constant looping for this much time.

But this is not the case on Linux with same client with invalid server IP address.

there is a check to break this loop as below  (different code for UNIX and Windows in libwebsocket)

    /* any socket with events to service? */
    	for (n = 0; n < (int)pt->fds_count; n++) {
		lws_sockfd_type fd = pt->fds[n].fd;

		if (!pt->fds[n].revents)
			continue;

		m = lws_service_fd_tsi(context, &pt->fds[n], tsi);
		if (m < 0) {
			lwsl_err("%s: lws_service_fd_tsi returned %d\n",
				 __func__, m);
			return -1;
		}

		/* if something closed, retry this slot since may have been
		 * swapped with end fd */
		if (m && pt->fds[n].fd != fd)
			n--;
	}

If I apply pt->fds[n].fd != fd  this condition in windows then above issue resolved.

Is this correct solution for above issue or I have to do something else.

And if this is correct ...will you be able to provide a patch for this ?

Arenoros:
I encountered the same problem, and it depends on whether it is just a wrong address or whether there is a port blocking

<details>
<summary>Log with normal behavior 10.10.10.1</summary>

```log
[2022/08/30 16:05:41:8569] U: LWS minimal ws client
[...]

PR #2715, opened 2022-08-24
windows: use select(1us) instead of connect to check connection state.

bogdad:
Hey : )

This pr made from out of this comment :
https://github.com/warmcat/libwebsockets/issues/2668#issuecomment-1178681601 (thank you @MacGe) !

It solves the following issue for us:

A windows connection problem on 4.3-stable branch, 
Seems on windows 10, establishing of tcp connection gets `WSAWOULDBLOCK`(10035)  upon first try in `lws_client_connect_3_connect:516` 
`m = connect(wsi->desc.sockfd,` 

And then our client code calls `lws_service()`. `_lws_plat_service_tsi`  immediately tries proceeds with checking the socket  `lws_client_connect_3_connect  lws_client_connect_check` : gets errno `WSAEINVAL` 10022 which it treats as  `LCCCR_CONTINUE`.

Then `_lws_plat_service_tsi`  does some processing, but does not check the socket again,  and schedules a timeout in 15 seconds.
```
_lws_plat_service_tsi: in 15000ms, count 2
_lws_plat_service_tsi: fd 5828 ev 0x300 POLLIN 768, POLLOUT 16
_lws_plat_service_tsi: fd 4980 ev 0x300 POLLIN 768, POLLOUT 16
```

then it continues waiting doing nothing for 15 seconds, logs below, confirmed in debugger that nothing really happens and we are just sleeping.

In 15 seconds lws closes the socket.
`lws_service_fd_tsi: Close and handled`

It is possible to delay our client calling the  `lws_service()` call  `lws_client_connect_check` gets an established socket, and everything works but without the delay I consistently get the `WSAEINVAL` behaviour.

In summary:
Seems handling `WSAEINVAL` from `lws_client_connect_check` gets in a state so poll decides to not check the connection any more / or windows does not update the socket status.

According to this https://stackoverflow.com/questions/15403192/wsaconnect-returns-wsaeinval-on-windowsxp one should not call connect after a previous call returns `WSAWOULDBLOCK`. so maybe that is what breaks socket.

I am not very sure regarding this, maybe a better way exists will happily change the approach if needed!
Did test the happy path, did test the timeout when we are connecting to an non existant port of existing dns name - behaves the same was as without the patch.

did not test `select`  error case, did not test connection takes more than 1 millisecond to be established,
overall not very sure no socket states would be lost.
would appreciate a review very much.

Building with 
```
set(LWS_ROLE_H1                   ON)
set(LWS_ROLE_WS                   ON)
set(LWS_WITH_MBEDTLS              YES)
set(LWS_HAVE_MBEDTLS_NET_SOCKETS  YES)
set(LWS_HAVE_mbedtls_ssl_get_alpn_protocol YES)
set(LWS_WITH_NO_LOGS              NO)
set(LWS_OPENSSL_SUPPORT           YES)
set(LWS_WITH_CLIENT               YES)
set(LWS_WITH_NETWORK              YES)
set(LWS_WITH_TLS                  YES)
set(LWS_WITHOUT_EXTENSIONS        YES)
set(LWS_WITHOUT_SERVER            YES)
set(LWS_WITH_FILE_OPS             ON)
set(LWS_WITH_POLL                 YES)
```

Below are logs/logging patch.
This is the gist of the log with buggy behavior (with additional added log statements 'xxxx smt' to get error codes and such), 
https://gist.github.com/bogdad/47ae5c8a8a27e198064bbb1b781e4a89

a gist of the log with this pr applied that successfully proceeds to embed tls handshake after creating connection.
https://gist.github.com/bogdad/ae98f65b40a6ea553b1431610e7d727f

, and a gist of patch that adds the logging for reference.
https://gist.github.com/bogdad/adbc93754612d2c7d6c2de9e24f5fe1d

please say if you want me to target main instead!

### DIFF
--- a/lib/core-net/client/connect3.c
+++ b/lib/core-net/client/connect3.c
@@ -109,38 +109,33 @@ lws_client_connect_check(struct lws *wsi, int *real_errno)
 	}
 
 #else
+	fd_set write_set;
+	struct timeval tv;
+	int ret;
 
-	if (!connect(wsi->desc.sockfd, (const struct sockaddr *)&wsi->sa46_peer.sa4,
-#if defined(WIN32)
-				sizeof(struct sockaddr)))
-#else
-				0))
-#endif
+	FD_ZERO(&write_set);
+	FD_SET(wsi->desc.sockfd, &write_set);
+
+	tv.tv_sec = 0;
+	tv.tv_usec = 1;
+
+	ret = select(wsi->desc.sockfd + 1, NULL, &write_set, NULL, &tv);
+	if (ret > 0) {
+		lwsl_wsi_debug(wsi, "conn OK");
 
 		return LCCCR_CONNECTED;
-
-	en = LWS_ERRNO;
-
-	if (en == WSAEISCONN) /* already connected */
-		return LCCCR_CONNECTED;
-
-	if (en == WSAEALREADY) {
-		/* reset the POLLOUT wait */
-		if (lws_change_pollfd(wsi, 0, LWS_POLLOUT))
-			lwsl_wsi_notice(wsi, "pollfd failed");
 	}
 
-	if (!en || en == WSAEINVAL ||
-		   en == WSAEWOULDBLOCK ||
-		   en == WSAEALREADY) {
-		lwsl_wsi_debug(wsi, "%s",
-				lws_errno_describe(en, t16, sizeof(t16)));
+	if (!ret) {
+	    lwsl_wsi_debug(wsi, "select timeout");
 
 		return LCCCR_CONTINUE;
 	}
+
+	en = LWS_ERRNO;
 #endif
 
-	lwsl_wsi_notice(wsi, "connect check FAILED: %s",
+	lwsl_wsi_notice(wsi, "connection check FAILED: %s",
 			lws_errno_describe(*real_errno || en, t16, sizeof(t16)));
 
 	return LCCCR_FAILED;


lws-team:
Does this still work (against main)... reduce the wait to 1us (no point blocking for 1ms).

```
From: Vladimir Shakhov <bogdad@gmail.com>
Date: Wed, 24 Aug 2022 16:06:10 +0200

windows: use select instead of connect to check  connection state

https://patch-diff.githubusercontent.com/raw/warmcat/libwebsockets/pull/2715

# Everything here is editable! You can modify the patch name, author,
# date, commit message, and the diff (if --diff was given).
# Lines starting with '#' will be ignored, and an empty message
# aborts the edit.
---

 lib/core-net/client/connect3.c |   35 +++++++++++++++--------------------
 1 file changed, 15 insertions(+), 20 deletions(-)

diff --git a/lib/core-net/client/connect3.c b/lib/core-net/client/connect3.c
index ebd0ca3740999e4aa49dd24ffcd0376a5bb7ac37..28a3d0e389806746f5d267a89be0bb1018a694cf 100644
--- a/lib/core-net/client/connect3.c
+++ b/lib/core-net/client/connect3.c
@@ -109,38 +109,33 @@ lws_client_connect_check(struct lws *wsi, int *real_errno)
        }

 #else
+       fd_set write_set;
+       struct timeval tv;
+       int ret;

-       if (!connect(wsi->desc.sockfd, (const struct sockaddr *)&wsi->sa46_peer.sa4,
-#if defined(WIN32)
-                               sizeof(struct sockaddr)))
-#else
-                               0))
-#endif
+       FD_ZERO(&write_set);
+       FD_SET(wsi->desc.sockfd, &write_set);

-               return LCCCR_CONNECTED;
+       tv.tv_sec = 0;
+       tv.tv_usec = 1;

-       en = LWS_ERRNO;
+       ret = select(wsi->desc.sockfd + 1, NULL, &write_set, NULL, &tv);
+       if (ret > 0) {
+               lwsl_wsi_debug(wsi, "conn OK");

-       if (en == WSAEISCONN) /* already connected */
                return LCCCR_CONNECTED;
-
-       if (en == WSAEALREADY) {
-               /* reset the POLLOUT wait */
-               if (lws_change_pollfd(wsi, 0, LWS_POLLOUT))
-                       lwsl_wsi_notice(wsi, "pollfd failed");
        }

-       if (!en || en == WSAEINVAL ||
-                  en == WSAEWOULDBLOCK ||
-                  en == WSAEALREADY) {
-               lwsl_wsi_debug(wsi, "%s",
-                               lws_errno_describe(en, t16, sizeof(t16)));
+       if (!ret) {
+               lwsl_wsi_debug(wsi, "select timeout");

                return LCCCR_CONTINUE;
        }
+
+       en = LWS_ERRNO;
 #endif

-       lwsl_wsi_notice(wsi, "connect check FAILED: %s",
+       lwsl_wsi_notice(wsi, "connection check FAILED: %s",
                        lws_errno_describe(*real_errno || en, t16, sizeof(t16)));

        return LCCCR_FAILED;
```

bogdad:
yes, it does, just checked! agree, 1 us is better.

bogdad:
changed base to main, applied your changes, thank you!

Mitch-Martinez:
To add to this commit I think it would be best to utilize the exception fd set in `select`. The exception fd set will allow select to identify if the connection fails while leaving it null the select call will happily ignore that the connection has failed and return a timeout.

This way, in addition to avoiding consecutive calls to `connect` windows clients can actually wait for the connect call to succeed or fail.

More or less the changes would look like 

```
fd_set write_set, except_set;
struct timeval tv;
int ret;

FD_ZERO(&write_set);
FD_ZERO(&except_set);
FD_SET(wsi->desc.sockfd, &write_set);
FD_SET(wsi->desc.sockfd, &except_set);

tv.tv_sec = 0;
tv.tv_sec = 1;

ret = select(wsi->desc.sockfd + 1, NULL, &write_set, &except_set, &tv);

en = LWS_ERRNO;

if (FD_ISSET(wsi->dec.sockfd, &write_set)) {
    /* actually connected */
    lwsl_wsi_debug(wsi, "select write fd set, conn OK");
    return LCCCR_CONNECTED;
}

if (FD_ISSET(wsi->desc.sockfd, &except_set)) {
    /* Failed to connect */
    lwsl_wsi_notice(wsi, "connect failed, select exception fd set");
    return LCCCR_FAILED;
}
if (ret == 0) {
    lwsl_wsi_notice(wsi, "select timeout");
    return LCCCR_CONTINUE;
}
/* if ret == SOCKET_ERROR, just return failure and log the errno */
```

@lws-team would this be a satisfactory switch to avoid calling `connect` as well as fix the spinning issue when there is not actually a server waiting for the connect?

This has fixed my issues but would like some input. I know you mentioned using a sul in [#2700 ](https://github.com/warmcat/libwebsockets/issues/2700) to recheck every 10 ms but I feel like this solution does not change much and still fixes the issue. I'm also not as familiar with lws sul as I would like to be.

lws-team:
> This way, in addition to avoiding consecutive calls to connect windows clients can actually wait for the connect call to succeed or fail.

Everything in lws must be nonblocking.  The nonblocking connect() call will succeed or fail, but if it succeeds, it is not telling you that it is connected.  It returned immediately and i's only telling you that the connection attempt is proceeding asynchronously.  Then what?  Spinning is objectionable but it does get the job done while letting other events be serviced while we are waiting, and ends with the connect success or fail.

It cannot simply block for many seconds and suppress everybody else's events on the event loop.

What will solve it with a minimum of objectionable behaviour is turn off POLLOUT wait in windows case, add a sul in the wsi to schedule checks the fate of the connection at intervals without sleeping or spinning.

If windows grew up and just supported Posix, this wouldn't be a problem... Posix connect() lets you set POLLOUT on your wait (you normal wait, poll() or epoll() or whatever) and signal POLLOUT event when the connect succeeds, and some other event if it fails.  So you just wait for a normal event to tell you what happened.

Mitch-Martinez:
I think my usage of "wait" is not really the best for what changes were suggested. I do understand your point that select still blocks for 1us and that a sul described like you just laid out would make much more sense so that no blocking has to happen at all. 

I guess I also misunderstood the spin issue. I've been mostly testing on an older branch (4.1) and should have updated to 4.3 to double check to make sure that the spin issue I'm seeing in 4.1 was the same talked about here. I've been experiencing an infinite spin on `lws_client_connect_check` when a server is not online. This is my biggest issue with connect because it never really lets me know that the connection failed or not. Just that the connection attempt is still proceeding and for some reason the connection timeout that was set is not triggering a way to stop attempting to connect.

Yeah I agree windows is really is annoying to work with. I wish it supported posix but that's just not the case right now.

I unfortunately do not feel equipped to handle adding a sul to schedule checks on the connect because I just don't have the spare time to familiarize myself with lws internals that much.

Thank you for your input.

Mitch-Martinez:
Select also does not need to block, it can return immediately if the timeout value is zero so it can poll just like connect was doing in this case.

I feel like these changes really boil down to wanting to get a more clear response of connection status and avoiding consecutive calls to `connect` because the winsock2 docs discourage it. For some reason my client was never getting any error codes from `connect` that indicated the connection failed and like I mentioned before, my client was spinning infinitely on `lws_client_connect_check`. `select` has not had these issues in my use case. 

I apologize again that I'm commenting on issues that really are not exactly about the issue I'm having (infinite spin for windows clients on `lws_client_connect_check` when servers are not available), but I feel like my issues are closely related to this one.

bogdad:
you got me interested : )
https://github.com/warmcat/libwebsockets/pull/2722

probably it misses something, but maybe as a start..

bogdad:
Closing this in favor of https://github.com/warmcat/libwebsockets/pull/2722

PR #2717, opened 2022-08-29
minimal-examples-lowlevel: fix misleading comment for the conceal value

lorenzo-arena:
Following the examples, I initially tried using `conceal_count` like:

```c
static const lws_retry_bo_t retry = {
    .retry_ms_table = backoff_ms,
    .retry_ms_table_count = LWS_ARRAY_SIZE(backoff_ms),

    .conceal_count = LWS_ARRAY_SIZE(backoff_ms) + 1,

    .secs_since_valid_ping = 3,
    .secs_since_valid_hangup = 10,

    .jitter_percent = 20,
};
```

expecting that the client would never give up. However it did after `LWS_ARRAY_SIZE(backoff_ms) + 1` attempts, as in the `lws_retry_get_delay_ms` method the conceal is calculated as 

```c
*conceal = (int)*ctry <= retry->conceal_count;
```

What is necessary instead is to set `conceal_count` to `0xffff` or `LWS_RETRY_CONCEAL_ALWAYS` if one wants the client to never give up.

### DIFF
--- a/include/libwebsockets/lws-callbacks.h
+++ b/include/libwebsockets/lws-callbacks.h
@@ -160,15 +160,6 @@ enum lws_callback_reasons {
 	 * the default callback action of returning 0 allows the client
 	 * certificates. */
 
-	LWS_CALLBACK_OPENSSL_CONTEXT_REQUIRES_PRIVATE_KEY	= 37,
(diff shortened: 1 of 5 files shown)


lws-team:
Thanks, it's pushed on main

PR #2727, opened 2022-09-17
delete_from_fd: assert if fd beyond the max_fds

ger202:
Signed-off-by: songbuhuang <544824346@qq.com>


lws-team:
Thanks, it's pushed on main.

Issue #2702, opened 2022-07-25
_lws_vhost_init_server returns 1 instead of -1 in case of an error


lukakudra:
When trying to start http server on an already taken port function `lws_create_vhost` does not return NULL.
While going through other functions that `lws_create_vhost` calls I concluded that the reason for this behavior is in function `_lws_vhost_init_server` located in file `server.c`.
In function `_lws_vhost_init_server` the following section of the code that starts at the line 481 is problematic:
```
a.af = AF_INET;
if (_lws_vhost_init_server_af(&a))
	return 1;
```
In my opinion the code section from above should return -1 instead of 1 or just return the call to function `_lws_vhost_init_server_af` like it does in the label `single` at the end of the function `_lws_vhost_init_server`.

This behavior can be reproduced using example program `lws-minimal-http-server` and `netcat`.
Before starting `lws-minimal-http-server` start `netcat` in listening mode on port 7681 with the following command:
```
nc -lk 7681
```
Then start `lws-minimal-http-server`. The output of `lws-minimal-http-server` will be the following:
```
$ ./lws-minimal-http-server
[2022/07/25 11:42:28:5763] U: LWS minimal http server | visit http://localhost:7681
[2022/07/25 11:42:28:5764] N: lws_create_context: LWS: 4.3.2-v4.3.2-12-g9bfca5b7, NET SRV H1 WS MbedTLS ConMon IPv6-absent
[2022/07/25 11:42:28:5764] N: __lws_lc_tag:  ++ [wsi|0|pipe] (1)
[2022/07/25 11:42:28:5765] N: __lws_lc_tag:  ++ [vh|0|netlink] (1)
[2022/07/25 11:42:28:5765] N: __lws_lc_tag:  ++ [vh|1|default||7681] (2)
[2022/07/25 11:42:28:5765] E: [null wsi]: lws_socket_bind: ERROR on binding fd 6 to port 7681 (-1 98)
```
The program `lws-minimal-http-server` does not exit with error code as it should, it keeps running although it didn't manage to bind the file descriptor to port 7681.

This behavior was not present in `v4.2-stable`, it is present in both `v4.3-stable` and current `main` branch and it was introduced in commit `cf2dbdc6a061427495aca16e2596d24af9c963bc`.

PR #2737, opened 2022-09-29
Always return result of _lws_vhost_init_server_af

tjwalton:
Fixes https://github.com/warmcat/libwebsockets/issues/2702.

I don't understand the full implications of the return value here, but this is consistent with the return value further down in the function, as noted in #2702.


SimonLuetkebohle:
In _lws_vhost_init_server
there is one further return 1 for the error "ipv6 not supported".
I think it would make sense to additionally change this into return -1 to be consistent.

The check if return value of the call of _lws_vhost_init_server < 0 in lib\core-net\vhost.c also assumes error value is -1, if I understand this part correctly.

lws-team:
Thanks forthe patch and sorry for the delay, it's pushed on main + v4.3-stable.

Issue #2744, opened 2022-10-19
Leaking fds when creating / destroying `lws_context` instances


ERobsham:
While investigating a port leak crash in software I maintain, I found that the source of the leak was due to the file descriptors returned by `pipe()` call in `lws_plat_pipe_create()` not getting cleaned up in the `lws_context_destroy()` call.

This can be easily reproduced by just recreating connections (running `lws_context_create()` /  `lws_context_destroy()` multiple times in the same program).

I created an example client that I was using to debug the leak / test a fix in [this branch](https://github.com/ERobsham/libwebsockets/tree/pipe_leak_example/minimal-examples-lowlevel/ws-client/minimal-ws-client-leak_example).

Will be putting in a PR for a potential fix shortly.

ERobsham:
PR for the Potential fix:   https://github.com/warmcat/libwebsockets/pull/2745

lws-team:
I pushed a patch that I think will solve this on v4.3-stable, v4.4-stable and main.

Here is the modified example I was using

https://warmcat.com/lws_fd_leak_example.c

PR #2745, opened 2022-10-19
Fixes leaking fds created by 'pipe()' call

ERobsham:
Updates the `lws_pt_destroy()` logic to ensure the pipe fds are closed if either `pt->dummy_pipe_fds[0]` or `pt->dummy_pipe_fds[1]` are still valid (previously was only checking `pt->dummy_pipe_fds[0]` which was resulting in the write fd to leak whenever a context was destroyed).

Potential fix for issue: https://github.com/warmcat/libwebsockets/issues/2744

### DIFF
--- a/lib/core/context.c
+++ b/lib/core/context.c
@@ -1797,9 +1797,9 @@ lws_pt_destroy(struct lws_context_per_thread *pt)
 		pt->pipe_wsi = NULL;
 	}
 
-	if (pt->dummy_pipe_fds[0]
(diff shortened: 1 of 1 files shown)


ERobsham:
@lws-team Just as an FYI, I discovered this while building off `4.2-stable`.   Not sure if its possible/makes sense to get this backported into there too?

lws-team:
Thanks forthe patch and sorry for the delay.  It's pushed on main + v4.3-stable.

PR #2792, opened 2022-12-12
Fix unused variable warnings when compiling without logs

orgads:


lws-team:
A version of this is pushed on main + v4.3-stable, thanks.

orgads:
Why did you keep the local `error` variables?

PR #2808, opened 2023-01-03
docs: set port to zero when using unix domain sockets

italo-parallelz:
When a server is set to listen to a unix domain socket, i.e. options field of `struct lws_context_creation_info` has the
`LWS_SERVER_OPTION_UNIX_SOCK` flag set, the port must be set to zero, otherwise the unix socket is never created.


lws-team:

PR #2824, opened 2023-01-27
Fix -Werror=enum-int-mismatch in lws_tls_server_abort_connection()

kraj:
GCC 13 is findinf this function signature mismatch.

../git/lib/tls/openssl/openssl-server.c:713:1: error: conflicting types for 'lws_tls_server_abort_connection' due to enum/integer mismatch; have 'int(struct lws *)' [-Werror=enum-int-mismatch]

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/lib/tls/openssl/openssl-server.c
+++ b/lib/tls/openssl/openssl-server.c
@@ -699,7 +699,7 @@ lws_tls_server_new_nonblocking(struct lws *wsi, lws_sockfd_type accept_fd)
 	return 0;
 }
 
-int
+enum lws_ssl_capable_status
 lws_tls_server_abort_connection(struct lws *wsi)
 {
 	if (wsi->tls.use_ssl)


kloczek:
Just hit this issue.
Is it possible to merge this PR and release new version?

a-w50:
Is there any reason, this PR hasn't been accepted yet?

corneliusclaussen:
Running into the same issue, please merge

heitbaum:
+1

oskarirauta:
Patch is incomplete, it addresses only openssl- fix should be made available for both openssl and mbedtls. Such patch is introduced in PR openwrt/packages#20949

I would recommend merging a fix for mbedtls as well in this commit.

lws-team:
The openssl patch seems to have been in both main / v4.3-stable for a while... I added Khem's change on both as well now.  Thanks and sorry for the long delay.

PR #2828, opened 2023-01-30
Fix pipe end socket fd left open leak on linux

martlaak-gmail:
lws_context_destroy() left pipe end fd open, causing fd leak visible in application open file desctiptions list.
When application was doing multiple context create/destroys cycles, it was finally causing "too many files open" error in application.


ERobsham:
FYI, [this PR](https://github.com/warmcat/libwebsockets/pull/2745) (and [issue](https://github.com/warmcat/libwebsockets/issues/2744)) is also aimed at plugging the fd leak.

PR #2832, opened 2023-02-07
Add guard for setrlimit

zoraaver:
LWS_HAVE_SYS_RESOURCE_H is correctly set by lws to prevent including sys/resource.h. However there is no corresponding guard for setrlimit (a function from that header). This causes a build failure on platforms which don't have sys/resource.h available.


zoraaver:
@lws-team any chance this could be merged? It's a pretty trivial change. Many thanks.

lws-team:
Thanks, sorry it took a long time.  It's pushed on main.

Issue #2855, opened 2023-03-23
lws_b64_decode_stateful truncates response


darhaywa:
Accurately sized output buffers have truncated decodes. 

Given the following base64 encoded string of 1708 characters, which decodes to 1279 bytes on https://cryptii.com/pipes/base64-to-binary, when passed through lws_b64_decode_string_len(inputString, 1708, outputBuffer, 1281) only 1278 bytes are returned. This continues until the output buffer size of 1283 is provided.

[...]

PR #2856, opened 2023-03-23
lws_b64_decode_stateful truncates response

darhaywa:
Addresses issue #2855 by allowing the parsing of the final byte when there are at least 3 bytes remaining in the buffer.

For every 4 bytes of input, a maximum of 3 bytes of output are generated when decoding the base64 string. The buffer space, therefore, only requires an additional 3 bytes of space. The code checks for space in the buffer before adding null termination.


xguo-prestolabs:
I think this patch can be merged.

lws-team:
Thanks for the patch and sorry for the long delay, it's pushed on main / v4.3-stable.

Issue #2864, opened 2023-04-07
dbus: timeout removes role_dbus_timer from list and calls dbus_timeout_handle to re-free the timer


maxvankessel:
Hello Andy,

I might have found an issue, unsure if its related to an older version of libdbus.
Sending a dbus message to a not responding service on the bus leads to a timeout.

https://github.com/warmcat/libwebsockets/blob/e8eb7d6bd66dde2156dbcd37b0e9048ed4006320/lib/roles/dbus/dbus.c#L283
`lws_dbus_sul_cb` removes the timer on a timeout, and libdbus will call `lws_dbus_remove_timeout`,
This again will remove it from the list, but a detached head is detected.

> [2023/04/07 09:45:18:1603] N: lws_dbus_sul_cb: firing timer
[2023/04/07 09:45:18:1603] I: lws_dbus_remove_timeout: t 0x55ef89fc3b10, data 0x55ef89fbc960
[2023/04/07 09:45:18:1603] I: handle_dispatch_status: new dbus dispatch status: 0
[2023/04/07 09:45:18:1604] E: lws_dll2_is_detached: dll2 0x55ef89fc5bc0: detached but next (nil), prev 0x55ef89fc5
lwsws: /home/max/git/libwebsockets/lib/core/lws_dll2.c:50: lws_dll2_is_detached: Assertion `0' failed.

I think:
https://github.com/warmcat/libwebsockets/blob/e8eb7d6bd66dde2156dbcd37b0e9048ed4006320/lib/roles/dbus/dbus.c#L296
This can be removed safely, same as the `lws_free(rdt)` from the `lws_dbus_sul_cb`, this way libdbus will clean up.

Let me know if you want me to create a PR.

lws-team:
Yes it seems broken... a PR would be good although I won't be able to apply it immediately.

maxvankessel:
Fast response.

I will create a PR.

PR #2869, opened 2023-04-17
dbus: prevent double free timeout event

maxvankessel:
SUL callback triggers a dbus timeout handle, handle removes the linked list entry and free's the memory attached to the entry. Remove the code where the the entry was removed by the timer callback, which triggered a double free of the same linked list entry.

See issue: https://github.com/warmcat/libwebsockets/issues/2864


lws-team:

PR #2872, opened 2023-04-24
fix ws parser for non pmd packet with pmd ext enabled

edwardz-prestolabs:
PMDR_NOTHING_WE_SHOULD_DO handle logic is missing from lws_ws_client_rx_sm compared to lws_ws_rx_sm. If we change the binance example from url subscribed future stream to live subscribed spot stream, we will see a dead loop for spot live subscribe response. Quoting logic from lws_ws_rx_sm would fix it.


xguo-prestolabs:
Any issue with this patch?

PR #2902, opened 2023-06-05
mbedtls-server: Fix broken client verification

dddaniel:
This fixes clients being able to connect with a certicate that was not signed by the configured CA when SSL_VERIFY_FAIL_IF_NO_PEER_CERT is set.

The issue only appeared when a client connects via IP address directly and not use a hostname.

When the hostname was used to connect, the SNI 'callback lws_mbedtls_sni_cb' overwrote the invalid verfiy mode of
MBEDTLS_SSL_VERIFY_OPTIONAL with MBEDTLS_SSL_VERIFY_REQUIRED by calling SSL_set_SSL_CTX.


lws-team:
It's pushed thanks.

PR #2917, opened 2023-06-27
tls: fix ssl connection error in raw connection

johnnychen-prestolabs:


lws-team:
Just removing this stuff will break other things... can't you make it conditional on not being RAW?

johnnychen-prestolabs:
> Just removing this stuff will break other things... can't you make it conditional on not being RAW?

Only connecting with RAW method enters this path, so it doesn't affect other connecting methods. TLS creation will be triggered later in lws_client_connect_4_established().

johnnychen-prestolabs:
I don't understand why we create TLS at this phase, and only for "RAW" method. Is there any other way out to make raw connection with TLS work?

johnnychen-prestolabs:
I've changed the commit to make the change minimized and now it's conditional on not being RAW. I'm not an experienced engineer and this is my first commit to github. Could you help me check its correctness?

> Just removing this stuff will break other things... can't you make it conditional on not being RAW?

lws-team:

Issue #2739, opened 2022-10-08
android build error


calvin2021y:
build with from master
```sh
CMAKE_COLOR_MAKEFILE:BOOL=ON
LWS_LOGS_TIMESTAMP:BOOL=ON
LWS_LOG_TAG_LIFECYCLE:BOOL=ON
LWS_PLAT_ANDROID:BOOL=ON
LWS_ROLE_H1:BOOL=ON
LWS_ROLE_WS:BOOL=ON
LWS_SUPPRESS_DEPRECATED_API_WARNINGS:BOOL=ON
LWS_WITHOUT_DAEMONIZE:BOOL=ON
LWS_WITHOUT_EXTENSIONS:BOOL=ON
LWS_WITHOUT_TESTAPPS:BOOL=ON
LWS_WITHOUT_TEST_CLIENT:BOOL=ON
LWS_WITHOUT_TEST_PING:BOOL=ON
LWS_WITHOUT_TEST_SERVER:BOOL=ON
LWS_WITHOUT_TEST_SERVER_EXTPOLL:BOOL=ON
LWS_WITH_CONMON:BOOL=ON
LWS_WITH_DIR:BOOL=ON
LWS_WITH_GZINFLATE:BOOL=ON
LWS_WITH_HTTP2:BOOL=ON
LWS_WITH_HTTP_BASIC_AUTH:BOOL=ON
LWS_WITH_HTTP_UNCOMMON_HEADERS:BOOL=ON
LWS_WITH_LIBUV:BOOL=ON
LWS_WITH_NETWORK:BOOL=ON
LWS_WITH_SSL:BOOL=ON
LWS_WITH_STATIC:BOOL=ON
LWS_WITH_SYS_ASYNC_DNS:BOOL=ON
LWS_WITH_TLS_SESSIONS:BOOL=ON
LWS_WITH_UDP:BOOL=ON
LWS_WITH_ZLIB:BOOL=ON
```

get error:

```sh

[  1%] Creating build include dir
[  1%] Built target GENHDR
[  3%] Building C object lib/CMakeFiles/websockets.dir/plat/unix/unix-misc.c.o
[  5%] Building C object lib/CMakeFiles/websockets.dir/plat/unix/unix-caps.c.o
[  5%] Building C object lib/CMakeFiles/websockets.dir/plat/unix/unix-pipe.c.o
[  7%] Building C object lib/CMakeFiles/websockets.dir/plat/unix/unix-service.c.o
[...]

PR #2924, opened 2023-07-02
dlo: fix compiler complaints 'redefinition of typedef'

NoelGraf:
Fix for #2739.
I get the same error when the pedantic flag is enabled.


inobelar:
@NoelGraf Greetings! It was resolved in this issue https://github.com/warmcat/libwebsockets/issues/3002 & this commit https://github.com/warmcat/libwebsockets/commit/48e09ddf51ea014a60d666f8eb780cd801c0d4b5

PR #2935, opened 2023-07-20
Suggestion to add a README.md to the READMEs folder

VictorieeMan:
When new, it can take some time getting familiar with the libwebsocket repo, I found the readme files in the READMEs folder to be useful. The job of clicking on each file and then glossing over their content, takes some time and slows down finding what is useful at the moment for the developer.

I suggest adding a README.md file, using relative links as headlines to each file, and then pairing each entry with a little short paragraph abstract/summary to faster help devs get to the relevant Readme content for them.

I generated a file with all these relative links and added three summaries. There are some work needed for making all summaries, but that can be done over time by those who are most concerned. When done, I think it may be a helpful time saver. On GitHub as we all know the file named README.md in a folder, is automatically shown as the readme for that folder.


xguo-prestolabs (approved):

lws-team:
Thanks for the patch.  Atm, gitohashi (on libwebsockets.org) and github don't agree on relative URL formats in their markup, so this can only work on one or the other (github as it stands), so it's waiting to modify gitohashi to work exactly the same.

PR #2970, opened 2023-09-23
netlink: fix errant route delete on NEWADDR

mpb27:
NEWADDR was errantly deleting route entities causing a disconnect of clients under the route that was errantly removed.  NEWADDR can be issued on a DHCP lease renew, which does not change any of the routes.

Specific to my case, it happend to delete the local-link IPv6 route entry every time a DHCP IPv4 lease was renewed (every hour).


lws-team:
Thanks a lot, it's pushed.

PR #2977, opened 2023-10-09
Remove usage of malloc_usable_size

kpcyrd:
Arch Linux is [preparing](https://gitlab.archlinux.org/archlinux/rfcs/-/blob/master/rfcs/0017-increase-fortification-level.rst) switching to -D_FORTIFY_SOURCE=3, upon inspection of all packages I noticed libwebsockets uses a function called `malloc_usable_size`. This function is "for diagnostic purposes only" and should not be used in release builds, as it's going to be considered a fatal memory violation with -D_FORTIFY_SOURCE=3 ([source](https://developers.redhat.com/articles/2022/09/17/gccs-new-fortification-level)).

From the [glibc documentation](https://man7.org/linux/man-pages/man3/malloc_usable_size.3.html):

> The value returned by malloc_usable_size() may be greater than
> the requested size of the allocation because of alignment and
> minimum size constraints.  Although the excess bytes can be
> overwritten by the application without ill effects, this is not
> good programming practice: the number of excess bytes in an
> allocation depends on the underlying implementation.
>
> The main use of this function is for debugging and introspection.

This patch deletes all code paths that depend on `LWS_HAVE_MALLOC_USABLE_SIZE` to be true, and `LWS_HAVE_MALLOC_USABLE_SIZE` itself.

Unfortunately this feature was somewhat exposed through `lws_get_allocated_heap`, since the consuming developer of this library might not have taken into account for `LWS_HAVE_MALLOC_USABLE_SIZE` to be false, so there might not be proper error handling in place for `return 0`, so this patch also removes this function entirely.

It may be possible to unconditionally do `allocated += adj;` in `_realloc` and `allocated = 0` in case of a free. `lws_get_allocated_heap` could then continue to exist as simply `return alloc;`, but I didn't understand the libwebsocket allocator code well enough to say for sure this is accurate.

Thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -604,10 +604,6 @@ CHECK_C_SOURCE_COMPILES(
 	"#include <malloc.h>
 	int main(int argc, char **argv) { return malloc_trim(0); }
 	" LWS_HAVE_MALLOC_TRIM)
-CHECK_C_SOURCE_COMPILES(
-	"#include <malloc.h>
-	int main(int argc, char **argv) { return (int)malloc_usable_size((void *)0); }
-	" LWS_HAVE_MALLOC_USABLE_SIZE)
 
 if (PICO_SDK_PATH)
 	set(CMAKE_REQUIRED_DEFINITIONS "-Dxxexit=exit")
--- a/cmake/lws_config.h.in
+++ b/cmake/lws_config.h.in
@@ -65,7 +65,6 @@
 #cmakedefine LWS_HAVE_HMAC_CTX_new
 #cmakedefine LWS_HAVE_MALLOC_H
 #cmakedefine LWS_HAVE_MALLOC_TRIM
-#cmakedefine LWS_HAVE_MALLOC_USABLE_SIZE
 #cmakedefine LWS_HAVE_mbedtls_md_setup
 #cmakedefine LWS_HAVE_mbedtls_net_init
 #cmakedefine LWS_HAVE_mbedtls_rsa_complete
--- a/include/libwebsockets/lws-misc.h
+++ b/include/libwebsockets/lws-misc.h
@@ -857,23 +857,6 @@ lws_dir_glob_cb(const char *dirpath, void *user, struct lws_dir_entry *lde);
 #endif
 
 /**
- * lws_get_allocated_heap() - if the platform supports it, returns amount of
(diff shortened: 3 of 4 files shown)


lws-team:
Thanks.  But what's up with just letting cmake fail the existence test for malloc_usable_size() instead (it should fail cleanly and nonfatally if -D_FORTIFY_SOURCE=3 is set when it runs the detection test for it).

Then lws will act like the api is not available in the libc and all well isn't it?

kpcyrd:
The function is still around and can be called, but writing beyond the amount of bytes your program has requested is going to cause a crash:

Consider this program:

```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <malloc.h>

int main() {
    char *zero = malloc(1024);
    if (zero == NULL) {
        printf("Memory allocation failed.\n");
        return 1;
    }
    memset(zero, 0, 1024);

    char *ptr = malloc(100);
    if (ptr == NULL) {
        printf("Memory allocation failed.\n");
        return 1;
    }

    size_t size = malloc_usable_size(ptr);
    printf("Usable size of allocated memory: %zu bytes\n", size);

    memcpy(ptr, zero, size);

    printf("Done!\n");
    free(ptr);

    return 0;
}
```

Looks like this when executed:
```
% gcc -D_FORTIFY_SOURCE=3 -O3 -o test test.c
% ./test
Usable size of allocated memory: 104 bytes
*** buffer overflow detected ***: terminated
[2]    138572 IOT instruction (core dumped)  ./test
% 
```

The program is still executing after `malloc_usable_size` has been called, but when `memcpy` attempts to write 104 bytes into the 100 byte sized buffer the program crashes fatally.

lws-team:
lws doesn't do anything like your test program above with the api.

The api is only used to keep a running estimate of how much heap is dedicated to lws in the case you composed lws into something larger, which is the typical case.  Nothing uses it as a pointer offset like your code above.  It's helpful to use it mainly when the memory is freed, we only have the pointer to the allocated usable area, we don't otherwise know the size of the allocation at that point.  This lets us subtract the size of the allocation from our running total of our usage is all.

We also use it to add to the running total, so when we subtract the size gleaned fom the api call, the size will match.

AFAIK there is no danger to anything with this usage with or without FORTIFY_SOURCE=3.

Issue #2979, opened 2023-10-14
IPV6_PREFER_SRC_PUBLIC socket option for lws_client_connect_via_info()?


dylanetaft:
Libwebsockets doesn't offer a way to get at the socket fd prior to connect to set an option, does it?

I see lws_client_connect_info stuct has an iface option that can take an IP.   I don't think IPV6_PREFER_SRC_PUBLIC works on Windoes, just Posix\Linux.  

Windows, the standard is to try to identify the IP off enumerating network devices to the best I can tell if you do not want to use the IPV6 privacy address outbound.  

I want to connect outbound on a SLAAC IPV6 configured host with privacy address using the public network IP. 

I am going down a rabbit hole of using libnl to grab the kernel's route table, the default route, and then the network interface, and pulling addressses off of it, and I can feed that to the iface field in the LWS struct I suppose. 

The other option I think would be to I think getsockname() off a socket established with that option to get the address the OS used with that hint, and then feed that to iface. 

I see the socket fd in the lws struct but accessing it is probably not a normal thing to do.

lws-team:
> I am going down a rabbit hole of using libnl to grab the kernel's route table,

Lws knows about netlink, and a single process can only open a single netlink socket at a time.  libnl has an incompatible license for lws so it has to do its own.  You can iterate network interfaces in a kinda-standard way with getifaddrs() too.  But it sounds you just want lws to set a socket flag.

> Libwebsockets doesn't offer a way to get at the socket fd prior to connect to set an option, does it?

No but you can add it quicker than fiddling with libnl and send a patch.

enum lws_client_connect_ssl_connection_flags in include/libwebsockets/lws-client.h declares a load of option flags including ones to control IP options.  You can add one here (you probably want to use currently absent 1 << 15 one) for your feature, something like LCCSCF_IPV6_PREFER_PUBLIC.  You can set this flag in `client_creation_info.ssl_connection` in your code when creating the info struct for the client creation call (it uses that for historical reasons).

`lws_plat_set_socket_options_ip()` in the platform-specific code (ie, there are different implementations of it per platform) is called just after client socket creation.  You can test the flag in there (see the linux one in `lib/plat/unix/unix-sockets.c` for an already-existing example of testing for those flags) and set your socket option in there.  I don't know either what windows wants but you can do whatever that is in the implementation in `lib/plat/windows/windows-sockets.c` the same way.

dylanetaft:
[...]

PR #2980, opened 2023-10-14
Add support for RFC5014 for Linux

dylanetaft:
Issue #2979
Add support for a socket flag to prefer connecting with the IPv6 public IP vs temporary private for IPV6 SLAAC configured machines - Standard documented in RFC5014


dylanetaft:
Hold, it's not right, sorry.

lws-team:
Thanks, I modified it a bit to use CMake to discover if linux/ipv6.h is available to get rid of the magic constant and pushed it on main.

PR #2985, opened 2023-10-20
openssl: Properly report OpenSSL error in lws_tls_client_connect

audricschiltknecht:
In case of an SSL_ERROR_SSL in lws_tls_client_connect, the lws_ssl_get_error call was calling lws_tls_err_describe_clear which cleared the OpenSSL error from the stack. Thus, the tls.err_helper attribute was set to the default value from ERR_error_string_n, masking the actual OpenSSL error message from client code.


lws-team:
Thanks for finding that.  I pushed it on main + v4.3-stable.

PR #2987, opened 2023-10-23
openssl: Add lws ctx ref to client vhost's SSL_CTX

nkarstens:
Adds a reference to the libwebsockets context to the OpenSSL context used by the client vhost. This allows SSL info callbacks to work correctly for clients, like it currently does for servers.


lws-team:
Thanks a lot, it's pushed on main and v4.3-stable.

nkarstens:
> Thanks a lot, it's pushed on main and v4.3-stable.

@lws-team No problem, thanks for your work on this project!

PR #3010, opened 2023-11-18
append OpenSSL library (which is found by pkg-config) with an absolute path

zzblydia:
when environment exists pkg-config and has installed openssl library , i notice log "OpenSSL libraries: /usr/lib64/libssl.so;/usr/lib64/libcrypto.so;**ssl;crypto**".  the "ssl;crypto" is from variable PC_OPENSSL_LIBRARIES.

in some platforms(e.g. windows with mingw, pkg-confg, openssl or suse12sp5 with pkg-config, openssl), it sometimes lead config error that it said HMAC_CTX_new not found.  when error occurs, i tried to delete line "list(APPEND OPENSSL_LIBRARIES ${PC_OPENSSL_LIBRARIES})" and found it worked.

so i think it's better to replace PC_OPENSSL_LIBRARIES with PC_OPENSSL_LINK_LIBRARIES to link library(which is found by pkg-config) with absolute path.

i test the solution on windows with clion and it passed.


lws-team:
Thanks, I pushed it on main + v4.3-stable.

PR #3025, opened 2023-12-08
fix build for iOS with CMake 3.28

gdevillele:
Hello there,

I needed to apply those few changes to be able to compile the lib for iOS using latest CMake 3.28.
I figured I would submit them here.

Changes to `cmake_minimum_required` could be removed.
It builds without them, but CMake prints warnings.

Thank you. 🙂

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,7 +25,7 @@
 if (PICO_SDK_PATH)
 	cmake_minimum_required(VERSION 3.13)
 else()
-	cmake_minimum_required(VERSION 2.8.12)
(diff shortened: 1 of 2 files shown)


lws-team:
Thanks... I also get the warnings on rocky with newer cmake, I added a patch to mass change the minimum cmake to 3.5... this has its own problems since you need then to use 3.5 even on older distros.  But it seems inevitable if newer cmake will explicitly fault trying to remain compatible with older cmake.

For main, I added your patch on top of that so with the cmake minimum already updated.

PR #3027, opened 2023-12-12
Fix prototype mismatch in private-network.h

Dolphindalt:
The implementation of `lws_tls_server_abort_connection` is defined as below.

```c
#if defined(LWS_AMAZON_RTOS)
enum lws_ssl_capable_status
#else
int
#endif
lws_tls_server_abort_connection(struct lws *wsi)
{
...
}
```

But in the header file, it is defined as the following.

```c
enum lws_ssl_capable_status
lws_tls_server_abort_connection(struct lws *wsi);
```

Compiling with a aarch64-none-linux-gnu compiler, the following error is produced.

image

This error does not appear when compiling with arm-linux-gnueabihf compiler.

The proposed change simply updates the prototype in the header to match the body. This resolves the compilation issue.

### DIFF
--- a/lib/tls/private-network.h
+++ b/lib/tls/private-network.h
@@ -167,7 +167,11 @@ lws_tls_server_new_nonblocking(struct lws *wsi, lws_sockfd_type accept_fd);
 enum lws_ssl_capable_status
 lws_tls_server_accept(struct lws *wsi);
 
+#if defined(LWS_AMAZON_RTOS)
 enum lws_ssl_capable_status
+#else
+int
+#endif
 lws_tls_server_abort_connection(struct lws *wsi);
 
 enum lws_ssl_capable_status


lws-team:
This was a bit confusing... main does not look like how you described.  In fact I solved this in 2020 the other way, by just using the enum for everything.

```
5c7b5af92 (Andy Green      2020-09-06 11:46:25 +0100 170) enum lws_ssl_capable_status
84a57540a (Andy Green      2019-01-13 06:58:21 +0800 171) lws_tls_server_abort_connection(struct lws *wsi);
```

It happened long enough ago, that on main branch, which is what the PR is trying to patch, and on v4.3-stable, there's no need for this patch.  And anything else is too old to worry about.  So you should probably upgrade to at least v4.3-stable.

Dolphindalt:
I see. Thank you for the response.

PR #3029, opened 2023-12-19
Fixing digest auth challenge parsing

simon0356:
Fix Digest auth challenge parsing.

Tested with Nginx and [nginx-http-auth-digest](https://github.com/atomx/nginx-http-auth-digest )

Digest auth parsing is working, but the new connection attempt with challenge response still not working.

I will look toward this issue.

### DIFF
--- a/include/libwebsockets/lws-tokenize.h
+++ b/include/libwebsockets/lws-tokenize.h
@@ -99,7 +99,7 @@ typedef enum {
 } lws_tokenize_state;
 
 typedef struct lws_tokenize {
-	char collect[128]; /* token length limit */
+	char collect[256]; /* token length limit */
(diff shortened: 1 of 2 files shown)


simon0356:
On the second commit, it is working on nginx with auth digest, i have not tried with TLS so far.
Valgrind report no memory leak, but i wonder if in some case tmp_digest could be lost, I will need advice on this.
Next step is to try with TLS and WS.

Simon

lws-team:
Thanks for the improvements.  I squashed the changes and split out the tokenizer one (it shouldn't really affect anything) and pushed on main... hopefully this should be the same as what you sent.

simon0356:
Thanks Andy

PR #3041, opened 2024-01-08
lecp: fix format_scan function for numeric longer than 2 digits and negative numbers

Ddom952:
For printing cbor map and array elements `format_scan` function is used to determine number of elements and two issue spotted regarding this function:
* not properly counting numeric constant longer than 2 digits e.g. `123` that result in LWS_LECPCTX_RET_FAIL and `odd map args` log error
* not properly counting negative const numeric that can cause LWS_LECPCTX_RET_FAIL and `odd map args` log error or not expected cbor with array

Changes:
* pop all digits if inside numeric
* all `-` case when assuming numeric
* add 4 write tests to api-test-lecp for to cover const numeric longer than 2 digits and negative const numeric inside map and array

### DIFF
--- a/lib/misc/lecp.c
+++ b/lib/misc/lecp.c
@@ -985,7 +985,7 @@ format_scan(const char *fmt)
 		}
 
 		if (numeric) {
-			if (*fmt >= '0' && *fmt <= '9')
+			while (*fmt >= '0' && *fmt <= '9')
 				fmt++;
 			numeric = 0;
 			if (*fmt != '(')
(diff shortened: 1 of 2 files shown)


lws-team:
Thanks a lot, it's pushed on main.

Ddom952:
Are there any plans to push it to v4.3-stable?

lws-team:
Sure, I just pushed it there too.

Ddom952:
Thanks

PR #3077, opened 2024-02-28
allow reserved bits and opcode, should close #3076

### LINKED ISSUE #3076: feature request: extension allow use rsv and opcode
In websocket protocol, there are 3 RSV bits, and 0x3-0x7, 0xB-0xF reserved frames.

   RSV1, RSV2, RSV3:  1 bit each

      MUST be 0 unless an extension is negotiated that defines meanings
      for non-zero values.  If a nonzero value is received and none of
      the negotiated extensions defines the meaning of such a nonzero
      value, the receiving endpoint MUST _Fail the WebSocket
      Connection_.

   Opcode:  4 bits

      Defines the interpretation of the "Payload data".  If an unknown
      opcode is received, the receiving endpoint MUST _Fail the
      WebSocket Connection_.  The following values are defined.

      *  %x3-7 are reserved for further non-control frames

      *  %xB-F are reserved for further control frames

libwesockets may allow custom extensions using these reserved bits and opcode.

- update rsv and opcode, it can be done in `LWS_EXT_CB_PACKET_TX_PRESEND`, however, it's only available if `LWS_WITHOUT_EXTENSIONS=OFF`, and this is something like only for `extension-permessage-deflate`.
- read rsv and opcode, currently there is `lws_get_reserved_bits`, no `lws_get_opcode`. Even is build with extension, reserved opcode would cause `LWS_CLOSE_STATUS_PROTOCOL_ERR`.

So, it maybe a way for libwebsockets provide such way for custom extensions, which means only for update / access rsv and opcode.

liudongmiao:
Provide a way for custom extensions to read and validate reserved bits and opcode.

For send reserved bits and opcode, extension can use `LWS_WRITE_RAW` directly.

### DIFF
--- a/include/libwebsockets/lws-client.h
+++ b/include/libwebsockets/lws-client.h
@@ -235,6 +235,18 @@ struct lws_client_connect_info {
 	 * wsi-specific logs into their own stream or file.
 	 */
 
+#if defined(LWS_ROLE_WS)
+	uint8_t		allow_reserved_bits;
+	/**< non-zero to allow reserved bits. You can get it by `lws_get_reserved_bits`.
+	 * Note: default zero means close the websocket connection for non-zero rsv.
+	 */
+
+	uint8_t		allow_unknown_opcode;
+	/**< non-zeo to allow unknown opcode. You can get it by `lws_get_opcode`.
+	 * None: default zero means close the websocket connection for unknown opcode.
+	 */
+#endif
+
 	/* Add new things just above here ---^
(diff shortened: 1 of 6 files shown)


lws-team:
Thanks, Released branches can't introduce new apis after the fact, so this is pushed on main.

liudongmiao:
@lws-team I'd like to say sorry. In the end, I implement a minimal websocket in our project without using lws.
Please consider to revert this pr.

However, from the api view, it may be included in lws.

PR #3082, opened 2024-03-03
fix build for wolfssl

liudongmiao:
For `wolfSSL_CTX_use_PrivateKey_buffer`, the signature is always `long`.
```c
int wolfSSL_CTX_use_PrivateKey_buffer(WOLFSSL_CTX*, const unsigned char*, long, int);
```

For `wolfssl`, there is no `extra_certs` like `boringssl`.

This patch works for main branch too, however need a cherry-pick.

// Is there any CI for this project?


lws-team:
Thanks, I pushed your patch for this on main + v4.3-stable.

PR #3084, opened 2024-03-05
make raw-skt don't read more than rx_buffer_size

liudongmiao:
When other roles like ws reading data, they follow `rx_buffer_size`, then fallback to context's `pt_serv_buf_size`. However, `raw-skt` don't follow `rx_buffer_size`, always use 0, then fallback to `pt_serv_buf_size`.

This PR make `raw-skt` don't read more than `rx_buffer_size` data.


lws-team:
Thanks, I pushed this on main + v4.3-stable

PR #3100, opened 2024-03-26
Fix lws_callback_all_protocol and some others

titofra:
Hello maintainers,
In `lws_callback_all_protocol`, you were checking if the current protocol is the targeted one by comparing its pointer to the pointer of the targeted protocol. However, this approach doesn't work as `wsi->a.protocol` is a copy of the protocols (see [lws_create_context() (context.c#L675)](https://github.com/warmcat/libwebsockets/blob/main/lib/core/context.c#L675)).
Instead, it is better to compare callback functions and names as you did there [pollfd.c#L646](https://github.com/warmcat/libwebsockets/blob/main/lib/core-net/pollfd.c#L646).
I noticed that some functions were using the same method to compare protocols, so I have fixed some of them in the meantime.


s-devine:
Also running into this issue.

lws-team:
Thanks for the patch, it's pushed on main with some small changes, please check it still works for you.

PR #3109, opened 2024-04-12
Fixed. Crash on client gateway multi-threaded model

yetyongjin:


lws-team:
Thanks... I can see you did a nice job adding tsi to the client pieces... I was only expecting people to care about SMP for server case, but of course the server case can easily also want to make clients.  And you fixed the hanging sul on client connect, that's a great fix SMP or not.  I pushed these pieces on main and v4.3-stable.

But I think whatever went wrong with the logging, taking the context mutex in the logging is likely to deadlock either now or later since it may already be held by the outer caller or its ancestors.  Generally, `__` prepend is indicating you should call it already owning mutexes, with the distinction so the `__` version won't try to acquire the locks since you must hold them.  So I think (without knowing exactly what sent you down that path) the right fix may probably be in the caller or its ancestors acquiring the context mutex, plus or minus how long it would have to hold it.

yetyongjin:
I changed (__lws_change_pollfd(w, 0, LWS_POLLOUT)) to lws_change_pollfd(w, 0, LWS_POLLOUT) in the lws_callback_on_writable() call. The consideration is: lws_callback_on_writable is an API called by the application layer. I do not want the application layer to care about the internal implementation of the library. Calling the locked version lws_change_pollfd at the API entry can ensure thread safety. Because during my testing, a crash was captured here. Of course, it can be choose to lock it at the application layer. I still hope that the application layer will not care about the details of the library and can directly call the lws_callback_on_writable() API in an external thread.

Regarding the locking of records, I think your concerns are justified. I locked here because I caught a wrong pointer access in the log list. Since my application only has the client side. If it is a mix of server and client, I think the risk of deadlock is really high with this change. I think a more appropriate approach would be to ensure thread safety when connecting and closing connections, but I'm not sure where to modify it.

lws-team:
> The consideration is: lws_callback_on_writable is an API called by the application layer.

I see... that makes sense... no the user code shouldn't need to know about internal lws locking.  I guess we have to try it.

> I think a more appropriate approach would be to ensure thread safety when connecting and closing connections, but I'm not sure where to modify it.

There are some helpers in ./lib/core/private-lib-core.h aimed at this kind of puzzle, eg, `lws_context_assert_lock_held(context)`.  They will blow an assert with SMP build if execution reaches it and the related lock is not held by the current thread, you can then get a backtrace of its ancestors to know who could / should have taken the lock.

Issue #3110, opened 2024-04-13
assert crash in _lws_change_pollfd


exploman:
VERSON: 4.3.3

POSITION: In function `_lws_change_pollfd`
  `assert(wsi->position_in_fds_table < (int)pt->fds_count);`
![image](https://github.com/warmcat/libwebsockets/assets/33505392/7a3a67ee-530c-44da-bf4d-e473c5c3c9d4)

BACKGROUND: We use libwebsockets as a high performance clients. We only have one context. There is only one thread do lws_service. We use `cmake -DLWS_MAX_SMP=16` to use `pt mutex` to protect `lws_callback_on_writable`. Every connection has there own thread. So there are many threads call `lws_callback_on_writable` on different connections when they have something to write.  But in performance test, after a long time, always crash in `lws_callback_on_writable` -> `_lws_change_pollfd`  

REASON:  I check the log. And I found when we call `lws_callback_on_writable` on the latest connection, another connection is destroyed.
![image](https://github.com/warmcat/libwebsockets/assets/33505392/ca9ec624-e927-4afb-87ac-9c323cde874d)
  So I have check the code about `__remove_wsi_socket_from_fds`. I found, in the fds array, the destroyed element will be exchanged with the last element. I check the code about `__insert_wsi_socket_into_fds` I found, the latest connection will be inserted at the end of the fds array as the last element.  But this exchange is not atomic, so lock protection is required. There is already lock protection in `__remove_wsi_socket_from_fds`. But when we do assert in lws_callback_on_writable -> _lws_change_pollfd, there is no lock protection. 
![image](https://github.com/warmcat/libwebsockets/assets/33505392/5a60527d-7471-4d44-834d-718f119555b3)
![image](https://github.com/warmcat/libwebsockets/assets/33505392/88a9c0de-d568-41ff-9305-82ba47159606)

ADVANCE: I further checked the code of `_lws_change_pollfd` and found that this function is not thread safe enough.  But I also found some thread safe code. 
![image](https://github.com/warmcat/libwebsockets/assets/33505392/f2133261-eeb9-431b-ad52-dae3b60f2182)
  I'm not sure if lwss_callback_on_writable is thread safe now. I know we have a thread safe function `lws_cancel_service_pt`. But I still want to use `lws_callback_on_writable` directly. 
[...]

PR #3111, opened 2024-04-13
fix crash on _lws_change_pollfd assert

exploman:


exploman:
issue: https://github.com/warmcat/libwebsockets/issues/3110

PR #3123, opened 2024-04-24
mbedtls-client: Enable the version is correct

ThePassionate:
After mbedtls version 3.0, support for TLS 1.0 and 1.1 has been removed.
For details, please see https://github.com/Mbed-TLS/mbedtls/issues/4286. 
If use a new version of mbedtls, connection errors may occur due to version reasons.
So we hope it can be configured according to the supported version.


lws-team:
Thanks, this is pushed on main.

PR #3124, opened 2024-04-29
bind to interface using SO_BINDTODEVICE socket option on Linux platforms

siliconja:
Use the SO_BINDTODEVICE socket option on Linux platforms when binding lws sockets. Current behavior binds to an address on the interface when the interface is specified in `lws_socket_bind()`. This forces traffic over a certain interface, which can be useful for multi-homed systems.

This provides similar behavior to libcurl and c-ares libs. On Linux this requires elevated privileges, and failure to bind is non-fatal.


lws-team:
Thanks, it's pushed on main.

PR #3130, opened 2024-05-07
mbedtls: fix compile on mbedtls 3.6+

orangepizza:
they moved mbedtls_x509_get_name into internal zone, this declares it again in a file I looked appropriate to hold one
actual calling for offended function is at mbedtls_extension.c , but as mbedtls 3.6 is LTS so I don't think it'd burn until 4.0


lws-team:
Thanks a lot for the patch.

If this fixes it, can you check if this fixes it instead?

```
diff --git a/lib/tls/mbedtls/mbedtls-extensions.c b/lib/tls/mbedtls/mbedtls-extensions.c
index bf275eef..107119c5 100644
--- a/lib/tls/mbedtls/mbedtls-extensions.c
+++ b/lib/tls/mbedtls/mbedtls-extensions.c
@@ -250,7 +250,12 @@ lws_mbedtls_x509_parse_general_name(const mbedtls_x509_buf *name_buf,
                *p = *p - 2;
 
                rfc822Name.MBEDTLS_PRIVATE_V30_ONLY(next) = NULL;
+
+#if (MBEDTLS_VERSION_MAJOR == 3) && (MBEDTLS_VERSION_MINOR >= 6)
+               ret = MBEDTLS_PRIVATE(mbedtls_x509_get_name( p, end, &rfc822Name ));
+#else
                ret = mbedtls_x509_get_name( p, end, &rfc822Name );
+#endif
                if (ret) {
                        lws_x509_clean_name(&rfc822Name);
                        return ret;
```

orangepizza:
Not sure if it'd work, that macro exposes variables but doesn't look like exposed function would work: I know it's just a pointer, but still

lws-team:
This is the way the code dealt with the avalanche of private symbols in 3.0, it should work fine.

orangepizza:
@lws-team looks like MBEDTLS_private trick fails at linking, not finding /usr/bin/ld: /usr/local/lib/libwebsockets.so: undefined reference to `private_mbedtls_x509_get_name' 
think it need to define old one

lws-team:
OK... then how about this as well

```
diff --git a/lib/tls/mbedtls/private-lib-tls-mbedtls.h b/lib/tls/mbedtls/private-lib-tls-mbedtls.h
index 162d9726..509b5ed5 100644
--- a/lib/tls/mbedtls/private-lib-tls-mbedtls.h
+++ b/lib/tls/mbedtls/private-lib-tls-mbedtls.h
@@ -57,3 +57,9 @@ lws_tls_mbedtls_cert_info(mbedtls_x509_crt *x509, enum lws_tls_cert_info type,
 int
 lws_x509_get_crt_ext(mbedtls_x509_crt *crt, mbedtls_x509_buf *skid,
                     lws_mbedtls_x509_authority *akid);
+
+#if (MBEDTLS_VERSION_MAJOR == 3) && (MBEDTLS_VERSION_MINOR >= 6)
+       int MBEDTLS_PRIVATE(mbedtls_x509_get_name(unsigned char **p, const unsigned char *end,
+                                                 mbedtls_x509_name *cur));
+#endif
+
```

orangepizza:
that wouldn't work, MBEDTLS_PRIVATE mecro is defiend as #define MBEDTLS_PRIVATE(member) private_##member
but get name function was just moved to different header file but not actually renamed
https://github.com/Mbed-TLS/mbedtls/blob/v3.6.0/include/mbedtls/private_access.h

lws-team:
Oh well :-)  let's do that then

lws-team:
```
diff --git a/lib/tls/mbedtls/private-lib-tls-mbedtls.h b/lib/tls/mbedtls/private-lib-tls-mbedtls.h
index 162d9726..cd368f19 100644
--- a/lib/tls/mbedtls/private-lib-tls-mbedtls.h
+++ b/lib/tls/mbedtls/private-lib-tls-mbedtls.h
@@ -57,3 +57,9 @@ lws_tls_mbedtls_cert_info(mbedtls_x509_crt *x509, enum lws_tls_cert_info type,
 int
 lws_x509_get_crt_ext(mbedtls_x509_crt *crt, mbedtls_x509_buf *skid,
                     lws_mbedtls_x509_authority *akid);
+
+#if (MBEDTLS_VERSION_MAJOR == 3) && (MBEDTLS_VERSION_MINOR >= 6)
+       int mbedtls_x509_get_name(unsigned char **p, const unsigned char *end,
+                                                 mbedtls_x509_name *cur);
+#endif
+
```

PR #3139, opened 2024-05-11
Restore erroneous modified assertion

exploman:
Incorrect modifications on assert: https://github.com/warmcat/libwebsockets/commit/e82028320345fd1942096b35a7015f9ab5ac4931
I fix the incorrect modification on assert by "push -f" at Sat Apr 13, but may not work. So I create a new pull request.


lws-team:
Thanks, it's pushed on main

PR #3164, opened 2024-06-04
windows: fix check of result of CreateFileW

### LINKED ISSUE #2920: libwebsockets-4.3.2 wrong check of result of CreateFileW
In lib/plat/windows/windows-file.c in _lws_plat_file_open
CreateFileW is used, which returns -1 (INVALID_HANDLE_VALUE) if the function fails (https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew).
On main branch the result is checked correctly with
`ret == LWS_INVALID_FILE`
(LWS_INVALID_FILE is defined as INVALID_HANDLE_VALUE)
On branch 4.3.2 this was changed to
`ret == NULL`
which does not work correctly.

nys3b8tvbsc:
Fix #2920


lws-team:
Thanks, it's pushed on v4.3-stable

PR #3165, opened 2024-06-05
Fix pipelined HTTP client connection on freertos platform

simon0356:
Hi Andy,

I have successfully port an apps from linux platform to a freertos + lwip one.
Everything work like a charm but i only got an issue with a pipelined HTTPS Client connection.

On freertos once established i'm getting WRITEABLE callback but i didn't get the RECEIVE callback.
I have observed that calling lws_service_do_ripe_rxflow like in unix-service.c will generate a POLLIN event and fix the issue.

I didn't get the same issue on a piplined WS Client.

Is this somethings that have been done on other platform and missed on freertos ?

Thanks
Simon

### DIFF
--- a/lib/plat/freertos/freertos-service.c
+++ b/lib/plat/freertos/freertos-service.c
@@ -207,6 +207,7 @@ again:
 		if (m)
 			n--;
 	}
+	lws_service_do_ripe_rxflow(pt);
 
 	if (a)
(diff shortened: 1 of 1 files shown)


lws-team:
Thanks, it's pushed on main.

simon0356:
Thanks !

PR #3172, opened 2024-06-24
Fix ACME

jdarpinian:
Increase polling time and poll the "order" endpoint instead of the "finalize" endpoint. These changes are required for ACME to work with Let's Encrypt as of 2024.


lws-team:

PR #3217, opened 2024-08-29
AI_V4MAPPED macro is missing on some platforms, e.g. OpenBSD, NetBSD, etc.

leleliu008:


lws-team:
Thanks... these things can be a bit awkward if the token is actually an enum.  But at least on Linux, it seems to be a define.  So it's pushed on main.

leleliu008:
how to add a condition `!defined(__NetBSD__)`

Issue #3216, opened 2024-08-28
Obtaining SSL_R_NO_SHARED_CIPHER error after a connection attempt


AssemblyJohn:
Hello!

I am trying to obtain the SSL connection error ( I see it is queried here: https://github.com/warmcat/libwebsockets/blob/c980be7cf7a6c49b87ecf31c1a854b54982bdf3f/lib/tls/mbedtls/mbedtls-client.c#L264 ), specifically the SSL_R_NO_SHARED_CIPHER error (193). It would be ideal to be able to get any error on a connection fail. 

I've tried the following code with no success, the returned error is 0:

```
case LWS_CALLBACK_CLIENT_CONNECTION_ERROR: {
        std::cout << "CLIENT_CONNECTION_ERROR: " << (in ? reinterpret_cast<char*>(in) : "(null)");
        ERR_print_errors_fp(stderr);

        unsigned long err_code = ERR_get_error();
        if (err_code) {
            char err_buf[128];
            ERR_error_string_n(err_code, err_buf, sizeof(err_buf));
            std::cout << "OpenSSL error: " << err_buf;
        }      
}
```

Any ideas much appreciated!

lws-team:
I think if you want more details in the CCE string, you will have to patch lws core pieces to report it in the string.  It's not like you're not getting a CCE just you want more details IIUI.

AssemblyJohn:
As per: https://libwebsockets.org/lws-api-doc-main/html/group__usercb.html#ggad62860e19975ba4c4af401c3cdb6abf7ad8c6207b0c4e732f3d507f0fb79370e8
you can see that the error string is not always relevant to the problem, SSL errors a bit missing. 

I would like not necessarily a string but a way to extract the int error code. Is that possible without crating my own libwebsocket branch, or do I need to fork it?

lws-team:
lws is FOSS, you can propose a patch.

lws is agnostic about the tls library, eg, mbedtls or openssl, and this information either needs unifying into an lws enum that can describe the situation without exposing tls library-specific data, or put into the string that already claims to describe the problem (again, it shouldn't refer to mbedtls-specific enums or whatever in the string, just text like "no shared cipher").

I guess if mbedtls has more verbose CCE string, that's not the end of the world and if someone else cares for openssl case, they can do the same.

AssemblyJohn:
Understood. The agnostic enum seems to be a better idea for this purpose, will propose a patch.

lws-team:
[...]

PR #3223, opened 2024-09-10
Update for detailed SSL connection code

AssemblyJohn:
Updated code to spit out a relevant OpenSSL error when a connection fails. The return string should be pars-able to obtain the error described like: "tls: error:0A00010B:SSL routines::wrong version number"

Closes: https://github.com/warmcat/libwebsockets/issues/3216


lws-team:
Thanks, it's pushed on main.

PR #3229, opened 2024-10-01
Simplify error logging in windows-sockets.c

orgads:


lws-team:
errno / LWS_ERRNO is delicate because depending on what you call between it being set and being used, it can be modified.  For example, if you want to do what you're doing here, it will work or break depending on what other arguments are in the log call and their order.

In this case there are currently no other arguments before the use that can change errno before the use.  But it's wrong to assume that sampling errno early like this is just noise.

Anyway thanks for the patch, it's pushed.

PR #3231, opened 2024-10-02
lws-map: Do not assume include order

orgads:
This header uses types that are defined in other headers, so include libwebsockets.h explicitly.

### DIFF
--- a/include/libwebsockets/lws-map.h
+++ b/include/libwebsockets/lws-map.h
@@ -42,6 +42,8 @@
  */
 //@{
 
+#include "../libwebsockets.h"
+
 typedef struct lws_map lws_map_t;
 struct lws_map_item;
 


lws-team:
Why does adding this include "fix duplicate typedef"?  What did the original warning or error look like?

orgads:
Sorry, my change was based on 4.3.3. This was already fixed in 8c8a2a5be3f7583a2678725ed6067976512ceca5 and b35128118e84be7ea7b56e50c3f16dddb8a365ac. The added include was just for correctness. I'll reword this commit and let you choose if you want to merge it.

orgads:
Done

lws-team:
This doesn't make any sense to me... including <libwebsockets.h> brings in lws-map.h in the first place.  You should just be including libwebsockets.h.

orgads:
Ok

PR #3233, opened 2024-10-02
tls-sessions: Pass correct type for printf %u

orgads:


lws-team:
This doesn't actually do what you think it does.  The format %u is defined to be an "unsigned int", and the size of an int is something the toolchain can decide.  For 64-bit, some like it to remain 4 bytes but other toolchains have it as 8 bytes.

The only correct casts for these are to the same type as the format, ie, unsigned int for this case, otherwise you create something that only builds correctly on your toolchain.

orgads:
Fixed, thanks.

lws-team:
Thanks, it's pushed on main + v4.3-stable

PR #3238, opened 2024-10-06
win32port: Convert all files to LF

orgads:
There is no reason to store source files with CRLF. If conversion is needed, Git should be configured with core.autocrlf.


lws-team:
Thanks for taking the time to make the patch.  There is a reason... this is how the files came to exist in the tree organically.  And although I can see it can have some value to normalize them, it is a massive churn, literally changing every line of all the win32 native pieces.  AFAIK the tools on both linux and windows side should maintain whatever line endings the file started with OK.  So I prefer to pass on this one.

orgads:
I was hit by this when I tried to apply a patch file that was converted to LF, and it failed because of mismatching line-endings.

Having files with CRLF means that if you create a patch file (with `git format-patch` or just `git diff`), the patch file will have mixed LF and CRLF line-endings, and it must be kept unchanged.

lws-team:
You can use `-l` with patch to make it blind to this.  These files were added years ago and have not caused trouble... I don't disagree all things being the same it's better to have a single kind of line ending, but since the native line ending differs on windows, changing them will create the opposite problem of CRLF leaking on them from windows tools anyway.  So I think it is better to leave them be.

orgads:
Git for Windows sets `core.autocrlf` by default, which converts all the text files to CRLF anyway. But I don't care that much. I'm fine with closing this.

PR #3239, opened 2024-10-07
improve net/ethernet.h header detection

snar:
On BSD systems, `<net/ethernet.h>` shall be included together with `<sys/types.h>` else compilation fails.


lws-team:
Thanks, it's pushed on main

PR #3248, opened 2024-10-20
Allow all HTTP method to be pipelined

simon0356:
Hello Andy,

I was trying to pipeline a HTTP Head request right after a GET and noticed on wireshark that the TCP connection  wasn't pipelined.

It seem that only GET, POST and PUT method are allowed to be pipelined, is there any reason ? Or just was not impleted so far ? 
I propose you the following patch to permit pipeline on all HTTP method.

I also noticed that the HTTP Head method as no LWS_CALLBACK_COMPLETED_CLIENT_HTTP callback, i propose a way to signal it, please let me known if somethings is wrong.

Thanks 

Simon

### DIFF
--- a/lib/core-net/client/connect2.c
+++ b/lib/core-net/client/connect2.c
@@ -196,10 +196,8 @@ lws_client_connect_2_dnsreq(struct lws *wsi)
 	}
 
 	/* only pipeline things we associate with being a stream */
-
-	if (meth && strcmp(meth, "RAW") && strcmp(meth, "GET") &&
-		    strcmp(meth, "POST") && strcmp(meth, "PUT") &&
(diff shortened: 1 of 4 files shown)


lws-team (commented):
This is no good, since it's perfectly legal for lws_hdr_simple_ptr() to return NULL

simon0356:
You right, I have push a new commit to check it before using strcmp which should work with C operator precedence.
The if condition is pretty huge now, maybe should I handle that case right after this one, instead of adding condition in it.
I have not tested yet, i will let you known.
Thanks !

lws-team:
Thanks, these should be pushed on main now.

simon0356:
Thanks ! Have a good day !

PR #3251, opened 2024-10-27
Fix another unused var error when logs are disabled

orgads:


lws-team:
Thanks, it's pushed on main

orgads:
Thanks! Can you please also backport to v4.3-stable?

lws-team:
Sorry for the delay, it's also pushed on v4.3-stable

orgads:
Thank you.

PR #3257, opened 2024-10-30
mbedtls-server: verify_option and auth_mode options should be valid

ikicha:
verify_options in mbedtls-server is slightly different from openssl-server. So made mbedtls-server have the same semantic about verify_option.

And also #2902 was the workaround fix for the 'weird' verify_option, so changed auth_mode to match verify_option's meaning.


lws-team:
Thanks, these should be pushed on main now

praveens3:
am using libwebsockets/4.3.3
wss server not verifying the client certificate, is this change available with 4.3.3?
Apologies, if this is not the right place to discuss this. help me/route to right place

Issue #3259, opened 2024-11-02
option to compile without libcap


johnhe4:
Would be nice to have a cmake option to compile without libcap since I don't need it for my use case.

I statically build libwebsockets into my dynamic lib. Nothing fancy in terms of websocket use cases, mostly text-based graphql subscriptions. Everything compiles and links fine for my dynamic lib, but when loading my lib into an app I get undefined references to libcap calls.

I prefer not to require consuming apps to add `-lcap` to their linker flags if they don't need to.

I created #3260

lws-team:
The PR should be pushed on main now.

PR #3260, opened 2024-11-02
Added option to disable libcap

johnhe4:
I threw this in because I don't need (or want AFAIK) libcap as a dependency. This assumes (rightly or wrongly) that libcap use cases are strictly linux, this can be tweaked as needed.
#3259


lws-team:
Thanks, this should be pushed on main now.

PR #3265, opened 2024-11-05
mqtt: fix MQTT client PINGREQ failure when using LIBUV as event driver.

baldwin-law:
When libuv is used as the underlying event-driven mechanism, the keepalive operation between the MQTT client and the server fails. The reason is that the keepalive operation should send a PINGREQ, but in fact, it is not sent. The problem lies in the fact that the modification of the underlying I-O-related events is not triggered, resulting in the inability to trigger I/O events. Through testing, modifying the mux_substream flag can solve this problem.


lws-team:
Thanks... does it make trouble to change the test to a runtime one like `if (lws_check_opt(context->options, LWS_SERVER_OPTION_LIBUV)) `?  The reason is it's possible to build lws with the LIBUV config, but actually select some other event loop at runtime.

lucas-law:
Of course, your approach is more elegant. I'll push the changes.

LeB134 (approved):

LeB134 (changes requested):
>

Issue #3177, opened 2024-07-03
detect memory leaks with mbedtls-3.5


xuchao0915170:
Hi，

I built lws-4.3-stable with mbedtls-3.5.2, ASAN option OPEN.
When I ran the lws-minimal-ws-client-ping demo，I found a memory leak warning below:
![image](https://github.com/warmcat/libwebsockets/assets/31056281/09b3dd00-6aa9-4681-9169-d5dc57c18882)
Then I rebuilt the lws demo with mbedtls-3.2.1, the warning is gone, so I guess this is a new version adapting problem.
I tried several ways to fix it but failed, so I report this issue instead, hope it will help~

lws-team:
Does the problem still exist on main branch?  It sounds like an internal allocation from mbedtls side that either didn't used to exist or was freed somewhere else before exit on earlier versions.

There's this but doesn't seem like a great fit https://github.com/Mbed-TLS/mbedtls/issues/3486

xuchao0915170:
I tried the newest develop version(mbedtls-3.6), the memory leak still exists:
![image](https://github.com/warmcat/libwebsockets/assets/31056281/fd342478-e202-4da7-b249-17e1d9959702)

lws-team:
This really seems like a change in mbedtls allocating something in its entropy function that now has to be freed in a step we didn't have to do on mbedtls earlier than 3.5.  You're better off asking mbedtls since without a crystal ball we can't know that later versions of mbedtls than exist at the time we integrated them are later going to start doing that.

They should tell you that "you now need to free those allocations by calling xxx" and then you can provide a patch on lws to do that (preferably included after checking MBEDTLS version requires it).

xuchao0915170:
This issue has been fixed by [https://github.com/warmcat/libwebsockets/commit/a71a62f51bec63b4eb1cd4a24fbb58ad3db23ad9](url)
, so I will close it. Thanks! @lws-team @baldwin-law

PR #3266, opened 2024-11-05
mbedtls: fix issue about memory leak

lucas-law:
https://github.com/warmcat/libwebsockets/issues/3177

### DIFF
--- a/lib/plat/unix/unix-init.c
+++ b/lib/plat/unix/unix-init.c
@@ -240,4 +240,9 @@ lws_plat_context_late_destroy(struct lws_context *context)
 		lwsl_err("ZERO RANDOM FD\n");
 	if (context->fd_random != LWS_INVALID_FILE)
 		close(context->fd_random);
+
+#if defined(LWS_WITH_MBEDTLS)
+	mbedtls_entropy_free(&context->mec);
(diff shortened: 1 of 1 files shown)


lws-team:
Thanks, it's pushed on v4.3-stable and main

simon0356:
Hello Andy, Lucas,

I noticed that the mbedtls freeing is not called at the same moment on FreeRTOS nor Unix plateform.
Is this a mistake or is there a good reason ? 
I could do a PR to put the free in lws_plat_context_late_destroy for freertos to do this in the same way.

Thanks
Simon

lws-team:
What code are we talking about with "mbedtls freeing"?  AFAIK we should only take care to destroy the entropy / rng stuff after any other mbedtls-specific allocations.  If it's happening that way already it's probably OK AFAIK,

simon0356:
I mean mbedtls_ctr_drbg_free and mbedtls_entropy_free are called in lws_plat_context_early_destroy for freertos and lws_plat_context_late_destroy for unix.

lws-team:
Right... yes given the only constraint is that we shouldn't do any mbedtls business after destroying the entropy stuff it's better to do as you suggest.  It's unlikely to break anything AFAIK doing it a bit later.

PR #3330, opened 2025-02-05
ESP-IDF: Removed explicit `esp_task_wdt_reset` call

vikramdattu:
Why is `esp_task_wdt_reset` needed?:
- If the task is running for longer without yielding, this leads to the task watchdog to trigger, then one should reset the task watchdog with the mentioned API.

Points under consideration while removing the call:
- `lws_service()` function call is not unyielding, meaning, it does wait on socket select. Even if this function is called in a loop, it should be the caller's responsibility to yield or call `esp_task_wdt_reset`.
-  Rightly so, the examples here do yield calling `taskYIELD`
- Calling `esp_task_wdt_reset` without calling `esp_task_wdt_add` has no real use and only leads benign errors.

### DIFF
--- a/lib/plat/freertos/freertos-service.c
+++ b/lib/plat/freertos/freertos-service.c
@@ -27,15 +27,8 @@
 int
 lws_plat_service(struct lws_context *context, int timeout_ms)
 {
-	int n = _lws_plat_service_tsi(context, timeout_ms, 0);
-
-#if !defined(LWS_AMAZON_RTOS) && defined(LWS_ESP_PLATFORM) && defined(CONFIG_ESP_INT_WDT)
-	esp_task_wdt_reset();
-#endif
-
-	return n;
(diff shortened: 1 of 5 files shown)


vikramdattu:
cc @Harshal5

glmfe:
@lws-team the libwebsockets library has been added as a component of esp-protocols, and merging this PR is crucial for us to move forward with adding other example codes, such as a server instance.

lws-team:
Can I suggest you take a more sensitive approach to what is already there than randomly ripping it out?

It's not like what is there is so golden it must last a thousand years, but, there are users of the existing code such as Amazon.  since it is enabled on AMAZON_FREERTOS currently, I think that should remain so for them.

If you want to avoid what it is doing, then use a cmake define to override, defeating it or somesuch, rather than act like esp is the only user of it.

PR #3332, opened 2025-02-07
Add systemd to the PkgConfig file if it's needed by the lib

mquinson:
Hello,

if I'm compiling an external project using the pkg config file, it fails with the following error on my machine:
```
FAILED: proxy 
: && /usr/bin/cc   CMakeFiles/proxy.dir/proxy.c.o -o proxy -L/opt/libwebsockets/lib -Wl,-rpath,/opt/libwebsockets/lib  -lm  -lwebsockets  -lcap && :
/usr/bin/ld: /opt/libwebsockets/lib/libwebsockets.so: undefined reference to `sd_is_socket_inet'
/usr/bin/ld: /opt/libwebsockets/lib/libwebsockets.so: undefined reference to `sd_listen_fds'
/usr/bin/ld: /opt/libwebsockets/lib/libwebsockets.so: undefined reference to `sd_is_socket_unix'
collect2: error: ld returned 1 exit status
```

This PR adds SystemD to the dependencies of the package if SystemD was detected on disk.

Thanks for this great project,
Mt

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -307,6 +307,13 @@ if (LWS_HAVE_LIBCAP)
 		set(lws_requires "libcap")
 	endif()
 endif()
+if (LWS_HAVE_SYSTEMD_H)
+	if (NOT lws_requires STREQUAL "")
+		set(lws_requires "${lws_requires},libsystemd")
+	else()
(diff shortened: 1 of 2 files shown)


lws-team:

PR #3335, opened 2025-02-12
mbedtls/ssl: free cert chain when mbedtls_client_preload_filepath ena…

ThePassionate:
hello, I have encountered some issues while using the mbedtls-wrapper for lws, and I would like to propose incorporating the following fixes:
1. fix the alignment in SSL_CTX_new
2. release memory if failed to parse cert chain in SSL_CTX_new
3. release cert chain during SSL_free call

I would appreciate it if these changes could be considered for inclusion in the project. Thank you for your time and consideration.
Best regards.


lws-team:
Thanks, this is pushed on main and v4.3-stable

PR #3354, opened 2025-03-26
adds LWS_MSVC_STATIC_RUNTIME option

### LINKED ISSUE #3353: option to statically link to MSVC runtime
I'm building a plugin, and (for the windows version) I am minimizing the number of DLLs required, including the MSVC runtime.
My plugin also links to protobuf, and they have a nifty option for building it and linking to the static runtime libraries, here: https://github.com/protocolbuffers/protobuf/blob/v28.3/CMakeLists.txt#L214
Note the pre-cmake-3.15 mechanism too.
Is that an option you would be interested in supporting?

gmabey:
closes #3353

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,10 @@ if (POLICY CMP0075)
 	cmake_policy(SET CMP0075 NEW)
 endif()
 
+if(POLICY CMP0091)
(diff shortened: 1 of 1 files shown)


lws-team:
Thanks, it's pushed on main

PR #3361, opened 2025-04-05
solaris/illumos build fix

jwntree:
In Solaris, the build fails in wol.c because solaris does not define ETHERADDRL instead it defines ETHER_ADDR_LEN.
Also, link with -lsocket needs unlike linux.

While building the tests, 
I have to add -Wno-conversion due to a build error related to FD_SET,
ctime_r is defined slightly different than Linux. I just added unset for solaris, but check using CHECK_C_SOURCE_COMPILES might be more better approach.


lws-team:
Thanks, it's pushed on main.

PR #3362, opened 2025-04-06
Allow applications to specify chroot and chdir for CGI mounts

ribes96:
When an application initializes an `lws_http_mount` for a CGI, it can now provide fields `cgi_chroot_path` and `cgi_wd` to change the root path and the working directory of the spawned child. If these fields are left NULL, the old behaviour is kept.

The CGI API now provides an additional `..._via_info` API that supports providing these fields. The old API remains there.

All examples have been updated so that they still compile with the new additions to `lws_http_mount` structure.

`lejp-conf.c` now understands keys `vhosts[].mounts[].cgi-chroot` and `vhosts[].mounts[].cgi-chdir` when reading a JSON config to initialize a mount.


lws-team:
Thanks, it's (all, hopefully) pushed on main.

PR #3363, opened 2025-04-07
Fix cgi QUERY_STRING key only field truncated

ribes96:
A request like 'http://host?abc' set the QUERY_STRING variable for CGIs to "ab" instead of "abc". This was because the code didn't account for key-only fields, only for key-value ones.


lws-team:

Issue #3368, opened 2025-04-15
Support building libwebsockets with AWS-LC


smittals2:
Hello,

I’m an engineer at AWS working on AWS Libcrypto (AWS-LC), an open-source cryptographic library maintained for AWS and their customers. We are committed to backwards compatibility. For this purpose we have CI jobs [here](https://github.com/aws/aws-lc/blob/main/.github/workflows/integrations.yml) asserting every change’s compatibility with many different open-source projects. We use these tests to catch compatibility regressions before they’re merged. We have already added libwebsockets to our CI [here](https://github.com/aws/aws-lc/blob/d1c1d72f3844442290a1f6d62e1cf5d3b8d12a33/tests/ci/cdk/cdk/codebuild/github_ci_integration_omnibus.yaml#L355). 

AWS-LC supports CPU-specific performance optimizations for AWS Graviton 2, [AWS Graviton 3,](https://github.com/aws/aws-lc/commit/ae87faf735c0241a115542b1c1022d125564bf55) and [Intel x86-64 with AVX-512 instructions](https://github.com/aws/aws-lc/blob/d4cecff8b3dd4584e2ba04f55073a4bd3289046a/crypto/fipsmodule/aes/asm/aesni-xts-avx512.pl). We’ve formally verified [a subset of](https://github.com/awslabs/aws-lc-verification#verified-code) AWS-LC’s cryptographic primitives, and continue to invest in expanding this coverage. AWS-LC has been [FIPS validated](https://github.com/aws/aws-lc/blob/0931fe2ff18ed4ad47473cbb8c11066e25fc26c5/crypto/fipsmodule/FIPS.md?plain=1) by NIST and we have 140-3 certificates for both [dynamic](https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/4759) and [static](https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/4816) builds. We would like to upstream support for AWS-LC into the mainline branch of libwebsockets. We believe that this would provide the best experience for users wishing to build libwebsockets against AWS-LC.

[...]

PR #3369, opened 2025-04-16
Add support to build with AWS-LC

smittals2:
Hello, 

I am opening this PR per issue #3368 with the changes required to build AWS-LC with libwebsockets. I took the liberty of making some additional changes for better readability - namely extracting repeated conditional logic into common macros.

Let me know if you have any feedback, thank you for your consideration!


lws-team (commented):
You shouldn't need LD_PRELOAD to make the built things work.  The tls library pieces should be included as part of the lws link.

lws-team (commented):
In test-client.c, you just remove the adaptation for different function signatures between the tls libs, without replacing it with a wrapper.  The different libs just randomly chose different types on some apis, sometimes dependent on lib version, without rhyme or reason.  Unfortunately this kind of adaptation is required, not just noise (and is case-by-case per api, so eg SSL_OPT_TYPE won't cover all cases).

smittals2:
> In test-client.c, you just remove the adaptation for different function signatures between the tls libs, without replacing it with a wrapper. The different libs just randomly chose different types on some apis, sometimes dependent on lib version, without rhyme or reason. Unfortunately this kind of adaptation is required, not just noise (and is case-by-case per api, so eg SSL_OPT_TYPE won't cover all cases).

Hi! Thanks for the feedback. Let me clarify the change.

Couple of contextual points:
1. `ERR_peek_error` and `ERR_get_error` return the earliest error on the stack, the former just reads, the latter reads and removes the error from the stack.
2. WolfSSL, LibreSSL, and OpenSSL all return an `unsigned long` value for `ERR_get_error` and `ERR_peek_error`. `ERR_error_string` in all three libraries also takes an `unsigned long` value for it's first parameter.
3. BoringSSL and AWS-LC both return `uint32_t` values for peeking and getting the error, and both take an `uint32_t` value as the first parameter to `ERR_error_string`.

So to accommodate these differences without having to handle casting with ifdefs, I moved `ERR_get_error()`'s call directly into `ERR_error_string`. The return value and parameter type will always match in all of the above TLS libraries. I used `ERR_peek_error` in place of where `ERR_get_error` was previously used so we can still store the error value for printing it later (Note: casting `unsigned long`/`uint32_t` to `int` as we do in test-client.c isn't a safe cast, I noticed in other places the error var is generally defined as `unsigned long`).

I have used this change in a couple of other places as well. If you feel this is more confusing than helpful, I can get rid of these. Let me know!

smittals2:
Hi @lws-team, just following up on this PR. Please let me know if you have any more feedback, thank you! :)

lws-team:
I pushed these patches on main.

I see the LD_PRELOAD stuff is coming from what was already in the README, I redid the sai tests for the other ssl-derived libs and this one, so they are installed in /usr/<libname> so they don't conflict... all of the dervied libs except wolfssl retain the include subdir "openssl" regardless of the lib name, so they can't otherwise coexist installed in the same tree.

I also looked into the genhash stuff and fixed the underlying causes, so all of them can build and test with genhash pieces now.

Currently ass the tests (including the aws-lc ones) are passing, see https://libwebsockets.org/sai/

PR #3378, opened 2025-04-29
passes *user to callback rather than NULL

rubienr:
Passes `*user` to callback instead of `NULL`.


lws-team:
Thanks, it's pushed on main.

PR #3388, opened 2025-05-14
adns: Report error code if callback destroyed wsi

mattfischer:


lws-team:
Pushed on main + v4.3-stable thanks

PR #3395, opened 2025-05-26
client-http: authorize LWS_...COMPLETED_CLIENT_HTTP to be called if status code is 204

simon0356:
Hello Andy,

I met an issue when trying to pipeline a new HTTP request after getting an HTTP response 204 (No content) which has no content-length header.
The LWS_CALLBACK_COMPLETED_CLIENT_HTTP is never called and then following HTTP request with same wsi is not calling  LWS_CALLBACK_CLIENT_APPEND_HANDSHAKE_HEADER.

This patch seem to fix the issue as i experiment it.

Thanks
Simon


lws-team:
Thanks... it makes sense it'd make trouble as it was.  It's pushed on main.

PR #3421, opened 2025-07-18
mbedtls-server: verify_option and auth_mode options should be valid

ikicha:
this patch is basically #3257 for 4.3-stable which fixes the client certificate issue.


lws-team:
Thanks, it's pushed on v4.3-stable.  Notice v4.4 is out in the last days and so v4.4-stable will get whatever attention is going rather than v4.3-stable.

PR #3423, opened 2025-07-22
Minor CMake maintenance

LecrisUT:
- Use higher-bound for the cmake policies. Simplifies the code an maintenance. Feel free to lower the higher bound if those CMake versions are not sufficiently tested for your standards. Might also need to propagate the upper-bound to the auxiliary files
- Move the `Check*` includes to after the language is enabled. Generally you want *all* `include` to happen after the `project`/`enable_language` since it can have implicit requirement of language enablement.
- Use `GNUInstallDirs` as baseline for install paths. This has been standardized since the CMake 3.0 days and distros and other package environments would make use of this to inject their own standard. This simplifies the life for them to not have to include a special setup for this package.
  - 3rd-party locations still use `CMAKE_INSTALL_*` variables since they are technically different from `LWS_INSTALL_LIB_DIR` (haven't seen an environment that does that though :shrug:)


lws-team:
Thanks, I pushed your patches on main and v4.4-stable.

lws-team:
I seem to have found a problem with these on an Ubuntu aarch64 box.

Prior to these patches, things were installed into /usr/local/lib/libwebsockets.so.[19,20]

After the patches, things are installed into /usr/local/lib64

After installing, an app consuming the installed lib happily for a long time dies in cmake, after I deleted the no longer up-to-date `/usr/local/lib/libwebsockets*`

```
CMake Error at /usr/local/lib/cmake/libwebsockets/LibwebsocketsTargets.cmake:93 (message):
  The imported target "websockets" references the file

     "/usr/local/lib/libwebsockets.a"

  but this file does not exist.  Possible reasons include:

  * The file was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and contained

     "/usr/local/lib/cmake/libwebsockets/LibwebsocketsTargets.cmake"

  but not all the files it references.

Call Stack (most recent call first):
  /usr/local/lib/cmake/libwebsockets/libwebsockets-config.cmake:16 (include)
  CMakeLists.txt:3 (find_package)
```

LecrisUT:
> After installing, an app consuming the installed lib happily for a long time dies in cmake, after I deleted the no longer up-to-date `/usr/local/lib/libwebsockets*`

You forgot to delete `/usr/local/lib/cmake/libwebsockets/` as well. That should be the only other relevant artifacts that conflict.

lws-team:
Oh that was it, thanks a lot for the quick reply.

PR #3441, opened 2025-08-14
lws/openssl-wrapper: Align ssl_handshake and openssl standard return values

ThePassionate:
https://docs.openssl.org/1.0.2/man3/SSL_do_handshake/#return-values
standard return code of ssl_handshake list following:
0:
The TLS/SSL handshake was not successful but was shut down controlled and by the specifications of the TLS/SSL protocol. Call SSL_get_error() with the return value ret to find out the reason.

1:
The TLS/SSL handshake was successfully completed, a TLS/SSL connection has been established.

<0:
The TLS/SSL handshake was not successful because a fatal error occurred either at the protocol level or a connection failure occurred. The shutdown was not clean. It can also occur of action is need to continue the operation for non-blocking BIOs. Call SSL_get_error() with the return value ret to find out the reason.

so ssl_error_read/write should return -1.
Also mapping ssl->err from mbedtls error to openssl error, not return mbedtls error directly.

### DIFF
--- a/lib/plat/windows/windows-spawn.c
+++ b/lib/plat/windows/windows-spawn.c
@@ -527,6 +527,9 @@ lws_spawn_piped(const struct lws_spawn_piped_info *i)
 		lws_sul_schedule(context, i->tsi, &lsp->sul,
 				 lws_spawn_timeout, i->timeout_us);
 
+       if (i->plsp)
+               *(i->plsp) = lsp;
+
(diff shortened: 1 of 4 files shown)


lws-team:
I notice that we change the return code meaning, but we don't adapt the caller.  Since mbedtls works today, are we sure that is a good idea?

PR #3495, opened 2025-11-12
MQTT connection error handling

johnhe4:
Playing with the MQTT client led to some incorrect error handling. For example, if the credentials were invalid, or the protocol was not supported by the server, the error log only showed `Parsing packet failed` and would not invoke the client callback. These changes attempt to address that.

### DIFF
--- a/lib/roles/mqtt/mqtt.c
+++ b/lib/roles/mqtt/mqtt.c
@@ -1029,17 +1029,27 @@ _lws_mqtt_rx_parser(struct lws *wsi, lws_mqtt_parser_t *par,
 			switch (par->conn_rc) {
 			case 0:
 				goto cmd_completion;
+			/* 3.1.1 errors [MQTT-3.2.3] */
 			case 1:
-			case 2:
-			case 3:
-			case 4:
-			case 5:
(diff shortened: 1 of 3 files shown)


lib/roles/mqtt/mqtt.c
```diff
@@ -1029,17 +1029,27 @@ _lws_mqtt_rx_parser(struct lws *wsi, lws_mqtt_parser_t *par,
...
+				goto send_reason_and_close;
 			case 3:
+				par->reason = LMQCP_REASON_SERVER_UNAVAILABLE;
+				goto send_reason_and_close;
 			case 4:
+				par->reason = LMQCP_REASON_BAD_CREDENTIALS;
+				goto send_reason_and_close;
 			case 5:
-				par->reason = LMQCP_REASON_UNSUPPORTED_PROTOCOL +
-						par->conn_rc - 1;
```
johnhe4: I'm guessing this logic used to work at one point? Maybe the MQTT error code enum was updated to support 5.0 errors and that's where this broke.

johnhe4: This approach should work for both 3.1.1 and 5.0

lws-team:
Thanks, it's pushed on main.

PR #3504, opened 2025-11-18
Minor edits and validations from static analyzer (Svacer)

chenuduss:
In PR some checks of return values ​​and  removal a useless check. All changes thanks to the findings of the Svacer


chenuduss:
So, is the edit in the commit 896b29d correct or not?

lws-team:
I already pushed a different solution a couple of days ago https://libwebsockets.org/git/libwebsockets/commit?id=0dfdcdf1716b81c4872c001cfac8e37496dc4d02

I pushed your other patches at that time but I am unsure about the checks on method and uri are in the right place.  Eg NULL method is later converted to GET after your check.

PR #3537, opened 2026-01-22
implement IANA cipher suite support for mbedtls backend 

pppaulpeter:
Implement parse_cipher_list_to_ids() to convert IANA names to mbedtls IDs
Update SSL_set_cipher_list() to apply ciphersuites immediately when active
Add cipher list application in server vhost initialization
Add capability guards for mbedtls session functions (compatible with 2.16.3)
Verified: cipher restrictions properly enforced during TLS handshake"


lws-team:
Thanks... as I mentioned I was hoping to avoid a migraine.

0) Was this done with the assistance of an LLM, if so which one?  It's not necessarily a problem, I am using Gemini 3.0 for some months, but I would like to know.

1) Please don't introduce using strtok() to lws.  There's a robust lws tokenizer "lws_tokenize" include/libwebsockets/lws-tokenize.h  that should be able to cope, if not, it's fairly easy to extend by flags.

2) Please don't push your development history.  Reverting something?  Fiddling with gitignore?  Nobody cares to know, just remove whatever it was from the sources, no need for another patch.  If your development is like that, squash the revert patch into the patch that added it so there's an antimatter type simplification explosion.

In order to remove all the churn (your delta was 13 patches...) I did a git diff from the basis point from main you used to your HEAD, this is a single patch capturing your changes that is a lot simpler than what you pushed.  Layered patches are good but if it just obscures the effect of your patch it is bad and I would much rather have one patch.

3) Rebase your patches on HEAD main before pushing.  When I applied the simplification patch, there were a handful of merge problems (even with -l) which I manually fixed up but one that implies your main basis and the current one are different enough.

4) What's the situation with "IETF" alg names that we are introducing here, and eg openssl or now schannel?  Openssl already liked them and we are just aligning mbedtls to use the same instead of nonstandard ones?

In summary: Please a) convert away from strtok to struct lws_tokenize, b) squash everything to one patch like I described above and c) push it on top of today's main branch, test then d) push again to your PR branch with that.

pppaulpeter:
> Thanks... as I mentioned I was hoping to avoid a migraine.
> 
> 0. Was this done with the assistance of an LLM, if so which one?  It's not necessarily a problem, I am using Gemini 3.0 for some months, but I would like to know.
> 1. Please don't introduce using strtok() to lws.  There's a robust lws tokenizer "lws_tokenize" include/libwebsockets/lws-tokenize.h  that should be able to cope, if not, it's fairly easy to extend by flags.
> 2. Please don't push your development history.  Reverting something?  Fiddling with gitignore?  Nobody cares to know, just remove whatever it was from the sources, no need for another patch.  If your development is like that, squash the revert patch into the patch that added it so there's an antimatter type simplification explosion.
> 
> In order to remove all the churn (your delta was 13 patches...) I did a git diff from the basis point from main you used to your HEAD, this is a single patch capturing your changes that is a lot simpler than what you pushed. Layered patches are good but if it just obscures the effect of your patch it is bad and I would much rather have one patch.
> 
> 3. Rebase your patches on HEAD main before pushing.  When I applied the simplification patch, there were a handful of merge problems (even with -l) which I manually fixed up but one that implies your main basis and the current one are different enough.
> 4. What's the situation with "IETF" alg names that we are introducing here, and eg openssl or now schannel?  Openssl already liked them and we are just aligning mbedtls to use the same instead of nonstandard ones?
> 
> In summary: Please a) convert away from strtok to struct lws_tokenize, b) squash everything to one patch like I described above and c) push it on top of today's main branch, test then d) push again to your PR branch with that.

yes, i use chatgpt5.2, claude,grok etc. i will check your valuable comment carefully and since this is my first git push for pull request, there must be a lot of problem.

PR #3566, opened 2026-04-07
parse_uri: handle '?' directly after authority per RFC 3986

### LINKED ISSUE #3521: The function lws_parse_uri has a defect when parsing the URL wss://xcx-login.yscq-wy.yscq.com?ip=s325-xcx-zfb.yscq-wy.yscq.com&port=8061.
The function lws_parse_uri has a defect when parsing the URL wss://xcx-login.yscq-wy.yscq.com?ip=s325-xcx-zfb.yscq-wy.yscq.com&port=8061.

pppaulpeter:
## Summary

- Fix `lws_parse_uri()` to recognize `?` as a delimiter between the authority and query components, per RFC 3986 §3.4
- URLs like `wss://host.com?key=val` previously returned the query string as part of the hostname; now correctly split into `ads="host.com"` and `path="?key=val"`
- Uses in-place memmove into the dead `://` space to NUL-terminate the hostname while preserving `?` in the path output

## Details

RFC 3986 allows the query component to appear directly after the authority with no path segment:

```
URI = scheme "://" authority [ "/" path ] [ "?" query ]
```

The existing code only stopped hostname/port scanning at `:` and `/`, so `?` was absorbed into the address field. This patch adds `?` as a stop condition in both the hostname and port scanning loops, then shifts the address one byte left to create room for a NUL terminator before `?`.

Tested with:
- Standard URLs: `https://host/path`, `http://host:8080/path`, `wss://host:9443/ws`
- Query-after-authority: `wss://host?k=v`, `http://host:3000?k=v`
- IPv6: `https://[::1]?debug=1`, `https://[::1]:9090?debug=1`
- Path+query: `http://host/path?k=v` (existing behaviour preserved)
- Unix sockets: `http://+/var/run/sock:/path` (existing behaviour preserved)
- Valgrind memcheck: 0 errors, 0 leaks

Fixes #3521

## Test plan

- [ ] Verify `wss://xcx-login.yscq-wy.yscq.com?ip=s325&port=8061` parses correctly
- [ ] Verify existing URL patterns are not affected
- [ ] Run `ctest` to confirm no regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lws-team:
Thanks.  Basically the original lws_parse_uri() was done 15 years ago and didn't take care about this case... if I had realized host.com?x=y was legal at the time, I wouldn't've done it like it is.

Instead I am adding a new patch to mark the old api as deprecated, and will implement a new api that creates the results on heap, accepts a const char * instead of needing a char * because it will delimit the string in-place, and has a destroy api to free the allocation.  It allocates a small struct + maximum of original URI length + 4.  This adapts all existing users inside lws to use the new api additionally.

https://libwebsockets.org/git/libwebsockets/commit?id=9d9a2dc75b5cfa3431ec073a14124a5ccc4de3d4

pppaulpeter:
Closing in favor of Andy's new heap-based API (9d9a2dc). Thanks for the better fix!

PR #3570, opened 2026-04-09
Fix strchr() and related functions for ISO C23 compliance

purofle:
Since glibc-2.43:

For ISO C23, the functions bsearch, memchr, strchr, strpbrk, strrchr, strstr, wcschr, wcspbrk, wcsrchr, wcsstr and wmemchr that return pointers into their input arrays now have definitions as macros that return a pointer to a const-qualified type when the input argument is a pointer to a const-qualified type.

The compilation test passed on my Glibc 2.43 system.


purofle:
merged https://github.com/warmcat/libwebsockets/commit/03afafa6b38f538231ffe0b967baea6ee2daef30

Issue #3403, opened 2025-06-11
Cross-build fail redefinition of 'struct in6_addr'


ribes96:
I used to be able to cross-compile libwebsockets using my toolchain. In fact, I can still compile latest stable version `libwebsockets-4.3.5`, but the build fails when compiling from the git tree:

~~~log
cc1: warning: command line option "-Wno-deprecated" is valid for C++/Java/ObjC++ but not for C
In file included from /usr/crosstool-ng/armv5-gcc-4.3.3-glibc-2.6.1-kernel-2.6.27/arm-unknown-linux-gnueabi/arm-unknown-linux-gnueabi//sys-root/usr/include/linux/ipv6.h:4,
                 from /home/arm/priv_git/libwebsockets/lib/plat/unix/unix-sockets.c:31:
/usr/crosstool-ng/armv5-gcc-4.3.3-glibc-2.6.1-kernel-2.6.27/arm-unknown-linux-gnueabi/arm-unknown-linux-gnueabi//sys-root/usr/include/linux/in6.h:31: error: redefinition of 'struct in6_addr'
/usr/crosstool-ng/armv5-gcc-4.3.3-glibc-2.6.1-kernel-2.6.27/arm-unknown-linux-gnueabi/arm-unknown-linux-gnueabi//sys-root/usr/include/linux/in6.h:52: error: redefinition of 'struct sockaddr_in6'
/usr/crosstool-ng/armv5-gcc-4.3.3-glibc-2.6.1-kernel-2.6.27/arm-unknown-linux-gnueabi/arm-unknown-linux-gnueabi//sys-root/usr/include/linux/in6.h:60: error: redefinition of 'struct ipv6_mreq'
In file included from /home/arm/priv_git/libwebsockets/lib/plat/unix/unix-sockets.c:31:
/usr/crosstool-ng/armv5-gcc-4.3.3-glibc-2.6.1-kernel-2.6.27/arm-unknown-linux-gnueabi/arm-unknown-linux-gnueabi//sys-root/usr/include/linux/ipv6.h:18: error: redefinition of 'struct in6_pktinfo'
make[2]: *** [lib/CMakeFiles/websockets_shared.dir/build.make:160: lib/CMakeFiles/websockets_shared.dir/plat/unix/unix-sockets.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:4473: lib/CMakeFiles/websockets_shared.dir/all] Error 2
[...]

PR #3572, opened 2026-04-20
[LINUX 3.12] Fix compilation.

ThomasDevoogdt:
- ops-netlink: define IFA_FLAGS for older kernels

IFA_FLAGS was added in Linux 3.14, see:
https://github.com/torvalds/linux/commit/479840ffdbe4242e8a25349218c8e0859223aa35.

So define it ourselves, it's not defined yet.

- unix-sockets: drop the <linux/ipv6.h> on linux < 3.12.0

Linux changed the <linux/ipv6.h> in v3.12.0, see:
https://github.com/torvalds/linux/commit/cfd280c91253cc28e4919e349fa7a813b63e71e8.

So just don't include it.

Fixes:
 - https://github.com/warmcat/libwebsockets/issues/3403


ThomasDevoogdt:
Merged in:
- 429d8a852b8acb751f130269c2817f8d955994a9
- daf09ce90f7295b86ac2ef30df1cc6a8a933b995

PR #3582, opened 2026-05-07
Authentitifcation digest TCP/TLS session reuse

simon0356:
Hello Andy,

As i use HTTP Auth digest along with websockets i want to fix the behaviour using mbedtls that close the TCP socket and TLS session when the digest authentification challenge is received on the client.

I describe my issue and "Claude Sonnet 4.6" has offering the patch here.

I have tested on a linux platform + mbedtls and valgrind to track memory leak (since we reuse the session and reset the header table) 
Also tested on a FreeRTOS + LwIP + mbedtls platform.

I would be happy to have your review on this and if possible to merge it in main.

Thanks !
Simon

### DIFF
--- a/lib/roles/h2/ops-h2.c
+++ b/lib/roles/h2/ops-h2.c
@@ -922,6 +922,7 @@ lws_h2_bind_for_post_before_action(struct lws *wsi)
 		switch (lws_check_basic_auth(wsi, hit->basic_auth_login_file,
 					     hit->auth_mask & AUTH_MODE_MASK)) {
 		case LCBA_CONTINUE:
+		case LCBA_AUTH_RETRY_KEEPALIVE:
 			break;
 		case LCBA_FAILED_AUTH:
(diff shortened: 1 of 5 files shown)


lws-team:
How do you feel about this version of the patch on _temp (it's essentially main plus a few patches) branch... does it also work for you?

https://github.com/warmcat/libwebsockets/commit/a393c8adb3216986037081c2247212bdae1ac8b3

Basically this is more cautious about collecting headers before making a decision, allowing to to defend against chunked / request smuggling.

simon0356:
I have tested your proposal on both setup : 
- Linux host + Mbedtls with valgrind check : Ok
- FreeRTOS + LwIP + Mbedtls : Ok .


It is ok to me, thanks for the review and for the integration.

I have in mind to provide you later a minimal example of websockets server with digest auth support in user callback to allow testing this feature in CI.

Thanks !
Simon

PR #3594, opened 2026-05-29
mbedtls: complete version 4 compatibility

stefan-sedlak:
This completes mbedTLS 4.x compatibility while keeping the change surface
localized to the mbedTLS backend and include boundaries, so existing 2.x and
3.x support is preserved.

mbedTLS 4.x moved a number of low-level details behind TF-PSA private headers
and also needs PSA-backed initialization in standalone crypto paths that older
versions did not.  This patch keeps those differences inside the mbedTLS
integration layer instead of spreading version-specific handling through
generic code.

The main pieces are:

- add the missing 4.x header / feature probing and private include routing
- add 4.x compatibility glue for key handling, x509 handling and crypto helpers
- initialize PSA-backed crypto on standalone RSA / x509 paths where 4.x needs it
- adapt 4.x HMAC setup and unsupported AES-XTS handling
- add local 4.x RSA CRT completion support for imported private keys
- keep the widened RSA completion trigger limited to 4.x so 2.x / 3.x retain
  their previous behaviour
- make the COSE RSA sign path use the returned signature length and fail hard
  instead of serializing an empty signature on error

Validation was kept conservative:

- mbedTLS 2.28.10: api-test-gencrypto, api-test-x509, crypto-cose-sign-7,
  crypto-cose-sign-8 all pass
- mbedTLS 3.6.6: api-test-gencrypto, api-test-x509, crypto-cose-sign-7,
  crypto-cose-sign-8 all pass
- mbedTLS 4.1.0: api-test-gencrypto, api-test-x509, crypto-cose-sign-7,
  crypto-cose-sign-8 all pass
- full 4.1.0 ctest run completed with only the known external warmcat trust
  failures remaining:
  - http-client-warmcat
  - http-client-warmcat-h1
  - http-client-h3

Full 2.x / 3.x broad reruns were also attempted, but those runs were dominated
by unrelated external network / CA-environment instability, so the focused
cross-version regression slice above is the reliable validation signal for this
patch.

Review Note:

It was developed by GPT-5.4. I tried to review it as much as possible. If it is garbage I will keep it on my private repo.


lws-team:
I think it's good to support mbedtls4 but probably these patches aren't quite the way.

Let me take a look tomorrow and see if we can match what your patches do in a slightly better way.

lws-team:
OK the current main is able to build against 4.1.0 mbedtls

https://libwebsockets.org/sai/index.html?task=A22AAA4437564A58089E246ADF285C6960DBB1B060FB1F1F2C5E6229ADF9B4C9

and older

https://libwebsockets.org/sai/index.html?task=A22AAA4437564A58089E246ADF285C69725ADAE520E8F00B3E4C4FABD49625B2

and pass all the ctests.

Either way you need LWS_WITH_MBEDTLS=1 (same for 2.x, 3.x and 4.x) and you can point to the 4.x mbedtls tree in one step with `-DCMAKE_PREFIX_PATH=/opt/mbedtls` or whereever the path is.

It does the necessary adaptations, unfortunately mbedtls itself doesn't support quic/h3 yet.

Please give it a try if you have time.

stefan-sedlak:
I tested it and looks OK for me. Thanks!

PR #3597, opened 2026-06-04
gitmodules: add missing entry for qifs

vikramdattu:
The qifs gitlink under `minimal-examples-lowlevel/api-tests/api-test-qpack`
has no `.gitmodules` entry, so any consumer running
`git submodule update --init --recursive` against this tree fails:

```
fatal: No url found for submodule path 'minimal-examples-lowlevel/api-tests/api-test-qpack/qifs' in .gitmodules
```

Add the entry pointing at `qpackers/qifs` (verified the pinned SHA exists there).


lws-team:
This is on main thanks.

PR #3598, opened 2026-06-04
quic: gate mbedtls_ssl_get_alpn_protocol on LWS_HAVE probe

vikramdattu:
`crypto-quic.c` calls `mbedtls_ssl_get_alpn_protocol()` unconditionally
under `LWS_WITH_MBEDTLS`, but `ssl_pm.c:1054` already guards the same
call behind `LWS_HAVE_mbedtls_ssl_get_alpn_protocol`. With an mbedtls
build that lacks the symbol (e.g. mbedtls 4.x where ALPN access
changed), QUIC linking fails with an undefined-symbol error.

Mirror the `ssl_pm.c` pattern so `prot` stays `NULL` and the
surrounding `if (plen)` branch is a clean no-op when the probe says
the symbol is unavailable. Two-line patch.


vikramdattu:
@lws-team PTAL, would it make sense to add a CI entry just for this? Let me know if needed.

lws-team:
Thanks, this is on main now.

FWIW AFAIK no upstream mbedtls can do quic / h3.  However I have a patch series

https://libwebsockets.org/git/mbedtls/log?h=development

which adds the missing pieces to upstream development branch, and it's tested in Linux + lws main branch and seems functional.

I plan to look into your new idf-6 with this next week, if you're interested you're very welcome to try it first.

vikramdattu:
Thanks for getting these onto main, Andy! For ESP-IDF v6 we're on stock IDF mbedTLS 4 and only use lws for the WSS signaling channel (plain TLS, no QUIC/H3).
All these fixes should unblock mbedtls support on Amazon KVS SDK as well  IMO. [PR link](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/2325)

So the quic/h3 mbedtls series isn't in the path, and hence, I would not be able to have useful contributions there. My related changes mostly on quic were to guard the usage properly.
Happy to get your feedback on espressif/esp-protocols#1070

PR #3600, opened 2026-06-09
mbedTLS 4: complete the FreeRTOS support paths

vikramdattu:
The mbedTLS 4 migration covered the generic (non-FreeRTOS) TLS paths but missed the FreeRTOS/ESP-IDF paths. Building libwebsockets on ESP-IDF v6 (ships mbedTLS 4 / PSA Crypto) hits removed symbols:

- `private-lib-tls.h`: the FreeRTOS branch includes `mbedtls/aes.h` and `mbedtls/gcm.h` unconditionally (both removed in v4). Gated behind `!LWS_HAVE_MBEDTLS_V4`, matching the generic branch right below it.
- `freertos-sockets.c`: falls back to the removed `mbedtls/net.h` because `LWS_HAVE_MBEDTLS_NET_SOCKETS` is never set on FreeRTOS (the probe doesn't run there). Use `mbedtls/net_sockets.h` on v4 — it still carries `mbedtls_net_context` and the `MBEDTLS_ERR_NET_*` codes the BIO callbacks need, even with `MBEDTLS_NET_C` off.
- freertos platform RNG (Amazon-RTOS init + the non-ESP32 `lws_get_random`) used `mbedtls_ctr_drbg_*`/`mbedtls_entropy_*` (removed in v4; the `mec`/`mcdc` context fields are already compiled out for v4). Gated behind `!LWS_HAVE_MBEDTLS_V4` with `psa_crypto_init()`/`psa_generate_random()`, mirroring `lib/plat/unix/unix-init.c`.

Verified on a real ESP-IDF v6 / mbedTLS 4 build: `LWS_HAVE_MBEDTLS_V4` is defined and all mbedTLS-4 header/symbol errors clear.

### DIFF
--- a/.sai.json
+++ b/.sai.json
@@ -58,14 +58,14 @@
                                 "cd build && SAI_CPACK=\"-G DEB\" ${cpack}"
 			]
 		},
-                "netbsd/aarch64BE-bcm2837-a53/gcc": {
-                        "default": false,
-                        "build": [
-				"mkdir -p build destdir; cd build; CCACHE_DISABLE=1 LD_LIBRARY_PATH=../destdir/usr/local/share/libwebsockets-test-server/plugins:../destdir/usr/local/lib cmake .. ${cmake}",
-                                "cd build && make -j$SAI_PARALLEL && rm -rf ../destdir && make -j$SAI_PARALLEL DESTDIR=../destdir install",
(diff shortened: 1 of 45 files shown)


lws-team:
Thanks, it's on main.

PR #3601, opened 2026-06-09
async-dns: fix uninitialized ai/sa46 in IPv6 numeric path

vikramdattu:
In the numeric-address fast path of `lws_async_dns_query()`, the `m == 4` (IPv4) branch derives `ai`/`sa46` from the cache entry `c`, but the `m == 16` (IPv6) branch uses them without deriving. When the cache entry already exists (the allocation block isn't entered), `ai`/`sa46` are uninitialized — GCC flags this as `-Werror=maybe-uninitialized` with `LWS_WITH_IPV6` enabled.

Derive `ai`/`sa46` from `c` in the IPv6 branch, exactly as the IPv4 branch just above does.

### DIFF
--- a/.sai.json
+++ b/.sai.json
@@ -58,14 +58,14 @@
                                 "cd build && SAI_CPACK=\"-G DEB\" ${cpack}"
 			]
 		},
-                "netbsd/aarch64BE-bcm2837-a53/gcc": {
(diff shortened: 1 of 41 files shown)


lws-team:
Thanks it's ok main

PR #3602, opened 2026-06-09
freertos: don't leave LWS_WITH_SPAWN enabled

vikramdattu:
`LWS_WITH_STUB` defaults ON and implies `LWS_WITH_SPAWN` in `CMakeLists-implied-options.txt` before the `LWS_PLAT_FREERTOS` block disables STUB — so SPAWN stays enabled on FreeRTOS even though STUB itself ends up off.

There's no spawn (fork/exec) support on FreeRTOS, and the spawn API's `lsp_cb_t` typedef uses `siginfo_t`, which fails to build with newer newlib (surfaces on ESP-IDF v6); it was latent on older newlib that happened to define `siginfo_t`.

Force `LWS_WITH_SPAWN` off in the FreeRTOS block.

### DIFF
--- a/.sai.json
+++ b/.sai.json
@@ -58,14 +58,14 @@
                                 "cd build && SAI_CPACK=\"-G DEB\" ${cpack}"
 			]
 		},
-                "netbsd/aarch64BE-bcm2837-a53/gcc": {
-                        "default": false,
-                        "build": [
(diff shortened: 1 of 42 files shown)


lws-team:
Thanks it's on main.

PR #3614, opened 2026-06-18
h2: client: fix interop with strict servers and large response headers

saghul:
Three independent h2-client bugs surfaced testing fetch against real
servers (txiki.js).  Symptoms ranged from connections dropped right after
ALPN to RST_STREAM(PROTOCOL_ERROR); they only bit a subset of servers, so
they had gone unnoticed against lenient peers (nghttp2, Cloudflare).

1. lws_h2_client_handshake() unconditionally reassigned the stream's sid
   to highest_sid_opened + 2 at header-send time, overwriting the sid 1
   the first client stream is already bound to (-> 3).  Strict servers
   (Google's GFE) reject a first client stream that isn't sid 1 with
   PROTOCOL_ERROR.  Only allocate a sid here when the stream doesn't
   already have one, and advance highest_sid_opened monotonically.

2. The client sent a Host header and no :authority pseudo-header (the
   :authority code was #if 0'd and the Host-skip guard commented out).
   h2 carries the authority in :authority (RFC 7540 8.1.2.3); GFE rejects
   a request with only Host as PROTOCOL_ERROR.  Send :authority instead;
   it is accepted by all h2 servers.

3. The stock SETTINGS_MAX_HEADER_LIST_SIZE of 4096 is smaller than the
   response headers many real servers send (github.com, GFE, wikipedia),
   so their streams were GOAWAY'd with ENHANCE_YOUR_CALM.  The public
   http2_settings[] override can't reach this index (the apply loop is
   bounded by LWS_H2_SETTINGS_LEN), so raise the stock default to 64KiB;
   actual header storage stays bounded by max_http_header_data.

### DIFF
--- a/lib/roles/h2/http2.c
+++ b/lib/roles/h2/http2.c
@@ -2558,7 +2558,19 @@ lws_h2_client_handshake(struct lws *wsi)
 	 * receives an unexpected stream identifier MUST respond with a
 	 * connection error (Section 5.4.1) of type PROTOCOL_ERROR.
 	 */
-	unsigned int sid = nwsi->h2.h2n->highest_sid_opened + 2;
+	unsigned int sid = wsi->mux.my_sid;
+
+	/*
+	 * If this stream was already bound to a sid (the first client stream is
+	 * bound to sid 1 at the h2 upgrade), keep it.  Only streams that don't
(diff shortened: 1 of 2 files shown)


saghul:
Landed in main.

PR #3617, opened 2026-06-19
Fix MSVC compilation warnings with mbedTLS

saghul:
- **mbedtls: cast sockfd to int when storing in net_context fd**
- **cmake: honor DISABLE_WERROR on MSVC**


saghul:
Landed in main.

PR #3623, opened 2026-07-02
Avoid path doubling issue in CMake files

hmenke:
There is no guarantee that `CMAKE_INSTALL_LIBDIR` (and for that matter `LWS_INSTALL_LIBDIR`) are relative paths, so that it would be safe to concatenate it with `CMAKE_INSTALL_PREFIX`. If `CMAKE_INSTALL_LIBDIR` is set to an absolute path, e.g. `/usr/lib` and `CMAKE_INSTALL_PREFIX` is set to `/usr` this leads to path doubling in various places where there now appears `/usr/usr/lib`.

Here we use the “full” variants of those variables instead which branch correctly on both absolute and relative paths.


lws-team:
How do you feel about this instead?

https://github.com/warmcat/libwebsockets/commit/e00e810b8c7ddb39a1527028ab1115fd7ef991ac

It looks like your patch as it is will have problems with user overrides LWS_INSTALL_LIB_DIR and pkg-config.

hmenke:
@lws-team LGTM :+1:

PR #3624, opened 2026-07-07
tls/mbedtls: accept non-NUL-terminated client_ssl_ca_mem for PEM parsing

vikramdattu:
### Problem

`lws_tls_client_create_vhost_context()` (mbedTLS backend) passes the caller's
`client_ssl_ca_mem` / `client_ssl_ca_mem_len` straight into
`mbedtls_x509_crt_parse()`:

```c
n = mbedtls_x509_crt_parse(ctx->ca_chain, ca_mem, ca_mem_len);
```

`mbedtls_x509_crt_parse()` only treats the buffer as **PEM** when it is
NUL-terminated *and* the passed length includes that NUL byte; otherwise it
parses the buffer as DER and a PEM CA fails with
`MBEDTLS_ERR_X509_INVALID_FORMAT` (`-0x2180`), logged as:

```
client CA: x509 parse failed: -8576
```

Callers commonly pass the *content* length (without the terminator) — which the
OpenSSL backend tolerates — so a PEM CA supplied via `client_ssl_ca_mem` cannot
be loaded on the mbedTLS backend. Note the same file already accounts for this
elsewhere by parsing with `strlen(pem) + 1`.

### Fix

Parse a NUL-terminated copy when the supplied buffer isn't already terminated;
otherwise parse in place. Safe for DER input too (ASN.1 is self-delimiting).

### Reproduced with

AWS Kinesis Video Streams WebRTC C SDK built against mbedTLS 4.1.0 and lws
`main` — the KVS signaling client supplies the AWS root CA bundle via
`client_ssl_ca_mem` with the content length, so every signaling TLS connection
failed to load the CA until the buffer was NUL-terminated.

### DIFF
--- a/lib/tls/mbedtls/mbedtls-client.c
+++ b/lib/tls/mbedtls/mbedtls-client.c
@@ -261,7 +261,28 @@ lws_tls_client_create_vhost_context(struct lws_vhost *vh,
 		if (!ctx->ca_chain)
 			return 1;
 		mbedtls_x509_crt_init(ctx->ca_chain);
-		n = mbedtls_x509_crt_parse(ctx->ca_chain, ca_mem, ca_mem_len);
+		/*
+		 * mbedtls_x509_crt_parse() only treats the buffer as PEM when it
+		 * is NUL-terminated and the length includes that NUL; otherwise
+		 * it falls back to DER and PEM CAs fail with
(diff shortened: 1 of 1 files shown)


lws-team:
Thanks, it's pushed on _temp (a few patches back from HEAD)... this will go to main / Sai as soon as I can fix the last failing quic-interop-runner test

PR #3628, opened 2026-07-08
client: sort-dns: skip AAAA results when ipv6 is disable

zzblydia:
While debugging this I noticed the dns side and the client connection aren't really tied together: 

with async dns the queries go out on the resolver's own wsi, get piggybacked between consumers and end up in a context-owned cache.
But LWS_SERVER_OPTION_DISABLE_IPV6 is a context / vhost option. So instead of teaching the shared resolver about it, I'm honouring it where a particular client wsi actually picks up the results — in lws_sort_dns(), which builds the candidate list for both the async and the blocking path.

wsi->ipv6 is already 0 in all the cases that matter here: **option set on the context or the vhost, or the client bound to an IPv4 source address**. Without this a "disabled" wsi can still connect over ipv6, eg via a cache entry some dual-stack consumer created earlier, and then things like lws_get_peer_addresses() read the peer sockaddr as sockaddr_in and get garbage.

The resolver still asks for both A and AAAA — on purpose. The blocking path does the same thing (getaddrinfo with PF_UNSPEC, glibc sends both), and a shared query might be serving a dual-stack consumer at the same time anyway.
Both paths gather everything and the per-wsi choice happens at the same place.

## Verification

I had tested on a dual-stack server within a local network.
**Environment:** Ubuntu 24.04 x86_64, `LWS_WITH_IPV6=ON`, `LWS_WITH_SYS_ASYNC_DNS=ON`.

* **Case 1: `DISABLE_IPV6` NOT set**
  * **Behavior:** The client wsi get two ip( ipv4 and ipv6) and firstly connects to the server via **IPv4**(RFC6724 default policy table).
* **Case 2: `DISABLE_IPV6` enabled**
  * **Behavior:** The client successfully ignores the cached AAAA/IPv6 results and connects via **IPv4**.


zzblydia:
Code reviewed and comments updated.
```c
    /*
	 * lws_sort_dns() filters out IPv6 results if wsi->ipv6 == 0, pick up
	 * the vhost / context ipv6 policy
	 */
```

PR #3632, opened 2026-07-17
h2: client: mark reused mux streams as mux_substream

saghul:
lws_wsi_h2_adopt() set client_mux_substream but not mux_substream on a
reused (pooled) HTTP/2 client stream, whereas the first stream on a
connection carries mux_substream = 1. The HPACK decoder only captures
custom / non-indexed response headers into the ah unknown-header list
(ah->unk_ll_head) when mux_substream is set, so on a reused client stream
those headers are never stored: the unknown-header list is left stale, and
lws_hdr_custom_length/copy/name_foreach then walk it and read out of bounds
(observed as a heap-use-after-free when the walk runs past the ah into an
adjacent freed header table).

This also silently drops all custom response headers on the 2nd+ request of
a pooled h2 client connection.

Set mux_substream on the reused client stream too, matching the first
stream, so HPACK captures custom response headers correctly.


lws-team:
Thanks, it's pushed on main

PR #3633, opened 2026-07-17
client: don't re-remove already-failed parallel connect attempts

vikramdattu:
A parallel (happy-eyeballs) connect attempt that fails synchronously in connect() (eg, EHOSTUNREACH for an unroutable AAAA result) is cleaned up at the failure site: socket closed, fds-table slot removed, `parallel_conns` entry invalidated. Its `position_in_fds_table` is never assigned (that only happens on the EINPROGRESS path).

When the surviving attempt then connects, the completion path iterates all `parallel_conns` slots up to `parallel_count` without checking `is_valid`, and `lws_remove_parallel_fd_safely()` uses the unassigned position to remove an unrelated fds-table slot and double-closes a possibly since-reused fd.

Depending on what the scrambled fds table hits, the connection loses its POLLOUT events (a client POST body is never sent and the transaction stalls to timeout), and `lws_context_destroy()` walks a stale entry into a freed wsi (SIGSEGV/SIGBUS).

**Reproduce:** from any host with A+AAAA DNS but no IPv6 route (eg, GitHub Actions macOS runners), run a plain client POST (minimal-http-client-post pattern) against a dual-stack endpoint such as `kinesisvideo.us-west-2.api.aws:443`. The v4 leg connects, the v6 leg fails with EHOSTUNREACH, `LWS_CALLBACK_CLIENT_HTTP_WRITEABLE` never fires, and teardown crashes in `__lws_close_free_wsi`.

**Fix:** guard `lws_remove_parallel_fd_safely()` against invalid/never-registered entries, and restore the `is_valid` check at the connected-path cleanup loop.

Verified: repro completes cleanly (headers, body, response, clean destroy) with the fix; a downstream consumer (AWS KVS WebRTC SDK test suite) runs crash-free on macOS arm64 + x86_64 with mbedTLS 4.1.


lws-team:
Thanks, I fetched it from your tree with it on to avoid github mush.

I guess this is not helping your situation today, but FYI lws on main has its own webrtc mixer plugin for the server side of this; there's also an example for doing the client side for webcam type usage.

PR #3634, opened 2026-07-18
ws-over-h2 (RFC 8441): flow control, permessage-deflate RSV1, client extended CONNECT, teardown crash — with selftest

thefallentree:


lws-team:
Thanks, that's pushed on main and is a big help.

PR #3635, opened 2026-07-18
ws-over-h2: don't conflate nwsi buffered output with the stream's par…

thefallentree:
…ked frames

The tx flow-control parking introduced in 57effc65 gates the park, drain and re-arm logic on lws_has_buffered_out(), which for an h2 stream also reports the NETWORK wsi's buflist_out -- the buffer that fills whenever lws_issue_raw() takes a partial socket write.  The drain logic, however, assumes the predicate describes the stream's OWN parked frames, which breaks four ways under ordinary socket-level backpressure:

 - lws_h2_ws_drain_parked_tx(): after the stream's last parked frame is consumed, a partial write left by its own lws_issue_raw(nwsi) keeps the while (lws_has_buffered_out(wsi)) loop alive via the nwsi branch; the next iteration reads the empty stream buflist and returns -1, and rops_perform_user_POLLOUT_h2() closes a healthy stream that had just flushed everything it owned.

 - rops_perform_user_POLLOUT_h2(): the post-drain re-arm is skipped whenever the nwsi still holds a partial, on the assumption that a WINDOW_UPDATE will re-arm the child.  When the peer's window is large (tx credit ample, bottleneck is the socket) no WINDOW_UPDATE is coming, requested_POLLOUT was already consumed, and the pending- writable accounting then drops POLLOUT on the connection: the user callback never fires again and the transfer stalls permanently.

 - the same gate pulls a SIBLING ws stream -- nothing of its own parked, POLLOUT requested -- into the drain once an earlier child's drain congests the nwsi in the same service pass, hitting the identical empty-buflist close.

 - lws_h2_frame_write(): the parking predicate fires on nwsi congestion caused by another stream, needlessly deferring fully-credited frames into the drain path instead of letting lws_issue_raw(nwsi) append them in order.

Test the stream's own buflist_out directly in all four places, and let a carries-ws child with nothing of its own parked fall through to normal servicing.  The existing api-test doesn't catch these because its drip-fed WINDOW_UPDATEs continually re-arm every child and localhost doesn't produce partial writes.

### DIFF
--- a/lib/roles/h2/http2.c
+++ b/lib/roles/h2/http2.c
@@ -706,7 +706,7 @@ int lws_h2_frame_write(struct lws *wsi, int type, int flags,
 	if (type == LWS_H2_FRAME_TYPE_DATA) {
 #if defined(LWS_ROLE_WS)
 		if (wsi->h23_stream_carries_ws &&
-		    (lws_has_buffered_out(wsi) ||
+		    (wsi->buflist_out ||
 		     lws_h2_tx_cr_get(wsi) < (int)len)) {
 			/*
 			 * ws-over-h2 (RFC 8441): not enough h2 flow-control
@@ -768,7 +768,13 @@ lws_h2_ws_drain_parked_tx(struct lws *nwsi, struct lws *wsi)
 {
 	uint8_t out[LWS_H2_FRAME_HEADER_LENGTH + 4096];
 
-	while (lws_has_buffered_out(wsi)) {
+	/*
+	 * Test the stream's own buflist directly: lws_has_buffered_out()
+	 * additionally reports the nwsi's buffer, which fills up whenever
+	 * lws_issue_raw() below takes a partial socket write and must not
+	 * keep us looping after our own parked frames are gone.
+	 */
+	while (wsi->buflist_out) {
 		uint8_t *seg = NULL;
 		size_t sl = lws_buflist_next_segment_len(&wsi->buflist_out,
 							 &seg);
--- a/lib/roles/h2/ops-h2.c
+++ b/lib/roles/h2/ops-h2.c
@@ -1089,13 +1089,20 @@ rops_perform_user_POLLOUT_h2(struct lws *wsi)
 
 		/* priority 1: post compression-transform buffered output */
 
-		if (lws_has_buffered_out(w)) {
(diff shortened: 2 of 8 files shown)


lws-team:
Thanks... this stuff is really quite hard... is there not a situation the patch doesn't take care of properly, where

```
#if defined(LWS_ROLE_WS)
               if (w->h23_stream_carries_ws) {
                       if (w->buflist_out) {
                               /* drain parked tx ... */
                               goto next_child;
                       }
               } else
#endif
               if (lws_has_buffered_out(w)) {
                       /* complete partial / apply backpressure ... */
                       goto next_child;
               }
 ```
 
if we are h23_stream_carries_ws, the `} else` as the end makes us skip backpressure checking and continues on to keeping packing a buflist forever, even if the nwsi is choked.
 
 How do you feel about this
 
 ```
 #if defined(LWS_ROLE_WS)
		if (w->h23_stream_carries_ws && w->buflist_out) {
			/*
			 * ws-over-h2: whole DATA frames parked by
			 * lws_h2_frame_write() waiting for tx credit
			 */
			if (lws_h2_ws_drain_parked_tx(wsi, w) < 0) {
				lwsl_info("%s signalling to close\n", __func__);
				lws_close_free_wsi(w, LWS_CLOSE_STATUS_NOSTATUS,
						   "h2 ws drain rxflow");
				wa = &wsi->mux.child_list;
				goto next_child;
			}
			if (!w->buflist_out)
				/* fully drained: let the user write */
				lws_callback_on_writable(w);
			
			wa = &wsi->mux.child_list;
			goto next_child;
		}
#endif
		/* Catch streams when nwsi is choked, INCLUDING ws-over-h2 streams */
		if (lws_has_buffered_out(w)) {
			lwsl_debug("%s: completing partial\n", __func__);
			if (lws_issue_raw(w, NULL, 0) < 0) {
				lwsl_info("%s signalling to close\n", __func__);
				lws_close_free_wsi(w, LWS_CLOSE_STATUS_NOSTATUS,
						   "h2 end stream 1");
				wa = &wsi->mux.child_list;
				goto next_child;
			}
			lws_callback_on_writable(w);
			wa = &wsi->mux.child_list;
			goto next_child;
		}
```

PR #3639, opened 2026-07-21
client: fix pipelining match when host and address differ

zzblydia:
I am using libwebsockets v4.3.3 as an http client, sending GET requests as a heartbeat to my server over a keep-alive connection.

In `lws_client_connect_info` I set `address` = "127.0.0.1" and `host` = "127.0.0.1:8081".

From packet capture I found the GET requests were not reusing the tcp connection.  After reading the logs
```
D: [wsicli|1|GET/h1/default/127.0.0.1]: lws_vhost_active_conns: check [wsicli|0|GET/h1/default/127.0.0.1] 127.0.0.1 127.0.0.1:8081 8081 8081
I: [wsicli|1|GET/h1/default/127.0.0.1]: lws_client_connect_2_dnsreq: adding as active conn
```
and stepping through the code, I found the match in `lws_vhost_active_conns` never succeeds:
```c
!strcmp(adsin, w->cli_hostname_copy)
```
because `adsin` is taken from `CIS_ADDRESS` while `cli_hostname_copy` is taken from `CIS_HOST`, so `127.0.0.1` is compared against `127.0.0.1:8081` and every transaction opens a new connection.

[Here is a minimal reproducible example for this issue](https://gist.github.com/zzblydia/4944809097d94e54327451bb6cc542ca)


lws-team:
Thanks, it's pushed on main

PR #3641, opened 2026-07-23
quic: select AEAD and header-protection cipher from negotiated suite

saghul:
The built-in QUIC/H3 stack derived its packet-protection AEAD and
header-protection (HP) cipher from the length of the exported TLS 1.3
traffic secret alone: a 48-byte secret was treated as AES-256-GCM and
anything else as AES-128-GCM. ChaCha20-Poly1305 was never selected.

Per RFC 9001 (5.3, 5.4) the AEAD and HP algorithms are fixed by the
negotiated TLS 1.3 cipher suite, not by the secret length. Because
TLS_AES_128_GCM_SHA256 and TLS_CHACHA20_POLY1305_SHA256 both use SHA-256,
they produce identical 32-byte secrets, so a ChaCha20 handshake was
mis-keyed with AES header protection and failed ("reserved bits" /
decryption failure) whenever ChaCha20-Poly1305 was negotiated.

Make the choice cipher-suite aware:

  - add enum lws_tls_quic_aead and wsi->tls.quic_aead, reported by the
    TLS backend from the negotiated suite;
  - every QUIC-capable TLS backend now reports it from the negotiated
    cipher as the traffic secrets are installed:
      * mbedTLS: maps the negotiated TLS 1.3 ciphersuite id in the
        traffic-secret export callback (the suite is only available there
        mid-handshake; ssl->session, which
        mbedtls_ssl_get_ciphersuite_id_from_ssl() reads, is not yet set);
      * OpenSSL/BoringSSL/AWS-LC/LibreSSL: from the SSL_CIPHER handed to
        set_{read,write}_secret();
      * wolfSSL: from SSL_get_current_cipher() in set_encryption_secrets();
      * GnuTLS: from gnutls_cipher_get();
      * SChannel: from the SEC_TRAFFIC_SECRETS symmetric alg / key size;
  - lws_quic_set_keys() selects the internal cipher_type from the
    reported AEAD, falling back to the legacy length heuristic only when
    a backend does not report it (still correct for AES-128 vs AES-256
    GCM, blind to ChaCha20 as before).

The correct SHA-256/SHA-384 HKDF hash is still chosen from the secret
length, which is unambiguous (it mirrors the suite's PRF hash size).

Note: the mbedTLS QUIC support patch must extend the
mbedtls_ssl_transport_ops::set_traffic_secrets callback with the
negotiated ciphersuite id (handshake->ciphersuite_info->id) for the
mbedTLS backend to report it.

### DIFF
--- a/include/libwebsockets/lws-quic.h
+++ b/include/libwebsockets/lws-quic.h
@@ -38,6 +38,22 @@ enum lws_tls_quic_secret_type {
 	LWS_TLS_QUIC_SECRET_SERVER_APPLICATION,
 };
 
+/*
+ * RFC 9001 4.1 / 5.3 / 5.4: the QUIC packet-protection AEAD and the header-
+ * protection cipher are both fixed by the negotiated TLS 1.3 cipher suite, and
+ * cannot be inferred from the traffic-secret length alone (TLS_AES_128_GCM_
+ * SHA256 and TLS_CHACHA20_POLY1305_SHA256 both use SHA-256 and so produce a
+ * 32-byte secret).  The TLS backend reports the negotiated AEAD to the QUIC
+ * role via wsi->tls.quic_aead so the correct algorithms are selected.
+ */
+enum lws_tls_quic_aead {
+	LWS_TLS_QUIC_AEAD_UNKNOWN,	   /**< backend didn't report; caller
+					     * falls back to length heuristic */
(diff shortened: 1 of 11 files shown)


saghul:
Note: I only tested the mbedTLS version. For it to work the mbedTLS QUIC patch needs to be adapted:

[mbedtls-quic-chacha20.patch](https://github.com/user-attachments/files/30296245/mbedtls-quic-chacha20.patch)

PR #3643, opened 2026-07-23
h3: client: send request bodies (fix :method encoding + add body-upload phase)

saghul:
### Summary

The built-in QUIC/HTTP-3 client could not send a request body. A `POST` (or any
body-bearing request) sent its HEADERS but no `DATA`, so it arrived at the peer
with an empty body. This PR fixes two independent bugs in `lib/roles/h3/ops-h3.c`
that were both required to make request bodies work end-to-end. h1 and h2 client
bodies were unaffected.

### Bug 1 — `:method` was always encoded as GET

`lws_h3_client_handshake()` read the request method only from the
`_WSI_TOKEN_CLIENT_METHOD` ah token and defaulted to `"GET"` when it was absent.
But h3 client mux child streams carry the request line in the generic client
**stash** (`wsi->stash->cis[]`) rather than ah header tokens — which is exactly
why the `:authority` and `:path` encoding right below it already fall back to
`wsi->stash->cis[CIS_ADDRESS]` / `cis[CIS_PATH]`. The method had no such
fallback, so every h3 client request went out as `:method GET`.

A `POST` therefore reached the peer as a `GET`: its `content-length` header
survived, but the server discarded the body because a GET isn't expected to
carry one. This is the primary cause of the "empty body" symptom.

Fix: source the method from `wsi->stash->cis[CIS_METHOD]` when the token is
absent, mirroring the existing `:authority`/`:path` handling.

### Bug 2 — no request-body upload phase after HEADERS

`rops_perform_user_POLLOUT_h3()` moved the stream straight from
`LRS_H2_WAITING_TO_SEND_HEADERS` to `LRS_WAITING_SERVER_REPLY` once the HEADERS
were written, and never drove the `LWS_CALLBACK_CLIENT_HTTP_WRITEABLE`
body-upload cycle the way the h1 and h2 clients do. So even with the correct
method, `client_http_body_pending` was set but the writeable callback was never
delivered and the body `DATA` frame was never framed/flushed.

Fix: mirror the h1 client state machine
(`lib/roles/http/client/client-http.c`) — after HEADERS, if
`client_http_body_pending`, enter an explicit `LRS_ISSUE_HTTP_BODY` phase that
keeps driving `LWS_CALLBACK_CLIENT_HTTP_WRITEABLE` (each write framed as an H3
`DATA` frame, `LWS_WRITE_HTTP_FINAL` setting `H2_STREAM_END` so the QUIC role
FINs the stream), arm `PENDING_TIMEOUT_CLIENT_ISSUE_PAYLOAD`, and only switch to
`LRS_WAITING_SERVER_REPLY` once the body (and its FIN) have been handed to the
transport. The branch is gated only on `client_http_body_pending` (not
`lws_has_buffered_out()`, which for h3 also counts the HEADERS still sitting in
the QUIC `pending_tx` queue and would wrongly pull bodyless GETs into the body
phase). Bodyless GET/HEAD requests are unchanged. There is no POLLOUT re-arm
inside the phase, matching h1: the user re-arms while it has more body, or the
QUIC stream flow-control window re-arms it on `MAX_STREAM_DATA`.

### Testing

Verified end-to-end against an h3-enabled server (client POST → server echoes the
body back):

- `POST` bodies from 176 B to 300 KB (multi-frame, multi-packet, exceeding the
  initial 65535-byte stream flow-control window) round-trip **byte-for-byte**.
- Server observes `:method POST` and receives the full body (before this change:
  `:method GET`, 0 body bytes received).
- No `FINAL_SIZE_ERROR` / `CONNECTION_CLOSE` / flow-control errors.

Compiles clean under `-Wall -Wextra -Wconversion -Wsign-compare` with
`LWS_WITH_HTTP3` / `LWS_ROLE_QUIC` / `LWS_WITH_CLIENT`.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -24,20 +24,30 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@v3
+        uses: github/codeql-action/init@v4
         with:
           languages: ${{ matrix.language }}
           config-file: ./.github/codeql.yml
           queries: +security-and-quality
 
-      - name: Autobuild
-        uses: github/codeql-action/autobuild@v3
-        if: ${{ matrix.language == 'javascript' || matrix.language == 'cpp' }}
+      - name: Autobuild javascript
+        uses: github/codeql-action/autobuild@v4
+        if: ${{ matrix.language == 'javascript' }}
+
+      - name: Build cpp
+        if: ${{ matrix.language == 'cpp' }}
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libuv1-dev libev-dev libglib2.0-dev
+          mkdir build
+          cd build
+          cmake ..
+          make -j$(nproc)
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@v3
+        uses: github/codeql-action/analyze@v4
         with:
           category: "/language:${{ matrix.language }}"
--- a/.sai.json
+++ b/.sai.json
@@ -278,6 +278,9 @@
 		"ipv6-off":	{
 			"cmake":	"-DLWS_IPV6=OFF"
 		},
+		"ipv4-off":	{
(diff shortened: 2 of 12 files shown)


saghul:
landed.

PR #3654, opened 2026-08-13
Name the arguments the public header docs actually describe

darkdi:
Was reading `lws_jwt_token_sanity` to work out what it hands back. The doc says `secs_left`, set to the number of seconds of valid auth left. The argument is `exp_unix_time`, and in `jws.c` it takes the `exp` claim straight off the token, so it is an absolute unix time. Anything trusting that line computes the wrong thing.

Went through the rest of `include/libwebsockets/` after that and the same thing turns up across a spread of headers. Most are ordinary renames the comment never followed, `pai` for `ai`, `blocksize` for `block_size`, `max_len` for `max`. Two are more than a name: `lws_jwe_encrypt` and `lws_jwe_auth_and_decrypt` still document the old `jose` and `jws` pair from before those were folded into `struct lws_jwe`, so each collapses into the single argument they take now. Where a rename left the description pointing at the wrong thing I took that too, `mgr` in `lws-state.h` was still described as the lws_context and `pwmseq` as a target brightness when it is a sequence definition.

Left alone deliberately: the `#define` helpers such as `lws_ptr_diff` and `lws_end_foreach_ll_safe`, where the documented name genuinely is the macro parameter, and the NXP SDK vendored under `minimal-examples/embedded/rt595`, which is not yours to carry. `lwsac_assert_valid` is the same case and nearly caught me out, since the docblock sits directly above `_lwsac_assert_valid` and that one does take `aco`, but the block is titled for the macro and `check` and `len` match the macro too.

### DIFF
--- a/include/libwebsockets/lws-async-dns.h
+++ b/include/libwebsockets/lws-async-dns.h
@@ -103,7 +103,7 @@ lws_async_dns_query(struct lws_context *context, int tsi, const char *name,
 /**
  * lws_async_dns_freeaddrinfo() - decrement refcount on cached addrinfo results
  *
- * \param pai: a pointert to a pointer to first addrinfo returned as result in the callback
+ * \param ai: a pointert to a pointer to first addrinfo returned as result in the callback
  *
  * Decrements the cache object's reference count.  When it reaches zero, the
  * cached object may be reaped subject to LRU rules.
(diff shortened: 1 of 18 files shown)


darkdi:
These changes are already on main, cherry-picked as 7d106f0f and f19c373f. Git also reports both commits from this branch as already applied, so I am closing the now-obsolete PR. Thanks!

PR #3657, opened 2026-08-20
gnutls x509 ecc curve bit fix

TheGreatSage:
Using the gnutls_ecc_curve_t types is incorrect for generating an x509 cert using any of them will result in a legacy 192 bit curve, which breaks things. You could opt to get rid of the the curve type and just hard code the bits in. I decided to make use of the get size function which should also work.


TheGreatSage:
Looks like this got added. Thanks. Keep up the good work.

PR #3661, opened 2026-08-31
openHiTLS: fix no-client and reduced-log builds

pi-ixel:
openhitls: fix no-client and reduced-log builds

Only compile the openHiTLS session implementation when TLS session
support is enabled, matching the other TLS backends.

Guard cipher-description code by the INFO log level to avoid unused
variable errors in Release builds with -Werror.


lws-team:
Thanks, it's in main and v5.0-stable

PR #3667, opened 2026-09-09
dht: clean up name when node ID allocation fails

junkilee80:
### Problem

When `lws_dht_create()` is given a name, it duplicates that string before
creating the node ID hash. If the hash allocation fails, the current error path
frees only the DHT context and returns, leaving the duplicated name allocated.

This requires an allocation failure while optional DHT support is in use, so the
change is presented as failure-path robustness rather than a security claim.

### Fix

Route the node ID allocation failure through the existing `fail:` cleanup path.
That path calls `lws_dht_destroy()`, which already handles a partially initialized
context and releases both `ctx->name` and the context. The existing error log is
preserved.

### Test

Add `api-test-dht-create`, which installs a custom allocator and fails the
`lws_dht_hash_create` allocation by its allocation reason. It checks that DHT
creation fails and that no allocations made by `lws_dht_create()` remain live.

On macOS arm64 with Apple Clang 21 and DHT/DHT backend enabled:

- the new test fails on the unmodified tree with one live allocation;
- the new test passes with this patch;
- the existing `api-test-dht-msg-parse` test passes;
- both shared and static library targets build successfully;
- the new test passes in an AddressSanitizer build without a sanitizer report.

The candidate was initially found with RepoAudit, and the production cleanup was
suggested by PailGen using Gemini 3.6 Flash. The regression test and submission
preparation were co-developed with OpenAI Codex based on GPT-5. The failure path,
ownership, final patch, and test were manually reviewed and verified.

### DIFF
--- a/lib/misc/dht/dht.c
+++ b/lib/misc/dht/dht.c
@@ -705,9 +705,7 @@ lws_dht_create(const lws_dht_info_t *info)
 
 	if (!ctx->myid) {
 		lwsl_err("ctx->myid creation failed\n");
-		lws_free(ctx);
-
-		return NULL;
+		goto fail;
 	}
 
 	if (info->v) {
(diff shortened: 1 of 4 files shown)


lws-team:
Thanks, it's pushed on main

